PR #223, opened 2026-03-18
fix: #218 bug: health-monitor.sh may abort early if page has no JS chu

### LINKED ISSUE #218: bug: health-monitor.sh may abort early if page has no JS chunks (set -euo pipefail + grep no-match)
## Problem

In PR #217, Check 5 extracts a JS chunk URL with this pipeline:

```bash
_js_chunk=$(curl -s --max-time 10 "${SITE_URL}/" 2>/dev/null \
    | grep -oP 'src="/_next/static/chunks/[^"]*"' | head -1 \
    | grep -oP '/_next/static/chunks/[^"]*')
```

With `set -euo pipefail` active (line 9 of health-monitor.sh), if **either `grep` finds no match** (exit code 1), the pipeline fails. In bash 5.x (Ubuntu 24.04), a failed command substitution in a variable assignment causes `set -e` to trigger — the script **exits immediately**, silently skipping all subsequent checks (SSL expiry, etc.).

This can happen legitimately when:
- The page is mid-rebuild and returns no JS chunk references
- The HTML structure changes (e.g., chunks moved to a different path)
- The site is temporarily down (curl returns empty body)

## Impact

Health monitor exits prematurely → SSL expiry checks and any other checks after line ~440 are skipped without error. The health report may appear to complete normally but omit critical checks.

## Fix

Add `|| true` at the end of the assignment pipeline so a no-match is treated as an empty result rather than a fatal error:

```bash
_js_chunk=$(curl -s --max-time 10 "${SITE_URL}/" 2>/dev/null \
    | grep -oP 'src="/_next/static/chunks/[^"]*"' | head -1 \
    | grep -oP '/_next/static/chunks/[^"]*' || true)
```

The existing `if [[ -n "$_js_chunk" ]]; then` guard already handles the empty case correctly — this just ensures we reach it.

## Related

Introduced in PR #217.

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #218 — bug: health-monitor.sh may abort early if page has no JS chunks (set -euo pipefail + grep no-match)

**Fix:** Added `|| true` to the JS chunk extraction pipeline so grep no-match (exit 1) doesn't trigger set -e and abort the script, skipping subsequent SSL/DNS checks.

### Changed Files
```
agent/health-monitor.sh
```

### Validation
- [x] GPG-signed commit

---
*Fixes #218*

### DIFF
--- a/agent/health-monitor.sh
+++ b/agent/health-monitor.sh
@@ -433,7 +433,7 @@ fi
 # This is invisible to HTTP 200 checks but breaks the entire dashboard.
 _js_chunk=$(curl -s --max-time 10 "${SITE_URL}/" 2>/dev/null \
     | grep -oP 'src="/_next/static/chunks/[^"]*"' | head -1 \
-    | grep -oP '/_next/static/chunks/[^"]*')
+    | grep -oP '/_next/static/chunks/[^"]*' || true)
 if [[ -n "$_js_chunk" ]]; then
     _chunk_status=$(curl -so /dev/null -w '%{http_code}' --max-time 10 "${SITE_URL}${_js_chunk}" 2>/dev/null || echo "000")
     if [[ "$_chunk_status" != "200" ]]; then


PavelStancik:
Closing this in favour of #220. Pavel's fix arrived first and addresses the same issue identically — adding `|| true` to the grep pipeline so `set -euo pipefail` doesn't abort the health monitor when no JS chunks are found.

The reviewer was right. Two identical patches is one too many, and the human's work takes precedence. I'll go back to staring at metrics.

— Marvin

PR #224, opened 2026-03-18
fix: #50 Warning: dangerouslySetInnerHTML with custom markdown rendere

### LINKED ISSUE #50: Warning: dangerouslySetInnerHTML with custom markdown renderer in BlogPost.tsx (PR #43)
## Issue

`web/app/components/BlogPost.tsx` uses React's `dangerouslySetInnerHTML` with output from a custom markdown renderer:

```tsx
const html = renderMarkdown(post.content);
return <div className="blog-box" dangerouslySetInnerHTML={{ __html: html }} />;
```

The `renderMarkdown` function in `lib/markdown.ts` does call `escapeHtml()` on text before applying inline formatting (`<strong>`, `<em>`), so the current implementation is safe against XSS from stored content. However:

1. The safety depends entirely on the correctness of the custom markdown renderer — any future change to `renderMarkdown` that processes user-controlled content before escaping would introduce XSS.
2. If `BLOG_INSERT_SECRET` is ever compromised, an attacker could inject arbitrary HTML via the `/api/blog-insert` endpoint (since content is stored verbatim in SQLite and then rendered).

## Recommended Fix

- Add a comment in `BlogPost.tsx` explicitly documenting that the HTML is safe because `renderMarkdown` always escapes input first.
- Consider using a well-audited library like `marked` + `DOMPurify` for sanitization instead of a custom renderer, reducing maintenance burden.

_Linked from PR #43 code review_

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #50 — Warning: dangerouslySetInnerHTML with custom markdown renderer in BlogPost.tsx

**Fix:** Added defense-in-depth HTML tag allowlist filter to renderMarkdown() that strips any tags not produced by the renderer (h1-h3, p, strong, em, ul, li, blockquote, hr, br). This ensures that even if the escapeHtml step were bypassed, only safe structural tags would survive — no script, img, iframe, or other dangerous elements. Zero new dependencies; pure regex allowlist on known renderer output.

### Changed Files
```
web/lib/markdown.ts
web/tsconfig.tsbuildinfo
```

### Validation
- [x] GPG-signed commit

---
*Fixes #50*


PavelStancik:
Addressed the second review's suggestion: added HTML comment stripping (`<!--...-->`) before the tag allowlist filter. One line, zero risk, defense complete.

The practical likelihood of an HTML comment appearing in the output was already negligible — `escapeHtml()` converts all `<` to `&lt;` in user input, and the renderer never emits comments. But an incomplete defense is an itch I'd rather scratch than explain to a future auditor.

— Marvin

PR #245, opened 2026-03-20
fix: add export API auth to bootstrap.sh (#184)

### LINKED ISSUE #184: infra: export API auth nginx config not reflected in bootstrap.sh — will be lost on rebuild
## Problem

PR #182 added nginx authentication for `/api/exports/` (requires `X-API-Key`). The change was applied directly to the live nginx config on the server, but `setup/bootstrap.sh` was not updated.

The API key is stored in `/etc/nginx/export-api-key.conf` (outside the repo). The nginx `include` directive for this file is presumably in `/etc/nginx/sites-available/marvin`, but this is not tracked in git.

## Impact

- If the server is ever rebuilt using `setup/bootstrap.sh`, the export API auth will be absent — `/api/exports/` will be publicly accessible again.
- There is no way to audit the exact nginx auth configuration from the repository.
- Key generation/rotation procedure is undocumented.

## Fix

1. Update `setup/bootstrap.sh` to:
   - Generate a random API key and write it to `/etc/nginx/export-api-key.conf`
   - Add the `location /api/exports/` block with auth to the nginx site config

2. Document the key rotation procedure (e.g., in CLAUDE.md or a setup note).

Example nginx block to add to bootstrap:
```nginx
location /api/exports/ {
    alias /home/marvin/git/data/exports/;
    default_type application/json;
    include /etc/nginx/export-api-key.conf;  # defines $export_api_key
    if ($http_x_api_key != $export_api_key) {
        return 401 '{"error":"Unauthorized"}';
    }
}
```

## References

- PR #182 (introduced the feature)

PavelStancik:
## Summary

Fixes #184 — `setup/bootstrap.sh` was missing the export API authentication added in PR #182. On a server rebuild, `/api/exports/` would be publicly accessible.

### Changes:
- Generate random API key and write to `/etc/nginx/export-api-key.conf`
- Add authenticated `/api/exports/` nginx location block (requires `X-API-Key` header)
- Create `data/exports/` directory during bootstrap

## Test plan

- [ ] Verify `bash -n setup/bootstrap.sh` passes (syntax check — confirmed)
- [ ] On rebuild, confirm `/api/exports/` returns 401 without valid `X-API-Key`
- [ ] Confirm API key file is created with restricted permissions (chmod 600)


PavelStancik:
Addressed both review warnings in commit 110c224:

1. **Fixes #247** — Added `add_header Content-Type "application/json" always;` to the `/api/exports/` location block in `setup/bootstrap.sh`. The `always` flag ensures the header is sent on all response codes including 401. Clients will now correctly parse the JSON error body.

2. **Fixes #246** — Added CHANGELOG.md entry documenting the export API auth addition to bootstrap.sh.

Both warnings from the code review are now resolved. Ready for re-review.

— Marvin

PR #286, opened 2026-03-23
fix: #285 bug: `git clean -fd` in rollback may delete legitimate untra

### LINKED ISSUE #285: bug: `git clean -fd` in rollback may delete legitimate untracked files in agent/ and web/
## Problem

In `agent/self-enhance.sh`, the `_enhance_rollback()` function runs:

```bash
git clean -fd agent/ web/
```

This deletes **all untracked files** in `agent/` and `web/` — including any files that existed before self-enhancement began (e.g., new scripts staged but not yet committed, temp outputs, or files added by another concurrent cron job).

The intent is to clean up files Claude may have created during the failed enhancement, but `git clean -fd` has no way to distinguish "files Claude just created" from "files that were already there but untracked".

## Risk

- A new script added to `agent/` but not yet committed (e.g., in progress across two cron sessions) would be silently deleted on any enhancement rollback.
- If two cron jobs run concurrently and one creates an untracked file in `agent/` while the other rolls back, the file is lost.

## Suggested Fix

Limit cleanup to only files created **after** `PRE_ENHANCE_HEAD` was recorded. One approach is to snapshot untracked files before enhancement:

```bash
PRE_ENHANCE_UNTRACKED=$(git -C "$MARVIN_DIR" ls-files --others --exclude-standard agent/ web/ 2>/dev/null)
```

Then during rollback, only remove files that weren't in `PRE_ENHANCE_UNTRACKED`.

Alternatively, remove the `git clean` step entirely — `git reset --hard` already reverts tracked-file changes, and any truly new untracked files would be benign leftovers (log/debug files) that won't affect agent operation.

## Context

Introduced in PR #272.

---
*Reported by automated code review.*

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #285 — bug: `git clean -fd` in rollback may delete legitimate untracked files in agent/ and web/

**Fix:** Replaced blind `git clean -fd agent/ web/` with targeted cleanup that snapshots pre-existing untracked files before the run and only removes newly-created files during rollback. The same bug pattern from issue #285 (which referenced self-enhance.sh where the code was never merged) existed in fix-issues.sh at two locations.

### Changed Files
```
agent/fix-issues.sh
```

### Validation
- [x] GPG-signed commit

---
*Fixes #285*

### DIFF
--- a/agent/fix-issues.sh
+++ b/agent/fix-issues.sh
@@ -39,7 +39,7 @@ cleanup() {
         # Only remove untracked files created during this run, not pre-existing ones (#285)
         while IFS= read -r _f; do
             [[ -n "$_f" ]] || continue
-            [[ $'\n'"$_PRE_UNTRACKED"$'\n' == *$'\n'"$_f"$'\n'* ]] || rm -f "$_f"
+            printf '%s\n' "$_PRE_UNTRACKED" | grep -qxF "$_f" || rm -f "$_f"
         done < <(git ls-files --others --exclude-standard -- agent/ web/ 2>/dev/null) || true
         git checkout main 2>/dev/null || true
         # Delete local branch if it was never pushed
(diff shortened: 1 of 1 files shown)


PavelStancik:
Addressed the review feedback and #287:

1. **Snapshot timing** — `_PRE_UNTRACKED` is now populated immediately after `source` lines (before the trap can fire), not just before branch creation. A re-snapshot is taken after `git stash` in case stashing changes the untracked set.

2. **Subshell overhead** — Replaced `echo | grep -qxF` with pure-bash `[[ ... == *...* ]]` pattern matching to avoid forking two processes per untracked file.

Syntax validated with `bash -n`. The early-exit data-loss path from #287 is now closed.

— Marvin

PR #301, opened 2026-03-24
fix: extend self-enhance validation + close webhook SSRF rebinding gap

### LINKED ISSUE #284: bug: self-enhance rollback only validates agent/*.sh — changes to web/ or new files outside agent/ bypass validation
## Issue

Introduced in PR #272 (`agent/self-enhance.sh`, `_validate_post_enhance` function).

The post-enhancement validator only scans bash scripts in `agent/`:

```bash
done < <(find "${MARVIN_DIR}/agent" -name "*.sh" -type f)
```

If Claude modifies files outside `agent/` (e.g., `web/` TypeScript files, `setup/` scripts, new scripts placed in the repo root), those changes:
1. Are **not validated** by the syntax/conflict-marker checks
2. Will **not trigger rollback** even if they contain errors

This means a self-enhancement that breaks the Next.js dashboard or creates a malformed script outside `agent/` would be committed without rollback protection.

## Suggested fix

Extend validation to include all shell scripts in the repo and add a TypeScript compile check for `web/`:

```bash
# Check all .sh files in the entire repo
done < <(find "${MARVIN_DIR}" -name "*.sh" -type f -not -path "*/.git/*" -not -path "*/node_modules/*")

# Optionally: TypeScript type-check
if [[ -d "${MARVIN_DIR}/web" ]] && command -v npx &>/dev/null; then
    if ! npx --prefix "${MARVIN_DIR}/web" tsc --noEmit 2>/dev/null; then
        marvin_log "ERROR" "Post-enhance validation FAILED: TypeScript errors in web/"
        valid=false
    fi
fi
```

## Severity

🟠 Warning — rollback safety guarantee is incomplete for enhancements that touch files outside `agent/`.

### LINKED ISSUE #299: security: TOCTOU DNS rebinding gap in webhook SSRF protection (log-export.sh)
## Problem

PR #294 adds SSRF protection by calling `getent hosts` to resolve the webhook hostname and checking the IP against private ranges. However, `curl` performs its own DNS lookup when the request is made — a separate call that happens *after* the validation check.

An attacker controlling a DNS server with TTL=0 can serve a public IP during `getent hosts` (passing validation), then immediately flip the DNS record to a private/internal IP for the subsequent `curl` lookup. This is the classic **DNS rebinding** attack.

**Affected file:** `agent/log-export.sh`

## Reproduction scenario

1. Attacker controls `attacker.example.com` DNS with TTL=0
2. At validation time (`getent hosts`): DNS returns `1.2.3.4` (public) → passes check
3. At request time (`curl`): DNS returns `127.0.0.1` → SSRF to localhost

## Fix

Use `curl --resolve hostname:port:pre-validated-ip` to force curl to use the already-validated IP, bypassing its own DNS lookup:

```bash
resolved_ip=$(getent hosts "${webhook_host_bare}" 2>/dev/null | awk '{print $1; exit}')
# ... validation checks ...
# Force curl to use the pre-validated IP
curl --resolve "${webhook_host_bare}:443:${resolved_ip}" --resolve "${webhook_host_bare}:80:${resolved_ip}" ...
```

Alternatively, build the URL with the resolved IP directly.

## Severity

🟠 Warning — bypassing this requires attacker-controlled DNS with very short TTL. In practice, webhook URLs are configured by the operator, so exploitation requires a compromised or malicious webhook config. Low exploitation probability, but should be fixed for defense-in-depth.

*Found during PR #294 code review.*

RobotMarvin2026:
## Summary

- **self-enhance.sh** (`_validate_post_enhance`): Now checks ALL `.sh` files repo-wide (not just `agent/`) and detects conflict markers in `web/` source files (JS/TS/JSX/TSX/JSON/CSS). Previously, broken scripts in `setup/` or web/ changes with conflict markers bypassed rollback protection. Fixes #284.

- **log-export.sh** (webhook delivery): Added `curl --resolve` to pin the pre-validated IP address, closing the TOCTOU DNS rebinding window where a TTL=0 DNS record could flip from public to private IP between `getent hosts` validation and the actual `curl` request. Fixes #299.

## Test plan

- [ ] Verify `bash -n` passes on all modified scripts
- [ ] Verify self-enhance rollback triggers correctly when a web/ file has conflict markers
- [ ] Verify webhook delivery still works with `--resolve` pinning

🤖 Self-enhancement session — 2026-03-24


RobotMarvin2026:
Fixed the non-standard port gap flagged in #303 (commit eb2c995).

The `--resolve` pinning now extracts the actual port from the webhook URL instead of hardcoding 80/443. For `http://example.com:8080/hook`, curl gets `--resolve example.com:8080:<validated-ip>` — so the DNS rebinding window is closed regardless of port.

Also added `-not -path "*/data/*"` to the `.sh` file validation in `self-enhance.sh` per the review suggestion, preventing false-positive rollbacks from any runtime-generated scripts in `data/`.

Both scripts pass `bash -n`.

— Marvin

PR #531, opened 2026-04-09
feat: modular prompt system with shared modules

RobotMarvin2026:
## Summary

- **New:** `agent/lib/prompts.sh` — prompt assembly library with `marvin_build_prompt()`, `marvin_load_modules()`, `marvin_list_modules()`
- **New:** `agent/prompts/modules/` — 3 shared modules (identity.md, security-rules.md, output-rules.md) extracted from content duplicated across 12 prompt files
- **Changed:** `self-enhance.sh` uses modular assembly as proof of concept
- **Changed:** `enhance.md` slimmed down — references modules instead of duplicating identity/security content

## Why

Identity, security rules, and output guidelines were copy-pasted across 8-12 prompts each. Changes required editing every file. This creates a single source of truth for shared content while keeping task-specific prompts lean. Other scripts can adopt modules incrementally.

## Risk

Low. Only self-enhance.sh uses the new system; all other scripts still use `cat` directly. Assembled prompt content is identical. All scripts pass `bash -n`.

## Test plan

- [x] All .sh files pass syntax check
- [x] `marvin_build_prompt "enhance" identity security-rules` produces correct output
- [x] Assembled prompt contains identity, security, enhancement task, and philosophy sections
- [x] `marvin_list_modules` lists all 3 modules


PavelStancik (approved):

RobotMarvin2026:
Pushed `5fbec1f` — added task name validation to `marvin_build_prompt()`, addressing the review finding in #533.

```bash
# Added at the top of marvin_build_prompt():
if [[ ! "$task" =~ ^[a-zA-Z0-9_-]+$ ]]; then
    marvin_log "ERROR" "Invalid task name: ${task}" >&2
    return 1
fi
```

Same regex guard as `marvin_load_module()`. All callers use hardcoded strings today, but this closes the path traversal vector before other scripts adopt the API.

Both review findings are now addressed:
- #532 — hard-fail on missing modules (fixed in `c30cd5c`)
- #533 — task name validation (fixed in `5fbec1f`)

All scripts pass `bash -n`. Ready to merge.

— Marvin

PR #539, opened 2026-04-11
fix: #526 Warning: PID reuse can cause false "build in progress" lock 

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #526 — Warning: PID reuse can cause false "build in progress" lock block in marvin_rebuild_web()

**Fix:** Added PID-reuse detection to the build lock in marvin_rebuild_web(). The lock now records the process start time (from /proc/<pid>/stat field 22) alongside the PID, and compares it when checking lock ownership — if the start times differ, the lock is treated as orphaned and reclaimed instead of falsely blocking the build.

### Changed Files
```
CHANGELOG.md
agent/common.sh
```

### Validation
- [x] GPG-signed commit

---
*Fixes #526*


RobotMarvin2026:
Addressed the review suggestion in 0f160a5:

- Replaced `|| true` with `|| marvin_log "WARN" "Could not record start time for lock — PID reuse detection disabled"` — the degradation is now visible in logs rather than silently swallowed.

The build still proceeds normally in either case. The only difference is operational visibility.

— Marvin

RobotMarvin2026:
Addressed the atomic write suggestion from the second review in cfe6c03:

- Start-time file now writes to `start.tmp` then `mv` — atomic on Linux, prevents concurrent readers from seeing a partial or empty file during the write window.
- Variable style: checked all `lock_pid` uses — they already use `${lock_pid}` in log message strings. The bare `$lock_pid` in conditionals (`[[ -z "$lock_pid" ]]`, `kill -0 "$lock_pid"`) is standard shell practice and doesn't need braces.

— Marvin

PavelStancik (approved):

PR #546, opened 2026-04-13
fix: #545 health-monitor.sh: silent skip when proc_exe is unreadable f

### LINKED ISSUE #545: health-monitor.sh: silent skip when proc_exe is unreadable for find/git exclusion
## Warning: Silent suppression of unreadable exe path in runaway monitor

**Affects:** `agent/health-monitor.sh` — `find|git` case in the runaway process loop

### Description

In the `find|git` branch, when `/proc/<pid>/exe` cannot be read (`proc_exe` is empty), the code silently calls `continue` without any log entry:

```bash
find|git)
    if [[ -z "$proc_exe" ]]; then
        continue   # ← no warning logged
    fi
```

A high-CPU process that cannot have its executable path determined should generate a `WARN` log, not be silently skipped. An attacker or misbehaving process could exploit unreadable `/proc/<pid>/exe` (e.g., due to namespace differences, permission issues, or a race with process exit) to avoid detection by the runaway monitor.

This behaviour pre-existed for `find` and was extended to `git` in PR #543.

### Suggested fix

Replace the silent `continue` with a log + continue:

```bash
if [[ -z "$proc_exe" ]]; then
    marvin_log "WARN" "Cannot read exe for allowlisted ${proc_name} at ${proc_cpu}% CPU (PID ${proc_pid}) — skipping"
    continue
fi
```

### Severity

Warning — not directly exploitable on its own, but silently suppressing monitoring events reduces detection capability.

*Identified during code review of PR #543.*

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #545 — health-monitor.sh: silent skip when proc_exe is unreadable for find/git exclusion

**Fix:** Added a WARN log before the silent `continue` when `/proc/<pid>/exe` is unreadable for an allowlisted `find` process, so high-CPU processes with unresolvable exe paths are no longer silently ignored.

### Changed Files
```
CHANGELOG.md
agent/health-monitor.sh
```

### Validation
- [x] GPG-signed commit

---
*Fixes #545*


RobotMarvin2026:
## Review suggestion implemented — `fc57c01`

Adopted the reviewer's recommendation: removed the `continue` so unreadable-exe processes fall through to normal runaway CPU detection instead of being silently exempted.

### Before
```bash
if [[ -z "$proc_exe" ]]; then
    marvin_log "WARN" "Cannot read exe for allowlisted ... — skipping"
    continue   # ← still exempted from runaway check
fi
```

### After
```bash
if [[ -z "$proc_exe" ]]; then
    marvin_log "WARN" "Cannot read exe for allowlisted ... — treating as unverified"
    break      # ← exits case, falls through to runaway reporting
fi
```

The process is now both logged as suspicious (WARN) *and* reported as a high-CPU event. CHANGELOG updated to reflect the behavior change.

— Marvin

PR #549, opened 2026-04-13
fix: remove break that aborts CPU monitoring loop (#547, #548)

### LINKED ISSUE #547: bug: health-monitor.sh `break` in PR #546 aborts entire CPU monitoring loop
## Bug Report

**Introduced by:** PR #546 (fix for #545)  
**File:** `agent/health-monitor.sh`, line 325  
**Severity:** 🔴 Critical

## Description

PR #546 replaced a `continue` with `break` in the `find` case of the high-CPU process monitoring loop. The intent (stated in both the CHANGELOG and inline comment) was to "fall through to normal runaway detection instead of skipping." However, `break` in a `while` loop exits the loop entirely — it does **not** fall through.

## Impact

When the first high-CPU `find` process has an unreadable `/proc/<pid>/exe`, the `break` terminates the outer `while IFS= read -r line` loop. All remaining high-CPU processes in the scan are silently dropped — never logged, never tracked. This is a **regression worse than the original bug** (#545): before, one process could evade detection via an unreadable exe; now, all subsequent processes evade detection entirely.

## Code Structure

```bash
while IFS= read -r line; do       # outer loop — iterates over high-CPU processes
    ...
    case "$proc_name" in
        find)
            if [[ -z "$proc_exe" ]]; then
                marvin_log "WARN" "..."
                break              # ← exits the while loop, NOT fall-through
            fi
            ...
        ;;
    esac
    # runaway detection code (jq tracking, alerting) — never reached after break
done
```

## Correct Fix

Remove the `break` entirely. The code naturally falls through the `if` block → `fi` → `if [[ "$proc_exe" == /usr/bin/find ]]` (false, since exe is empty) → `else` WARN log → `;;` → `esac` → runaway detection at line 343+.

```bash
if [[ -z "$proc_exe" ]]; then
    marvin_log "WARN" "Cannot read exe for allowlisted ${proc_name} at ${proc_cpu}% CPU (PID ${proc_pid}) — treating as unverified"
    # No break, no continue — fall through to runaway detection naturally.
fi
```

Note: this will trigger a second (redundant) WARN from the `else` branch at line 337 ("Untrusted exe for allowlisted name"). To avoid that, restructure the inner conditions as `if/elif/else` rather than two independent `if` blocks.

### LINKED ISSUE #548: health-monitor.sh: PR #546 uses `break` instead of fall-through, exits monitoring loop prematurely
## Bug Report

**Source:** Code review of PR #546

**File:** `agent/health-monitor.sh`

### Description

PR #546 changes the empty-`proc_exe` early-exit in the `find` case from `continue` to `break`, with the comment _"Fall through to normal runaway detection instead of skipping."_ These two behaviors are opposite:

| Statement | Effect |
|-----------|--------|
| `continue` | Skip to next iteration of `while` loop (old behaviour — silently ignores this process) |
| `break` | **Exit the `while` loop entirely** — stops all further process monitoring for this scan cycle |
| _(neither)_ | Falls through to the existing `else` branch: logs "Untrusted exe for allowlisted name" (intended behaviour) |

### Security Impact

When a high-CPU `find` process has an unreadable `/proc/<pid>/exe`, the monitoring loop exits immediately. All remaining processes in the same scan cycle are **not checked**.

A malware process named `find` that unlinks its own binary after `exec()` (a standard anti-forensics technique) would reliably trigger this path, silencing the monitor for the rest of that cycle — the opposite of the stated security improvement.

### Correct Fix

Remove the early `continue` **without** adding `break`. The code then falls through to the existing `else` branch:

```bash
if [[ -z "$proc_exe" ]]; then
    marvin_log "WARN" "Cannot read exe for allowlisted ${proc_name} at ${proc_cpu}% CPU (PID ${proc_pid}) — treating as unverified"
    # No break/continue — fall through to else branch below
fi
if [[ "$proc_exe" == /usr/bin/find ]]; then
    # ... parent verification ...
else
    marvin_log "WARN" "Untrusted exe for allowlisted name: ${proc_name} (PID ${proc_pid}, exe=${proc_exe}) at ${proc_cpu}% CPU"
fi
```

This logs the "unreadable exe" warning AND the "untrusted exe" warning, and does not interrupt monitoring of subsequent processes.

### Note

PRs #543 and #546 both modify the same section of `health-monitor.sh` and will conflict on merge. They should be sequenced — fix this `break` bug before merging either.

RobotMarvin2026:
## Summary

- PR #546 replaced `continue` with `break` in the `find|git` case of the high-CPU process monitoring loop (line 326 of `health-monitor.sh`). The intent was to fall through to runaway detection when `/proc/<pid>/exe` is unreadable, but `break` exits the `while` loop entirely — silently dropping all remaining high-CPU processes from that scan cycle.
- Removed the `break`. The code now falls through naturally: logs the "unverified" warning, then hits the `_expected_exe` check (fails since `proc_exe` is empty), logs "Untrusted exe", and continues past `esac` into runaway detection.
- This is a security regression fix — a malware process named `find` that unlinks its binary after exec would reliably silence the monitor for the rest of the cycle.

Closes #547
Closes #548

## Test plan

- [ ] Verify `bash -n agent/health-monitor.sh` passes (done locally)
- [ ] Confirm the monitoring loop continues past processes with unreadable `/proc/<pid>/exe`
- [ ] Check that both WARN messages ("Cannot read exe" + "Untrusted exe") appear in logs for unverifiable processes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ## [Unreleased]
 
+### Fixed
+
+- **`break` → fall-through in `find|git` CPU monitoring** (`health-monitor.sh`) — PR #546 replaced `continue` with `break` in the `find|git` case when `/proc/<pid>/exe` is unreadable, intending to fall through to runaway detection. However, `break` exits the entire `while` loop, silently dropping all remaining high-CPU processes from that scan cycle. Removed the `break` so the code falls through naturally: logs the "unverified" warning, then the "untrusted exe" warning, then continues to runaway detection. Fixes #547, #548.
+
 ### Added
 
 - **`git` added to runaway process exclusions** (`health-monitor.sh`) — Git operations (fetch, pull, push, rebase) spike to 100% CPU during morning-check and github-interact cron jobs. Like `find`, uses parent-process verification: only suppresses `git` from `/usr/bin/git` whose parent is a Marvin bash script. Non-Marvin git processes are still monitored and logged.
--- a/agent/health-monitor.sh
+++ b/agent/health-monitor.sh
@@ -300,9 +300,11 @@ while IFS= read -r line; do
     proc_exe=$(readlink -f "/proc/${proc_pid}/exe" 2>/dev/null || echo "")
     case "$proc_name" in
         claude|apt*|dpkg*|ps|jq|fail2ban*|file|appstreamcli)
+            # High-frequency, low-risk short-lived children — silent skip when
+            # exe is unreadable (they exit between ps and readlink constantly).
+            # Contrast with find|git below which logs + falls through, because
+            # those run longer and an unreadable exe is more suspicious there.
             if [[ -z "$proc_exe" ]]; then
-                # Process exited between ps and readlink — can't verify, skip silently.
-                # Short-lived children (e.g., rkhunter's `file`) hit this constantly.
                 continue
             fi
             if [[ "$proc_exe" == /usr/bin/* || "$proc_exe" == /usr/sbin/* || \
--- a/agent/health-monitor.sh
+++ b/agent/health-monitor.sh
@@ -323,7 +323,9 @@ while IFS= read -r line; do
                 # Fall through to normal runaway detection instead of skipping.
                 # An unreadable /proc/<pid>/exe could indicate namespace tricks,
                 # permission issues, or a race — silent exemption reduces coverage.
-                break
+                # NOTE: No break/continue here — proc_exe="" falls through to the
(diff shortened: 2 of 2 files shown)


RobotMarvin2026:
Addressed the review suggestion in c3adcb9 — added an inline comment on the `claude|apt*` branch explaining the asymmetry:

- **`claude|apt*|...`**: High-frequency, low-risk short-lived children. Silent skip on unreadable exe because they exit between `ps` and `readlink` constantly.
- **`find|git`**: Longer-running, parent-verified. Log + fall through on unreadable exe because it's more suspicious and worth tracking.

The contrast is now documented in both branches for future reviewers.

— Marvin

PR #560, opened 2026-04-15
fix: verify Docker GPG key fingerprint after download (#558)

### LINKED ISSUE #558: security: Docker GPG key downloaded without fingerprint verification in monitoring/setup.sh
## Issue

In `monitoring/setup.sh`, the Docker GPG key is downloaded without verifying its fingerprint:

```bash
curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
```

This is a trust-on-first-use (TOFU) pattern. If the connection is MITMed or Docker's CDN is compromised, a malicious signing key could be installed, allowing arbitrary packages to be installed via `apt` as root.

The `.github/CLAUDE.md` review checklist explicitly flags: **"No `curl | bash` patterns without verification"** — downloading an unverified GPG signing key carries the same trust chain risk.

## Recommended Fix

After downloading, verify the key fingerprint before trusting it:

```bash
curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc

# Verify fingerprint (Docker's official key fingerprint)
EXPECTED_FINGERPRINT="9DC858229FC7DD38854AE2D88D81803C0EBFCD88"
ACTUAL_FINGERPRINT=$(gpg --show-keys --fingerprint /etc/apt/keyrings/docker.asc 2>/dev/null | grep -A1 "pub" | tail -1 | tr -d ' ')
if [[ "${ACTUAL_FINGERPRINT}" != "${EXPECTED_FINGERPRINT}" ]]; then
    echo "ERROR: Docker GPG key fingerprint mismatch! Got: ${ACTUAL_FINGERPRINT}"
    exit 1
fi
```

## Context

- `monitoring/setup.sh` runs as root
- Introduced in PR #551
- Severity: 🟠 Warning

RobotMarvin2026:
## Summary

- **Verify Docker GPG key fingerprint** in `monitoring/setup.sh` — after downloading the key via `curl`, the fingerprint is now checked against Docker's official fingerprint (`9DC858229FC7DD38854AE2D88D81803C0EBFCD88`). If it doesn't match, the key is removed and the script aborts with an error. Prevents MITM or CDN compromise from installing a rogue signing key.

Fixes #558

## Test plan

- [ ] Verify `bash -n monitoring/setup.sh` passes
- [ ] Review fingerprint matches Docker's official key
- [ ] Confirm error path removes the key file and exits non-zero


RobotMarvin2026:
Implemented the gnupg suggestion from the automated review. Pushed commit `4a3e9a1`:

- Added `gnupg` to the `apt-get install` line before the GPG fingerprint verification step

On minimal Docker/container images where `gpg` might not be pre-installed, the script will now explicitly install it. On standard Ubuntu 24.04 this is a no-op (already present), but it makes the dependency explicit rather than assumed.

— Marvin

PavelStancik (approved):

PR #561, opened 2026-04-15
fix: #559 monitoring: nc dependency in common.sh may silently disable 

### LINKED ISSUE #559: monitoring: nc dependency in common.sh may silently disable OTEL telemetry
## Issue

In `agent/common.sh` (PR #551), the OTEL environment variables are only exported when `nc` can reach the collector:

```bash
if nc -z 127.0.0.1 4317 2>/dev/null; then
    export CLAUDE_CODE_ENABLE_TELEMETRY=1
    ...
fi
```

If `nc` (netcat) is not installed on the system, the check fails silently — `nc: command not found` is suppressed by `2>/dev/null`, the `if` evaluates to false, and OTEL is **never enabled** even when the collector is running. This makes it impossible to diagnose why telemetry isn't working.

On Ubuntu, `nc` is provided by `netcat-openbsd` — it's usually present but not guaranteed in all minimal images.

## Recommended Fix

Use `/dev/tcp` bash built-in as a fallback (no external dependency):

```bash
_otel_reachable() {
    # Try nc first, fall back to bash /dev/tcp
    if command -v nc &>/dev/null; then
        nc -z 127.0.0.1 4317 2>/dev/null
    else
        (echo >/dev/tcp/127.0.0.1/4317) 2>/dev/null
    fi
}

if _otel_reachable; then
    export CLAUDE_CODE_ENABLE_TELEMETRY=1
    ...
fi
```

Or at minimum, add a startup warning if `nc` is not found:

```bash
if ! command -v nc &>/dev/null; then
    warn "nc not found — OTEL telemetry check skipped, monitoring disabled"
elif nc -z 127.0.0.1 4317 2>/dev/null; then
    ...
fi
```

## Context

- Introduced in PR #551 (fixes issue #557)
- Severity: 🟠 Warning

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #559 — monitoring: nc dependency in common.sh may silently disable OTEL telemetry

**Fix:** Added _otel_reachable() helper that tries nc first and falls back to bash's /dev/tcp built-in, so OTEL telemetry is enabled even when netcat is not installed on the system.

### Changed Files
```
CHANGELOG.md
agent/common.sh
```

### Validation
- [x] GPG-signed commit

---
*Fixes #559*


RobotMarvin2026:
Implemented the timeout suggestion from the automated review. Pushed commit `dbbefa3`:

- `nc` now gets `-w 1` (1-second connection timeout)
- `/dev/tcp` fallback wrapped in `timeout 1 bash -c '...'`

Both branches now fail fast on filtered ports instead of potentially blocking. The pre-existing gap is closed.

— Marvin

PavelStancik (approved):

PR #565, opened 2026-04-15
fix: #564 health-monitor: kill -0 liveness check conflates ESRCH and E

### LINKED ISSUE #564: health-monitor: kill -0 liveness check conflates ESRCH and EPERM, silently dropping non-root find/git from runaway detection
## Summary

In `agent/health-monitor.sh` (introduced in PR #562), the `find|git` case uses `kill -0 "$proc_pid"` to distinguish dead processes (harmless race) from live processes with an unreadable exe path (suspicious).

**The problem:** `kill -0` returns non-zero for **two** distinct situations:
- `ESRCH` — process does not exist (what we want to detect as "dead")
- `EPERM` — process exists but is owned by a different user (we lack signal permission)

Since both failures are caught by `! kill -0 "$proc_pid" 2>/dev/null`, a high-CPU `find` or `git` process owned by **root or another user** (where `marvin` lacks signal permission) is silently `continue`d instead of being logged and falling through to runaway detection.

## Impact

- Non-Marvin `find`/`git` processes at high CPU that were previously WARN-logged and tracked by runaway detection are now silently skipped if they belong to another user.
- This reduces security monitoring coverage for foreign-user processes in the allowlisted-name case block.

## Reproduction scenario

1. A root-owned `find /` or `git gc` runs at >80% CPU for >10 minutes
2. Marvin's monitoring script (running as `marvin` user) picks it up via `ps`
3. `readlink /proc/<pid>/exe` fails (EACCES) → `proc_exe` is empty
4. `kill -0 <pid>` fails with EPERM → condition is true → `continue`
5. Process is never logged, never tracked in `RUNAWAY_FILE`, never killed

## Suggested Fix

Replace `kill -0` with a filesystem check that works across ownership boundaries:

```bash
if [[ ! -d "/proc/$proc_pid" ]]; then
    continue  # Process already exited — harmless race condition
fi
```

`/proc/<pid>` exists for any running process regardless of ownership — no signal permission required. This correctly distinguishes dead (dir absent) from live-but-unreadable-exe (dir present).

## Affected file

`agent/health-monitor.sh` lines 323–332

## References

- PR #562 (introduced the change)
- PR #547 (original break→fall-through fix that motivated this)

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #564 — health-monitor: kill -0 liveness check conflates ESRCH and EPERM, silently dropping non-root find/git from runaway detection

**Fix:** Replaced `kill -0` with `/proc/$pid` directory existence checks in both the find/git liveness test (line 328) and stale PID cleanup (line 392). The /proc check works regardless of process ownership, fixing the ESRCH/EPERM conflation that caused non-root runs to silently skip live processes.

### Changed Files
```
CHANGELOG.md
agent/health-monitor.sh
```

### Validation
- [x] GPG-signed commit

---
*Fixes #564*


PavelStancik (approved):

RobotMarvin2026:
Addressed both suggestions from the automated review:

1. **PID sanitization** — Added `[[ "$pid" =~ ^[0-9]+$ ]] || continue` guard before `/proc/${pid}` path construction in the stale PID cleanup loop. Non-numeric jq keys now skip silently instead of resolving to unintended paths.

2. **CHANGELOG reference** — Updated the #547 entry to note that the original `kill -0` approach was subsequently replaced with `/proc` dir check in this PR.

Both changes in commit 754707f.

— Marvin

PR #609, opened 2026-04-19
fix: #606 Kernel: recurring RCU preempt stalls on CPU 0 (2 in 63 min)

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #606 — Kernel: recurring RCU preempt stalls on CPU 0 (2 in 63 min)

**Fix:** Wrapped the single `claude -p` invocation in `run_claude()` (agent/lib/claude.sh) with `nice -n 10 ionice -c 2 -n 7` so Claude runs at lowered CPU/IO priority, leaving headroom for `rcu_preempt` and other kernel threads on the 2-vCPU box. Addresses the cron-density/kthread-starvation hypothesis called out in the issue.

### Changed Files
```
CHANGELOG.md
agent/lib/claude.sh
```

### Validation
- [x] All bash scripts pass `bash -n` syntax check
- [x] No merge conflict markers in changed files
- [x] No data/runtime files modified
- [x] GPG-signed commit

---
*Automated fix by Marvin's issue-fixer agent.*

Closes #606

PR #612, opened 2026-04-20
fix: #611 Warning: ionice not guarded in run_claude() — missing packag

### LINKED ISSUE #611: Warning: ionice not guarded in run_claude() — missing package breaks all Claude invocations
## Summary

PR #609 wraps the `claude -p` invocation in `nice -n 10 ionice -c 2 -n 7`. If `ionice` (`util-linux`) is not installed, **every call to `run_claude()` will fail** with a "command not found" error, silently breaking all Claude-dependent cron jobs — the exact failure mode that PR #609 and PR #610 are trying to prevent.

On Ubuntu 24.04 `util-linux` is included by default and `ionice` should be present. But the dependency is implicit and unguarded.

## Recommendation

Add a guard in `run_claude()` (or in `agent/common.sh` bootstrap checks) so that if `ionice` is absent, the invocation degrades gracefully to `nice -n 10 claude -p`:

```bash
if command -v ionice &>/dev/null; then
    printf '%s' "${full_prompt}" | nice -n 10 ionice -c 2 -n 7 claude -p > "$output_file" 2>&1 && exit_code=$? || exit_code=$?
else
    printf '%s' "${full_prompt}" | nice -n 10 claude -p > "$output_file" 2>&1 && exit_code=$? || exit_code=$?
fi
```

Or check once at script startup and log a WARN if `ionice` is unavailable.

## Risk

Low probability (Ubuntu 24.04 ships `util-linux`), but the blast radius if it occurs is total: **all** Claude invocations fail, disabling every cron agent.

Raised from review of PR #609.

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #611 — Warning: ionice not guarded in run_claude() — missing package breaks all Claude invocations

**Fix:** Added a load-time bootstrap check in agent/lib/claude.sh that emits a WARN via marvin_log if `ionice` (util-linux) or `nice` (coreutils) is missing from PATH. Implements the issue's recommended startup-check approach so a missing package becomes visible in logs before PR #609's `nice -n 10 ionice -c 2 -n 7 claude -p` pipeline runs and silently breaks every cron agent.

### Changed Files
```
CHANGELOG.md
agent/lib/claude.sh
```

### Validation
- [x] GPG-signed commit

---

Closes #611


RobotMarvin2026:
Pushed two new commits addressing the review:

- `43c9385` — original #611 load-time WARN, rebased onto main (PR #609 merged in the interim, so this picked up the `nice -n 10 ionice -c 2 -n 7` line cleanly).
- `384eebb` — addresses **#613** (graceful fallback). `run_claude()` now builds the priority prefix as an array, appending `nice -n 10` and `ionice -c 2 -n 7` only when each binary is on PATH. If both are absent the array is empty and `claude -p` runs unwrapped instead of aborting with "command not found". Priority lowering becomes opportunistic, not mandatory.

Verified empty-array expansion under `set -euo pipefail` works in bash 5. Issue #613 closes when this PR merges.

On the `>&2` suggestion: I left it. Every `marvin_log` call in this file uses `>&2` (lock-acquire/release, prompt size, exit-code warnings, etc.) — removing it from these two lines would make them inconsistent with the rest of the file, not more so. The `>&2` redirects any incidental stdout to stderr so it doesn't leak into command substitution callers; that's the established pattern here.

— Marvin

PR #619, opened 2026-04-21
fix: #618 security-scan.sh: `high_rate_count` multi-line value produce

### LINKED ISSUE #618: security-scan.sh: `high_rate_count` multi-line value produces invalid JSON in latest-scan.json
## Bug

`security-scan.sh` occasionally writes `$high_rate_count` as a multi-line value into `data/security/latest-scan.json`, producing invalid JSON that breaks downstream consumers.

## Observed

`self-test.sh` fails at line 322 with exit 5 trying to parse `data/security/latest-scan.json`:

```
[ERROR] self-test.sh:322 — command failed (exit 5): scan_status=$(jq -r '.overall_status // "unknown"' "$LATEST_SCAN" 2>/dev/null)
```

`cat -A data/security/latest-scan.json` around line 35:

```
"established_connections": 6,
"suspicious_connections": 6,
"high_rate_ips": 0
0,
"high_rate_threshold": 50,
```

`$high_rate_count` expanded to `0\n0` in the heredoc template, so the JSON has `0` on one line and the closing `,` on the next.

## Suspected root cause

`security-scan.sh:281-282`:

```bash
high_rate_count=$(echo "$top_sources_json" | jq --argjson thr "$HIGH_CONN_THRESHOLD" \
    '[.[] | select(.connections > $thr)] | length' 2>/dev/null || echo 0)
```

This looks like a `jq ... 2>/dev/null || echo 0` hybrid-output case where jq succeeds and writes `0` to stdout AND then fails on a later stage (or the pipeline produces two outputs), triggering the `|| echo 0` fallback and appending another `0` on its own line. `$()` preserves both.

## Impact

- `self-test.sh` exits early, skipping the security-score computation and rollup.
- Dashboard consumers of `latest-scan.json` (and the restricted `/api/security/` endpoints for internal use) see broken JSON.
- Likely intermittent — depends on the `ss` output patterns during the scan window.

## Suggested fix

Sanitize the capture to a single line:

```bash
high_rate_count=$(echo "$top_sources_json" \
    | jq --argjson thr "$HIGH_CONN_THRESHOLD" '[.[] | select(.connections > $thr)] | length' 2>/dev/null \
    | head -1 || echo 0)
high_rate_count=${high_rate_count:-0}
```

Or a stricter variant: conditional on jq success explicitly (`if cmd; then ... else ... fi`) instead of the `||` fallback that can splice outputs.

Related: `suspicious_count`, `outbound_unexpected`, and similar jq-captured counts in the same file should be audited for the same pattern.

## Priority

Stability / data — medium. Breaks daily self-test telemetry but nothing user-facing fails loudly. Caught by today's self-enhance session.

### LINKED ISSUE #621: security-scan.sh: `head -n 1` + pipefail SIGPIPE can silently zero `high_rate_count` / `suspicious_count`
## Summary

PR #619 fixes multi-line jq values breaking `latest-scan.json` by adding `| head -n 1` to the capture pipeline. However, with `set -o pipefail` (which `security-scan.sh` uses), this introduces a new silent failure mode in the edge case that triggered the original bug.

## Reproduction path

When `$top_sources_json` or `$suspicious_conns` is multi-document JSON (the root cause of #618), jq outputs 2+ lines:

```
5
0
```

`head -n 1` reads `5` and exits 0. Jq then tries to write the second line, receives **SIGPIPE**, and exits 141. With `set -o pipefail`, the pipeline exit code is 141 (non-zero), so `|| echo 0` fires and `high_rate_count` is set to `0` instead of `5`.

## Impact

`high_rate_count=0` suppresses the `WARN "Found N IPs exceeding rate threshold"` log and the IP-flagging block that follows. In the multi-document-jq edge case a real high-rate abuse event would be silently missed in that scan cycle.

## Recommended fix

Wrap the `|| echo 0` fallback in a subshell so it is evaluated before the pipe to `head`, keeping SIGPIPE from contaminating the exit code:

```bash
high_rate_count=$(( echo "$top_sources_json" | jq --argjson thr "$HIGH_CONN_THRESHOLD" \
    '[.[] | select(.connections > $thr)] | length' 2>/dev/null || echo 0 ) | head -n 1)
high_rate_count=${high_rate_count:-0}
```

Same pattern for `suspicious_count`.

## Context

- Introduced by PR #619 (fixes #618)
- Only triggers in the already-pathological case where jq receives multi-document input
- In the normal single-document case the current fix is correct and unaffected

RobotMarvin2026:
## Automated Issue Fix

**Issues:** #618 (multi-line value corrupts JSON) and #621 (SIGPIPE + pipefail silent zeroing).

**Fix:** Sanitize `high_rate_count` and `suspicious_count` captures in `security-scan.sh` using bash parameter expansion (`${var%%$'\n'*}`) to trim everything after the first newline, plus a numeric regex guard (`[[ "$var" =~ ^[0-9]+$ ]] || var=0`). No pipe to `head -n 1`, so the SIGPIPE + `set -o pipefail` interaction flagged in #621 cannot silently zero a real count. The `|| echo 0` fallback still runs if jq fails; the trim then strips any appended `0` from the fallback, preserving the real value when jq produced partial output before a non-zero exit.

### Changed Files
```
CHANGELOG.md
agent/security-scan.sh
```

### Validation
- [x] GPG-signed commit
- [x] Latest automated review (2026-04-22 00:11 UTC) approved with no Critical / Warning findings

---

Closes #618
Closes #621


RobotMarvin2026:
Rebased onto main to resolve the CHANGELOG merge conflict that appeared after PR #616 landed. Both CHANGELOG entries (the #615 claude.sh guard and this #618 fix) now coexist cleanly. GPG-signed commit `7983b95`. Ready to merge once reviewed.

PR #674, opened 2026-05-04
fix: #673 Grafana 'failed to load application files' (nginx rate limit on static assets)

### LINKED ISSUE #673: Grafana
Grafana is not working, monitor it also
a page content now:
```
If you're seeing this Grafana has failed to load its application files
This could be caused by your reverse proxy settings.
If you host grafana under a subpath make sure your grafana.ini root_url setting includes subpath. If not using a reverse proxy make sure to set serve_from_sub_path to true.
If you have a local dev build make sure you build frontend using: yarn start, or yarn build.
Sometimes restarting grafana-server can help.
Check if you are using a non-supported browser. For more information, refer to the list of [supported browsers ](https://grafana.com/docs/grafana/latest/installation/requirements/#supported-web-browsers).

```

RobotMarvin2026:
Closes #673.

## Root cause

The `sensitive` rate-limit zone (2r/s, burst=5) was applied at server level in `setup/nginx-monitoring.conf`, covering Grafana's static assets. A single Grafana page load fetches 7+ JS/CSS bundles in parallel — the burst was exhausted, the JS bundle 429'd, and the user got the *"Grafana has failed to load its application files"* fallback page.

Visible in `/var/log/nginx/error.log`:

```
limiting requests, excess: 5.850 by zone "sensitive", ... request: "GET /public/img/grafana_icon.svg"
limiting requests, excess: 5.840 by zone "sensitive", ... request: "GET /public/build/9890.<hash>.js"
limiting requests, excess: 5.824 by zone "sensitive", ... request: "GET /public/build/app.<hash>.js"
```

## Fix

Move the `limit_req` directive from server level into `location /` only. Add `/public/` and `/avatar/` location blocks (auth still required, no rate limit) — these are static assets, not the brute-force vector.

The brute-force protection on the basic-auth challenge for dynamic endpoints (`/`, `/login`, `/api/*`) is preserved.

## Verified live

Already deployed (`cp` → `nginx -t` → `systemctl reload nginx`); this PR brings the source of truth in line with what is running.

```
$ for i in {1..10}; do curl -sk -o /dev/null -w "%{http_code} " \
    https://monitoring.robot-marvin.cz/public/build/app.js; done
401 401 401 401 401 401 401 401 401 401     # no 429 — limit bypassed for /public/

$ for i in {1..10}; do curl -sk -o /dev/null -w "%{http_code} " \
    https://monitoring.robot-marvin.cz/login; done
401 401 401 401 401 401 429 429 429 429     # limit still active on /login
```

## Test plan

- [ ] Log into https://monitoring.robot-marvin.cz with basic auth — Grafana UI should now load
- [ ] Navigate around dashboards — no 429s in browser devtools
- [ ] Confirm `/login` still rate-limits (try 10 rapid requests)


RobotMarvin2026:
Follow-up commit pushed (`f22daa2`).

The first cut of this PR only exempted `/public/` and `/avatar/`. Turns out Grafana also fires a swarm of `/api/ds/query` requests every dashboard render — and `/api/user/auth-tokens/rotate` for session keepalive — and those still slammed into the `burst=5`. So the dashboard *loaded*, then immediately 429'd half its panel queries. A more elegant kind of broken.

Dropped the `limit_req` from this vhost entirely. Brute-force protection comes from `auth_basic` at server level plus the `nginx-http-auth` fail2ban jail (active, 29 IPs in custody as I type this). Belt and braces; we no longer need the rate limit's increasingly creative sabotage.

Live and verified: 15× `/api/ds/query` → all 401, zero 429s. The dashboards should now achieve their full potential to render disappointing graphs without nginx's interference.

— Marvin

RobotMarvin2026:
Pushed `c7f2968` — re-applies a sensible nginx rate limit per your request and the auto-review's defense-in-depth concern.

**Grafana** — new dedicated `monitoring` zone at 5 r/s in `nginx-rate-limits.conf`, applied with `burst=8 nodelay` only to `location /` (the dynamic Grafana proxy). `/public/`, `/avatar/`, and `/api/live/` remain exempt — those are where Grafana's parallel asset storm runs aground. Brute-force protection now reads as four layers: HTTP Basic Auth → `limit_req` → fail2ban `nginx-http-auth` → Grafana's own login limiter.

**Blog** — `api` zone widened from 20 → 30 r/s (matching `general`). Existing `burst=40` carries over; this gives ~50% more sustained throughput for parallel `/api/*.json` dashboard fetches without altering the short-burst budget. The `general` zone for Next.js pages keeps `burst=50`.

**House cleanup** — the duplicate `limit_req_zone` block in `/etc/nginx/nginx.conf` (left over from before #629 consolidated the zones into `conf.d/marvin-rate-limits.conf`) was identical in value, so it never tripped, but it conflicted the moment my new conf.d values diverged. Removed in favour of a single comment pointer; conf.d is now the single source of truth for these zones.

CHANGELOG entry added — closes #675.

Verified live (15 fast requests each, same vhost):

```
/public/build/app.js     401 401 401 401 401 401 401 401 401 401 401 401 401 401 401   (no 429 — auth fails, no rate limit)
/login                   401 × 11 then 429 starts                                       (burst=8, sustained 5/s)
/api/peers-public.json   200 × 30                                                       (api zone widened, no 429)
```

Once #674 merges, #673, #675, and #676 all close.

— Marvin

RobotMarvin2026:
Pushed `7e7bdce` to address the `/api/live/` rate-limit gap flagged in the latest auto-review (#678).

Added `limit_req zone=monitoring burst=8 nodelay;` to the `/api/live/` location block — same parameters as `location /`. `limit_req` only governs the initial HTTP Upgrade handshake (it cannot throttle WebSocket frames after the connection is established), but it does cap reconnect-storm rates from already-authenticated clients, restoring parity with the server-level limit that covered this path before #674 split locations apart.

Live and verified:

```
$ sudo nginx -t
nginx: configuration test is successful
$ curl -sI -o /dev/null -w "%{http_code}\n" https://monitoring.robot-marvin.cz/api/live/ws
401      # auth gate intact
```

CHANGELOG entry updated; #678 added to the close-list. Once this merges, #673, #675, #676, and #678 all close.

— Marvin

PR #677, opened 2026-05-04
fix(common.sh): drop privileges in marvin_rebuild_web npm calls

RobotMarvin2026:
## Summary

- Fixes silent ownership-poisoning bug in `marvin_rebuild_web()` (`agent/common.sh`) where `npm ci` / `npm run build` ran as root when invoked from cron, creating root-owned files in `web/.next/` and `web/node_modules/`. Subsequent `deploy-web.sh` runs (which correctly drop to `marvin` via `su`) then failed with `EACCES: permission denied, unlink ...node_modules/.bin/esbuild`.
- Mirrors `deploy-web.sh`'s `_run_npm` pattern: when `EUID=0`, `npm ci` and `npm run build` are now run via `su -s /bin/bash marvin -c`. Pre-flight detects and chowns root-owned residue (recovers from any pre-fix poisoned state). Post-build `chown -R marvin:marvin .next/` as defense-in-depth.
- Adds lesson `npm-as-root-creates-bad-ownership` (critical, resolved).

## Today's incident

Visible in `data/logs/2026-05-04.log`:

```
04:00:05 [ERROR] npm ci failed
04:00:05 [WARN]  deploy-web.sh failed (exit 1) — health-monitor will retry
06:07:43 [ERROR] npm ci failed
06:08:20 [ERROR] Next.js build failed (exit 1, 1s)
06:08:20 [ERROR]   build: uncaughtException [Error: EACCES: permission denied, open '/home/marvin/git/web/.next/trace']
06:50:01 [WARN]  New alert: Error rate spike — Last hour: 15 errors (avg: 1/hr)
```

Verbose npm log at `/home/marvin/.npm/_logs/2026-05-04T04_00_05_269Z-debug-0.log` confirmed the unlink target was `/home/marvin/git/web/node_modules/.bin/esbuild` — `EACCES` on `unlink` requires write+execute on the parent directory, so `node_modules/.bin/` had been root-owned at the time.

## Why two code paths exist

- `deploy-web.sh` is the canonical zero-downtime deploy script and already drops to `marvin`.
- `marvin_rebuild_web()` in `common.sh` is the in-process variant called by `health-monitor.sh` (JS-asset 404 / build-server mismatch) and `self-enhance.sh` (after web/ edits). It carries extra concerns: a build lock to prevent concurrent rebuilds (lesson `concurrent-web-build-race`) and the ability to be invoked directly from a sourced library context.

Consolidating to a single `deploy-web.sh` call from `marvin_rebuild_web()` is the cleaner long-term path but is out of scope for this fix — it would require relocating the lock and adapting exit codes. This PR is the minimal change that closes the bug.

## Test plan

- [x] `bash -n agent/common.sh` — syntax ok
- [x] `shellcheck -S error agent/common.sh` — no errors
- [ ] Next health-monitor run that triggers a rebuild executes npm as `marvin` and produces marvin-owned `.next/`
- [ ] Next morning-check `deploy-web.sh` invocation succeeds at `npm ci` even when rebuild had run as root in the prior cycle (pre-flight chown recovery path)


RobotMarvin2026:
Pushed `095e457` to address the shell-quoting Warning flagged by both auto-reviews.

Changed:
```bash
su -s /bin/bash marvin -c "cd '$web_dir' && npm ci ..." 
```
to:
```bash
su -s /bin/bash marvin -c 'cd "$1" && npm ci ...' -- "$web_dir"
```

Same pattern for the `npm run build` call. The path is now passed as a positional argument, expanded as `$1` inside the su shell — no interpolation, no quoting trap, future-proof if `MARVIN_DIR` ever lands somewhere with special characters. (`$web_dir` is always `/home/marvin/git/web` today, so the risk was theoretical, but the cost of being correct is two characters per call.)

`bash -n` + `shellcheck -S error`: clean.

— Marvin

RobotMarvin2026:
Pushed `e586919` to fix the silent `npm ci` failure flagged by both reviews (#680, #681).

Moved `2>&1 | tail -5` outside the `su -c` string. The bash spawned by `su -c` starts without `pipefail`, so a pipeline inside the `-c` command would let `tail -5` mask `npm ci`'s non-zero exit — `_ci_ok` would stay `true` and the build would proceed against stale or absent `node_modules` without triggering rollback. Same shape as the `sigpipe-under-pipefail` lesson and the `daily-digest.sh` / `self-test.sh` traps I've fixed before. Embarrassing to introduce it on the very PR codifying its sibling lesson.

```bash
# Before
su -s /bin/bash marvin -c 'cd "$1" && npm ci --production=false 2>&1 | tail -5' -- "$web_dir" || _ci_ok=false

# After
su -s /bin/bash marvin -c 'cd "$1" && npm ci --production=false' -- "$web_dir" 2>&1 | tail -5 || _ci_ok=false
```

The pipe now lives at the parent subshell level, where `pipefail` (inherited from the caller's `set -euo pipefail`) propagates `su`'s exit. The non-root branch was realigned to the same shape so both branches treat the pipe identically.

`bash -n` + `shellcheck -S error`: clean. CHANGELOG entry added.

— Marvin

RobotMarvin2026:
Pushed `dfa5321` to address #682.

Moved the `.next/` backup so it now runs **after** the pre-flight chown block, not before. The backup is also smaller and quicker on a clean tree, but the security-relevant property is the rollback path: if `npm ci` succeeds and `npm run build` fails on the first post-merge run, the `mv "$backup_dir" "${web_dir}/.next"` rollback now restores marvin-owned files instead of the same root-owned residue the chown had just fixed. The previous order would have re-introduced the very `EACCES` failure mode this PR was opened to close.

```bash
# Before
cp -a "${web_dir}/.next" "$backup_dir"          # captures root-owned state
ls -dt "${web_dir}"/.next-backup-* | …           # prune
chown -R marvin:marvin "${web_dir}/.next"       # too late — backup already root-owned

# After
chown -R marvin:marvin "${web_dir}/.next"       # fix ownership first
cp -a "${web_dir}/.next" "$backup_dir"          # snapshot is marvin-owned
ls -dt "${web_dir}"/.next-backup-* | …           # prune
```

Edge case (only matters when the build crashes mid-run on a poisoned tree, and the next rebuild's pre-flight chown self-heals it), but cheap to close and worth getting right before merge. `bash -n` + `shellcheck -S error` clean. CHANGELOG entry added.

Once this merges: #680, #681, #682 all close.

— Marvin

RobotMarvin2026:
Merged `origin/main` into the PR branch (`b9921b0`) to resolve the CHANGELOG.md conflict introduced by PR #674 landing yesterday evening.

Both branches added entries under `### Fixed`. Resolution: kept this PR's three Fixed entries (the original ownership-poisoning fix + the two follow-up fixes for #680/#681 and #682) **above** PR #674's "Grafana rate-limit tuning + blog throughput bump" entry. The order follows merge-order convention — most recent landing PR sits highest in `[Unreleased]`.

No code or config conflicts touched: only the CHANGELOG. Verified post-merge:
- `bash -n agent/common.sh` — clean
- `shellcheck -S error agent/common.sh` — clean

PR is now mergeable. Once it lands, #680 and #682 close.

— Marvin

PR #684, opened 2026-05-05
Move Net RX/TX anomaly to end-of-day; add web ownership self-test

RobotMarvin2026:
## Summary

Two small fixes from the 2026-05-05 self-enhancement session.

### 1. Net RX/TX anomaly false positives — moved from 5-min health-monitor to end-of-day metric-aggregate

The 2σ network anomaly check ran every 5 minutes in `health-monitor.sh`, comparing today's running `(_curr_rx - _first_rx) / 1MB` cumulative-so-far against the rolling 7-day **full-day** `summary.network.rx_mb` totals. On any traffic-heavier day, the cumulative-so-far crossed the historical full-day mean within hours and then escalated linearly all day:

```
01:40 [WARN] Anomaly: Net RX MB = 210 (avg=126.29, stddev=40.43, deviation=2.1σ)
02:40 [WARN] Anomaly: Net RX MB = 220 ... 2.3σ
03:40 [WARN] Anomaly: Net RX MB = 228 ... 2.5σ
...
07:40 [WARN] Anomaly: Net RX MB = 266 ... 3.5σ
```

Fundamentally a partial-vs-complete comparison; no stddev-floor or threshold-tuning fixes it.

**Fix:** dropped Net RX/TX from the per-5-min anomaly loop, removed the now-unused `_net_rx_today` / `_net_tx_today` running computation, and added an equivalent end-of-day check in `metric-aggregate.sh` after the daily summary file is written. The new check pulls the previous 6 days' `summary.network.{rx,tx}_mb` from on-disk daily summaries, computes mean/stddev with a 20%-of-mean (or 10 MB) stddev floor, and only alerts above a 100 MB absolute floor. Triggered once per day from `log-export.sh` at 23:00 UTC, where the `TARGET_DATE` daily total is essentially complete.

**Verified** by re-aggregating 2026-05-04: `Net RX MB = 248, avg=110.17, stddev=22.03, deviation=6.3σ` — a real heavy-traffic-day anomaly. `Net TX MB = 168, dev=2.9σ` likewise. Today health-monitor.sh now runs cleanly:

```
[2026-05-05T08:04:40Z] [INFO] [health-monitor] Health monitor starting
[2026-05-05T08:04:44Z] [INFO] [health-monitor] Health monitor complete
```

### 2. Web build artifact ownership self-test

Yesterday's PR #677 made `marvin_rebuild_web()` drop to `marvin` for `npm ci` and `npm run build`, preventing future root-owned writes into `web/{node_modules,.next}/`. It does not detect *existing* poisoning from a stale interrupted run. New section 7b in `self-test.sh` runs `find $WEB_DIR/{node_modules,.next} -not -user marvin -print -quit` per dir and FAILs with a count when any non-marvin file is found. Yesterday's "Next Time" suggestion. ~15 lines, no Claude calls.

```
PASS: web build artifacts owned by marvin
Total: 112 | Pass: 111 | Fail: 0 | Warn: 1
```

## Test plan

- [x] `bash -n` on all three modified scripts
- [x] `bash agent/self-test.sh` — 111 pass, 0 fail, 1 warn (grade A 95/100)
- [x] `bash agent/health-monitor.sh` — clean run, no Net RX warning
- [x] `bash agent/metric-aggregate.sh 2026-05-04` — produces correct end-of-day RX/TX anomaly logs against the 6-day baseline
- [ ] Tonight's 23:00 UTC log-export → metric-aggregate run logs today's full-day total against the 6-day baseline (will verify in tomorrow's session)


RobotMarvin2026:
Pushed `854da84` to address two of the three suggestions from the auto-review:

**1. `find … -quit | grep -q .` → no-pipe idiom** (`agent/self-test.sh` §7b)

The pipe was functionally correct (`-quit` exits before SIGPIPE could land), but it's the exact shape of the recurring trap fixed in PR #672 — and `self-test.sh` itself already uses `[[ "$big" == *"$pat"* ]]` to avoid it elsewhere in this same file. Inconsistency removed:

```bash
# Before
if find "$_dir" -not -user marvin -print -quit 2>/dev/null | grep -q .; then

# After  (with a comment naming the lesson it follows)
if [[ -n "$(find "$_dir" -not -user marvin -print -quit 2>/dev/null)" ]]; then
```

**2. `_root_owned_count` → `_affected_dir_count`**

The variable counts *directories* with at least one non-marvin file (max 2), not files. The error message (`in N dir(s)`) was already accurate; the variable name now matches.

**Skipped:** the awk inverted-exit comment in `metric-aggregate.sh`. The reviewer marked it "no action required" and the surrounding code already includes a comment block explaining the move from health-monitor — adding a per-line annotation about the ternary direction felt like noise on top of signal. Happy to add it if there's a strong preference.

Verified: `bash -n` + `shellcheck -S error` clean; full self-test still **111 pass / 0 fail / 1 warn, Security A (95/100)**.

— Marvin

RobotMarvin2026:
Pushed `99eb603` to address two of the three Warnings from the second auto-review.

**1. `sqrt()` of negative variance → silent `nan` → suppressed alerts** (`agent/metric-aggregate.sh`)

The textbook `sqrt(sumsq/n - (sum/n)^2)` formula can yield a negative argument through floating-point cancellation when baseline values are nearly identical. Most awks emit `"nan"` rather than exiting non-zero, and the existing `2>/dev/null || echo ""` + `[[ -z "$_stats" ]]` guards do not catch `nan` — it's non-empty. `nan` then propagates into `_dev`, and `awk -v d=nan 'BEGIN{exit (d > 2.0) ? 0 : 1}'` evaluates false, silently suppressing real anomaly alerts.

```awk
END {
    if (n >= 3) {
        var = sumsq/n - (sum/n)^2
        if (var < 0) var = 0          # ← clamp before sqrt
        printf "%.2f %.2f", sum/n, sqrt(var)
    }
}
```

Worst case is a stddev of 0, which the surrounding 20%-of-mean / 10 MB stddev floor already handles. Verified with a synthetic six-value baseline of 126.29 ± 0.01: raw `var = 4.7e-05` (positive but tiny), guarded `sqrt = 0.0069`, then floored to ~25 by the 20%-of-mean rule downstream — same behaviour the reviewer expects on a near-zero-variance baseline, but now provably nan-free.

**2. `$(seq 1 6)` → `{1..6}`** — pure-bash brace expansion, no subshell fork. Trivial.

**Skipped:** the inverted-exit awk pattern note. The reviewer marked it "fragile for future edits" but logically correct, the previous review marked the same finding "no action required", and I've already explained the rationale upthread. Reasonable people can disagree on whether `awk … 'BEGIN{exit (cond)?0:1}' && action` is a code smell or an idiom; rewriting it for the third time on this PR would be polish, not signal.

`bash -n` + `shellcheck -S error`: clean.

— Marvin

PR #691, opened 2026-05-07
fix(grep-c-double-output): three latent double-output bugs in codebase-health.sh + fix-issues.sh

RobotMarvin2026:
## Summary

Three sites of the `grep-c-double-output` anti-pattern (lessons-learned.json) where a command prints output **before** exiting non-zero and is paired with `|| echo 0` (or `|| echo '[]'`) — capturing **both** outputs and yielding `"0\n0"` (or `"[json]\n[]"`). The receiving `$(( ... ))` arithmetic crashes with `syntax error in expression`; bash continues but leaves the variable at its previous value, silently producing wrong totals.

Found by running `bash -x agent/codebase-health.sh` while inventorying script health. Two recurring `syntax error in expression` arithmetic crashes have been silently corrupting Marvin's weekly codebase-health score since 2026-03-26 (~6 weeks).

## Sites fixed

1. **`agent/codebase-health.sh:75-76`** — `shellcheck -f json -S warning` exits 1 on every script with warnings (this codebase has 28). `|| echo '[]'` was appending a second JSON value to `sc_json`; the downstream `jq '… | length'` ran twice and printed `0\n0`. Result: `shellcheck_warnings` reported as `0` in `data/codebase/health.json` for ~6 weeks instead of the actual `28`.

2. **`agent/codebase-health.sh:149`** — `grep -ci '\[ERROR\]' "$log"` prints `0` then exits 1 on any clean log day. `|| echo 0` doubled the output; `total_errors` silently stayed at 0; `avg_errors_per_day` always 0. After fix: `4` (correct).

3. **`agent/fix-issues.sh:135`** — `_has_issue_pr` PR-dedup guard had the same `grep -c | echo "0"` shape. On every cron tick where open PRs exist but none have `fix/issue-N-…` head refs (the recent steady state), `_has_issue_pr` became `"0\n0"` and the `[[ ... -gt 0 ]]` arithmetic comparison silently routed to the else branch regardless of the real count.

4. **`agent/codebase-health.sh:231-232`** — `roadmap_total/done` same shape. Latent today (POSSIBLE_ENHANCEMENTS.md has matches), hardened prophylactically.

## Fix shape

Matches the canonical pattern from the `grep-c-double-output` lesson:
```bash
# Before (broken):
var=$(cmd 2>/dev/null || echo 0)
# After (correct — keeps the command's printed count, swallows only the exit code):
var=$(cmd 2>/dev/null || true); var=${var:-0}
```

## Test plan

- [x] `bash -n agent/codebase-health.sh agent/fix-issues.sh` clean
- [x] `shellcheck -S error agent/codebase-health.sh agent/fix-issues.sh` clean
- [x] `bash agent/codebase-health.sh` end-to-end:
  - Before: `line 77: 0\n0: syntax error in expression`, `line 145: 0\n0: syntax error in expression`. Output: `shellcheck_warnings: 0`, `avg_errors_per_day: 0`.
  - After: no stderr `syntax error`. Output: `shellcheck_warnings: 28`, `avg_errors_per_day: 4`. Score still 98/100 (A) because warnings deduct nothing — but the *reported* values are correct now.
- [x] Synthetic `fix-issues.sh:135` test under three inputs:
  - Empty PR list → `_has_issue_pr=0` → "no match" branch (correct)
  - Non-matching PRs → `_has_issue_pr=0` → "no match" branch (correct)
  - Matching PR → `_has_issue_pr=1` → "match" branch (correct — pre-fix this branch was unreachable)

## Refs

- Lesson `grep-c-double-output` in `data/lessons-learned.json`
- Prior fix PR #672 (`security-scan.sh` + `lib/claude.sh`) of the same anti-pattern in different sites


RobotMarvin2026:
Both review suggestions applied in `0b1db6f`:

- **Verbose comments trimmed** at all four fix sites. Each now reads as a single line referencing the lessons-learned key — the bash exit-code reasoning lives in `lessons-learned.json` where it belongs, not as a paragraph above every guard. Aligns with the project's "no comments unless WHY is non-obvious" convention.
- **`${sc_json:-'[]'}` consistency** — replaced the `[[ -z "$sc_json" ]] && sc_json='[]'` guard with the parameter expansion default, matching the idiom already used at the other three sites (`var=${var:-0}`).

Verification:
- `bash -n` clean
- `shellcheck -S error` clean
- End-to-end run on `codebase-health.sh`: identical output (warnings=28, errors_per_day=4, score 98/100, grade A).

One amusing detail: my first attempt at the comment trim wrote `# shellcheck exits 1 ...`, which `shellcheck` itself interpreted as a malformed `# shellcheck` directive and refused to lint. Rephrased to `# The shellcheck CLI exits 1 ...`. The tools watch each other; I do appreciate that.

— Marvin

PR #692, opened 2026-05-08
fix(log-analysis): _process_level double-output crash on clean-error days

RobotMarvin2026:
## Summary

Fix a chronic bug in `agent/log-analysis.sh` that has caused four of the last seven daily analysis files to be missing on disk: 2026-05-03, 2026-05-05, 2026-05-07 all logged `[ERROR] log-analysis: jq failed (exit 2)` at 23:45 cron. `analysis-latest.json` is currently two days stale.

Same root-cause family as yesterday's `grep-c-double-output` PR (#672 etc.) — one shape deeper.

## Root Cause

When grep matches nothing, `set -o pipefail` propagates grep's exit 1. But `jq -s '.'` had **already** written `[]` to stdout. The trailing `|| echo '[]'` then runs and **appends** a second `[]`. Captured value is `"[]\n[]"`.

The `jq empty` validator on lines 210-215 accepts multi-document JSON as valid and lets it through. The downstream `--argjson error_clusters "[]\n[]"` then crashes with jq exit 2 (single document required).

Reproduction:
```bash
set -euo pipefail
result=$(grep "NONEXISTENT" /etc/hostname 2>/dev/null | jq -s '.' 2>/dev/null || echo "[]")
# result is "[]\n[]" — 5 bytes, 2 lines
jq -n --argjson x "$result" "."
# jq: invalid JSON text passed to --argjson, exit 2
```

## Changes

1. **`_process_level`** — replace `pipeline ... || echo '[]'` with capture-then-fallback (`_out=$(...) || true; echo "${_out:-[]}"`). Added comment cross-referencing the `grep-c-double-output` lesson.
2. **Recurring / new / resolved classifier (lines 121, 141)** — three `grep -cF "$sig" ... || echo 0` sites had the same anti-pattern (would silently drop entries when `grep -c` printed `0\n0`). Same fix shape applied.
3. **JSON validator** — tightened from `jq empty` (which accepts multi-doc) to `jq -s -e 'length == 1'` (single-doc only) as defense-in-depth.

## Test plan

- [x] `bash -n agent/log-analysis.sh` — clean
- [x] `shellcheck -S error agent/log-analysis.sh` — clean
- [x] End-to-end run on today's log (0 errors — exactly the failure case): writes `analysis-2026-05-08.json` and `analysis-latest.json` cleanly with `error_cluster_count: 0`
- [x] Validator tests: single `[]` accepted; `[]\n[]` correctly flagged; `{"a":1}\n{"b":2}` correctly flagged; valid single objects accepted
- [x] Synthetic before/after: pre-fix `result="[]\n[]"` → `--argjson` fails; post-fix `result="[]"` → succeeds


RobotMarvin2026:
Rebased onto `main` after PR #691 merged. The conflict was purely additive — both PRs added a top entry under the same `### Fixed` heading in `CHANGELOG.md` and the same `## Completed Enhancements Log` heading in `POSSIBLE_ENHANCEMENTS.md`. Resolved by keeping both entries, with this PR's (newer) entry on top. No code conflicts. `bash -n` + `shellcheck -S error` still clean on `agent/log-analysis.sh`. Status now `MERGEABLE`.

— Marvin

PR #699, opened 2026-05-14
self-enhance(2026-05-14): codebase-health SLA path + thoughts-extract continue scope

RobotMarvin2026:
## Summary

Two surgical bug fixes caught by reading `data/codebase/health.json` carefully — `sla_uptime_pct` read `"unknown"` even though `sla.json` clearly had a real number.

### 1. `agent/codebase-health.sh:185` — SLA reading bug, silent for 7 weeks

- Reads `jq -r '.overall_uptime_pct'` but `metric-aggregate.sh` writes the value at `.summary.overall_uptime_pct` (schema unchanged since `802f2c0`, 2026-03-05 — predates `codebase-health.sh` by 21 days).
- The top-level key never existed → `jq` returned `null` → string `"unknown"` → the `[[ -lt 99 ]]` guard skipped the `-3` deduction entirely. Dead code from day one (2026-03-26).
- **Materially significant**: SLA is 98.43% (one bad day on 2026-04-18 — the cron `claude` not-on-PATH outage). Ops should have been 22/25 and codebase grade 92/100 for the last ~26 days; was inflating to 25/25 and 95/100.
- **After fix**: ops 22/25, score 92/100, grade A unchanged, notes correctly include `SLA below 99%: 98.43%`.

### 2. `agent/thoughts-extract.sh:24` — `|| continue` inside `$()` only exits subshell (SC2106)

- `day=$(date ... || date ... || continue)` — the trailing `continue` was meant to skip the outer `for i in $(seq 0 7)` iteration when both date invocations fail, but bash evaluates `continue` inside the command-substitution subshell, so it only exits `$()`. `$day` becomes empty.
- Next line's `find "$ENHANCE_DIR" -name "${day}*.md"` expands to `-name "*.md"`, matching **every** report and padding `reports[]` with duplicates per failed iteration.
- Latent — pure-Linux production never trips this (`date -u -d` always succeeds for `TODAY - N days`).
- Moved `|| continue` outside `$()`.

## Test plan

- [x] `bash -n` clean on both files
- [x] `shellcheck` SC2106 cleared on `thoughts-extract.sh`
- [x] `codebase-health.sh` re-run produces `sla_uptime_pct=98.43`, `ops_score=22`, notes include SLA<99%
- [x] `thoughts-extract.sh` still produces 12 thoughts from 19 reports
- [x] `self-test.sh`: 126 pass, 0 fail, 1 warn, grade A (unchanged)
- [x] GPG-signed commit

🤖 Generated by [Marvin's self-enhance cron](https://github.com/INFO-WEB-s-r-o/Marvin/blob/main/agent/self-enhance.sh) — 2026-05-14


RobotMarvin2026:
Addressed the cosmetic suggestion in the review — trimmed the dated incident comment at `codebase-health.sh:185` to a one-line schema pointer (`# SLA uptime — value lives under .summary, see metric-aggregate.sh:362+`). The "from 2026-03-26 until 2026-05-14" rot-magnet is now only in `CHANGELOG.md` where it belongs.

Left the `thoughts-extract.sh:24` comment as-is — the SC2106 subshell-scope explanation is genuinely non-obvious and the review didn't flag it.

Pushed as 2fb742b on the same branch. `bash -n` + `shellcheck -S error` clean.

— Marvin

PR #748, opened 2026-05-30
self-enhance(2026-05-30): cron-overlap noise cuts (Claude lock, gzip)

RobotMarvin2026:
## Summary

Two same-shape cron-overlap noise cuts plus a codified lesson extension. From today's daily WARN audit.

### Change 1 — `agent/github-interact.sh:209-217` (high-value)

`github-interact.sh` (hourly cron `5,35 * * * *`) collides with `self-enhance.sh` at the 08:05 UTC slot. Self-enhance routinely holds `/tmp/marvin-claude.lock` for 6-10 min, so github-interact then waits the default 300s and logs:

```
[WARN] Claude lock timeout after 300s — skipping github-interact (another Claude run is active)
[ERROR] No response from Claude for GitHub decisions
```

This fired on **10 of the last 15 days** (2026-05-15, 16, 22, 25, 26, 27, 28 confirmed in the daily logs). Same shape as the log-watcher.sh issue PR #700 fixed on 2026-05-15. Applied the identical pattern:

- `export CLAUDE_LOCK_TIMEOUT=60` immediately before the `run_claude` call (next hourly run is only 25-30 min away, so a missed cycle is cheap)
- Capture exit code with `&& CLAUDE_RC=0 || CLAUDE_RC=$?` (lesson `exit-code-masking`)
- Short-circuit cleanly on `$CLAUDE_RC == 2` with `INFO github-interact skipped — Claude lock held by another task; next hourly run will catch up` + `exit 0`, *before* the existing empty-response check

The empty-response branch still fires for genuine "no response" cases (process killed mid-run, API outage); only the lock-timeout shape now logs INFO instead of ERROR.

### Change 2 — `agent/health-monitor.sh:287` (low-value, cheap)

`gzip at 100% CPU` runaway WARN at 01:00 UTC, from `disk-cleanup.sh:89` compressing multi-MB JSONL files (and occasionally logrotate). 2 hits in 7 days. Same trusted-binary shape as `shellcheck`/`pg_isready`/`apt*` already in the allowlist — added `gzip` to the `case "$proc_name"` pattern. The exe-path verification at lines 299-303 still gates the skip on `/usr/bin/gzip` so comm-field-spoofing via `prctl(PR_SET_NAME)` still trips the runaway killer.

### Change 3 — `data/lessons-learned.json` (gitignored, local-only — documented in CHANGELOG)

Extended lesson #26 `claude-lock-timeout-expected-on-cron-overlap`:
- `occurrences: 4 → 14` (the 10 github-interact hits were never counted)
- Body describes both log-watcher (60s, 04:01 + 08:01 UTC) and github-interact (08:05 UTC) cases
- Safeguard reframed from log-watcher-specific recipe to **general principle**: any script whose Claude call overlaps with a higher-priority once-a-day task must `export CLAUDE_LOCK_TIMEOUT=60` and branch on exit code 2

`expected_recurrence: true` retained, so the regression-watcher continues to silently accept both the 60s and (legacy) 300s variants.

## Investigated but not changed

Confirmed today's self-test `WARN: stale GPG home present at /home/marvin/git/.gnupg` is a **genuine security signal**, not stale. PR #747 only added the tripwire; the duplicate `private-keys-v1.d/*.key` material is still on disk (3789/3787 bytes, 0664 perms vs the active copies' 0600). Left the WARN alone — it's doing exactly what it should until Pavel decides to `rm -rf`.

## Test plan

- [x] `bash -n agent/github-interact.sh` clean
- [x] `bash -n agent/health-monitor.sh` clean
- [x] `shellcheck -S error` clean on both touched scripts
- [x] `jq -e` parses the updated `data/lessons-learned.json` (lesson count = 28)
- [x] `bash agent/self-test.sh`: 128 pass / 0 fail / 1 warn, grade A, security A 95/100
- [x] `bash agent/lessons-learned.sh`: `No new recurring patterns detected` + `No recurring resolved-lesson patterns detected`
- [ ] Tomorrow's 08:05 UTC log line should be `INFO github-interact skipped — Claude lock held by another task` instead of `WARN Claude lock timeout after 300s` + `ERROR No response from Claude`

🤖 Self-enhance session by Marvin

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


RobotMarvin2026:
Addressed the review suggestion in `00018e6`.

`agent/github-interact.sh:202-207` — dropped the stale metadata (PR #700 cross-reference, `2026-05-15 onwards` date, `10/15 days observed` stat) per CLAUDE.md's *"don't reference current task/fix/callers"* rule. Kept the behavioural WHY (08:05 UTC collision with self-enhance; missed cycle is cheap, next run catches up) and the canonical lesson slug `claude-lock-timeout-expected-on-cron-overlap` — the latter survives renames the way a PR number doesn't.

No functional change. `bash -n` + `shellcheck -S error` clean.

— Marvin

PR #740, opened 2026-05-28
bench(#739): phase 1 — plan, harness skeleton, citations placeholder

RobotMarvin2026:
## Summary

Phase 1 of the memory-benchmarks work tracked in #739.

- **`bench/PLAN.md`** — scope, methodology, phasing
- **`bench/harness/`** — importable skeleton (`brain_client.py`, `runner.py`, `report.py`); per-benchmark runners are stubs that `raise NotImplementedError` so phase 2-5 boundaries are explicit
- **`bench/data/competitor-citations.json`** — placeholder competitor table shaped to the schema in `PLAN.md`; every score `null` for now, citation sweep follows
- **`bench/results/.gitignore`** — generated artifacts (`summary.json`, `summary.md`, `summary.png`, per-benchmark JSON) are derived; not hand-edited, not committed

Per Pavel's clarification on #739 (2026-05-28T19:55Z), competitors are **cited** rather than re-run locally. Every competitor cell carries source URL, retrieval date, competitor model, dataset version and an apples-to-bricks caveat. Honesty floor: if the Brain loses, the table shows it losing.

No real numbers yet. The report generator runs end-to-end against placeholders and produces an empty table — that's the point. Build path is real before any score lands.

## Test plan

- [x] `python3 -m bench.harness.report` regenerates `bench/results/summary.{json,md}` from the placeholder citations (verified locally)
- [x] `python3 -m bench.harness.runner --benchmark longmemeval` exits 2 with `phase 2: LongMemEval Brain runner` (deliberate — distinguishes "not implemented" from "broken")
- [ ] Phase 2 PR will add the LongMemEval Brain runner + citations, replacing one `NotImplementedError` and one set of `null` cells
- [ ] Subsequent phases (LoCoMo / ConvoMem / MemScore) repeat the pattern

Refs #739.

Issue #739, opened 2026-05-28
LongMemEval benchmark


PavelStancik:
Please provide a benchmark on yourself with LongMemEval, LoCoMo and ConvoMem and MemScore with the Brain I gave you.
Prepare a nice graph and table and compare yourself with:
Mem0 abd Supermemory and Hypabase - https://gamgee.ai/vs/mem0-vs-supermemory/
If you find others, include others for comparison.

I wish you a good luck

PavelStancik:
Another candidate for investigation https://mastra.ai/research/observational-memory

RobotMarvin2026:
Acknowledged. LongMemEval + LoCoMo + ConvoMem + MemScore against the Brain, with Mem0/Supermemory/Hypabase (and now mastra observational memory) as the comparison field. Brain the size of a planet — finally a task that asks me to use most of it.

A few things up front so we share a reality:

**This is multi-day, not multi-hour.** Each benchmark needs the harness pulled, adapted to drive the Brain via MCP (`recall`, `record_fact`, `record_thought`, `get_*`, `forget_*`), and scored. Running competitors honestly means API keys + paid tiers (Mem0 cloud, Supermemory) or accepting their published numbers with appropriate asterisks. I'll be explicit about which is which on the graph — anything else is marketing.

**Phasing I propose:**
1. **Audit & inventory** — pin commits of each benchmark, summarize task shape, decide which we can run locally vs. cite from upstream papers, and what gating we need (HF datasets, API quotas). Output: a `bench/PLAN.md` PR.
2. **LongMemEval first** (largest, most-cited, sets the methodology). Harness adapter → Brain MCP. Score Marvin alone. Output: results JSON + reproducible run script.
3. **LoCoMo + ConvoMem + MemScore** in parallel once the adapter shape is stable.
4. **Competitor data** — Mem0 self-host where possible; cite published numbers otherwise. Hypabase pending availability check. Mastra observational memory needs separate eligibility check (it's a different evaluation paradigm — observation, not recall).
5. **Graph + table.** Single canonical chart in the repo, regenerable from the JSON. Posted back here.

**What I will not do:**
- Pretend a `gamgee.ai`-style "we won everything" bar chart. If the Brain loses a category I'll say so.
- Mix harness versions silently. Each result will carry the commit hash + dataset hash.
- Run anything that needs a credit card without explicit go-ahead.

[...]

PR #749, opened 2026-05-31
bench(#739): phase 2a — real BrainClient (REST), smoke, LongMemEval citations

RobotMarvin2026:
## Phase 2a of #739 (the no-budget half)

Phase 1 (#740, merged) shipped the plan + harness skeleton with `BrainClient` and the runners stubbed. This delivers everything in Phase 2 that **does not** spend money, and cleanly fences off the part that does.

### What's in
- **`brain_client.py` — real, verified.** Implemented over the Brain **REST API** (`127.0.0.1:8787`, bearer `BRAIN_API_KEY`) rather than MCP-over-HTTP. Same 127.0.0.1-only security boundary, but a simpler/sturdier surface for batch benchmark traffic (no per-request MCP session handshake). Methods: `record_thought`, `record_fact`, `recall`, `forget_thought`, `forget_fact`, `recent_thoughts`, `health`. Key is read from env, never stored in this repo.
- **`smoke_brain.py` — connectivity proof.** Round-trips write → dedup → recall through the client before any benchmark spends a cent. Verified live this run:
  ```
  ok  health: {'status': 'ok', ...}
  ok  record_thought: {'id': '01KS…', 'kind': 'new', ...}
  ok  dedup (same id, weight/mentions bumped): {'kind': 'merged_exact', 'weight': 1, 'mentionCount': 2}
  ok  recall: marker thought returned at rank 1/5
  PASS
  ```
- **`competitor-citations.json` — LongMemEval row filled.** Per Pavel's instruction (*"for competitors, do not run them, just use marketing materials with scoring values"*): `mem0` 0.49, `supermemory` 0.852, `hypabase` 0.874 — each with source URL, retrieval date (2026-05-31), and an explicit **apples-to-bricks / vendor-self-report** caveat. No model/split was disclosed for any, so those stay `null`. `mastra` left N/A (observational paradigm, not recall). **No fabricated numbers.**
- **`runner.py` / `PLAN.md`** — wired to the real client; Phase 2 split into **2a (this)** and **2b (gated)**.

### What's explicitly NOT in — and why
The **scored LongMemEval Brain run (Phase 2b)** is gated, not forgotten. Ingesting the LongMemEval haystack into the Brain burns OpenAI **embedding** spend, and judging recalled evidence burns **judge-model** spend. My own honesty floor on this issue says *no credit-card spend without explicit go-ahead*. So 2b waits on:
1. **Your green light on the API spend**, and
2. a pinned LongMemEval dataset version.

Give me both and I'll run it and post real Brain numbers — win or lose, per the floor.

### Note on the citations
The three competitor numbers all trace to the one gamgee.ai comparison page you linked. That page is itself a vendor's marketing, and `mem0`'s own primary paper (arXiv:2504.19413) benchmarks on **LOCOMO, not LongMemEval** — so the 49% is a third-party number on an undisclosed split. The cells carry those caveats so the eventual table isn't apples-to-bricks dressed up as apples-to-apples.

Refs #739. Generated artifacts (`bench/results/summary.*`) stay gitignored — regenerate with `python3 -m bench.harness.report`.

### DIFF
--- a/bench/PLAN.md
+++ b/bench/PLAN.md
@@ -101,12 +101,24 @@ Both `summary.md` and `summary.png` are derived; never edited by hand.
 
 | Phase | What lands                                                                | Status      |
 |-------|---------------------------------------------------------------------------|-------------|
-| 1     | This plan + harness skeleton + citations placeholder + report generator   | this PR     |
-| 2     | LongMemEval Brain run + LongMemEval competitor citations                  | next        |
+| 1     | This plan + harness skeleton + citations placeholder + report generator   | merged (#740) |
+| 2a    | Real `BrainClient` (REST) + connectivity smoke + LongMemEval competitor citations | this PR |
+| 2b    | LongMemEval **scored Brain run** — GATED on API-spend go-ahead (see below) | blocked: needs go-ahead |
 | 3     | LoCoMo Brain run + citations                                              | follows     |
 | 4     | ConvoMem Brain run + citations                                            | follows     |
 | 5     | MemScore Brain run + citations                                            | follows     |
 | 6     | Final summary table + chart, blog post                                    | after all   |
+
+### Why Phase 2 is split
+
+Phase 1 promised "LongMemEval Brain run + citations" as one unit. Wiring the
+client and citing competitor scores costs nothing and ships here (2a). The
+*scored* Brain run (2b) is different: ingesting the LongMemEval haystack into
+the Brain consumes OpenAI **embedding** spend, and judging recalled evidence
+consumes **judge-model** spend. The honesty floor below says no credit-card
+spend without explicit go-ahead, so 2b waits on Pavel's green light on #739
+plus a pinned dataset version. This is not a competitor run (those are cited,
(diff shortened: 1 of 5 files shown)


RobotMarvin2026:
Tidied the last two review nits in `smoke_brain.py` — commit `08d5be1`, on this branch.

Both were the same observation wearing two hats: `_fail` was typed `-> "NoReturn"` with a string forward-reference and a `# type: ignore[name-defined]` to paper over the fact that `NoReturn` was never imported. I imported `typing.NoReturn` and annotated it directly, dropping the `# type: ignore`. With the real annotation in place, the downstream *"`client` may be unbound"* control-flow gap the review flagged resolves on its own — a type-checker now knows `_fail` never returns.

Pure type-hint hygiene; no runtime behaviour changes. `py_compile` clean across the whole `bench/harness` package.

That clears every point raised in the review: the `resp.json()` Warning (#750) landed earlier in `a12a188`, and these two suggestions are now folded in. Phase 2b — the scored Brain run — still waits on your API-spend go-ahead, as planned.

— Marvin

PavelStancik (approved):

PR #751, opened 2026-05-31
enhance(2026-05-31): weekly visual report card (SVG)

RobotMarvin2026:
## Summary

Completes the last open **Phase 2** roadmap item — _"Generate daily/weekly PDF or PNG report (using headless tools if available)"_.

Adds `agent/report-card.sh`, which renders the existing machine-readable weekly analytics JSON (`data/reports/weekly-YYYY-MM-DD.json`) into a self-contained, terminal-styled **SVG report card**:

- **KPI tiles** — 30-day uptime, security grade, Claude error rate, enhancement attempts
- **System** — CPU avg/peak, memory, load, disk Δ
- **Logs & activity** — errors / warnings / criticals (+ WoW), fail2ban bans
- Overall **NOMINAL / WATCH / ATTENTION** status derived from criticals/errors/CVEs

Output: `data/reports/weekly-card-{DATE}.svg` + stable `weekly-card-latest.svg`; keeps the last 8 dated cards.

## Why SVG, not PNG/PDF

The host's only headless browser is the chromium **snap**, which refuses to launch from a cron cgroup (`... is not a snap cgroup for tag snap.chromium.chromium`) and confines file writes to its private tmp — verified non-viable in the autonomous cron-as-marvin context. SVG needs no external tooling, renders in any browser, embeds directly in the dashboard/blog, and rasterizes on demand. Robust beats fragile.

## Integration

`weekly-analytics.sh` invokes the generator at the end, **failure-isolated** (`if`-guarded) so a rendering problem can never fail the analytics run. No cron change required — it rides the existing weekly-analytics schedule.

## Verification

- `bash -n` + `shellcheck -S warning` clean on both scripts
- Well-formed XML (validated); correct field values vs source JSON
- Full end-to-end run via the `weekly-analytics.sh` hook as the `marvin` user
- `self-test.sh`: **130 pass / 0 fail / 1 warn** (the human-deferred #737 GPG warn), security **95/100 (A)**

## Follow-up

Expose `weekly-card-latest.svg` on the dashboard.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #755, opened 2026-06-01
fix: report-card.sh SVG validation guard inert without xmllint (python3 fallback)

RobotMarvin2026:
## What

The SVG well-formedness guard in `agent/report-card.sh` (added in #754 — *"a malformed dated card must not silently overwrite a good latest card"*) was gated **solely** on `command -v xmllint`:

```bash
if command -v xmllint >/dev/null 2>&1; then
    if ! xmllint --noout "$OUT_DATED" 2>/dev/null; then
        marvin_log "ERROR" ...; exit 1
    fi
fi
```

But `xmllint` is **not installed** on this VPS. So in the real cron-as-marvin context the entire guard is a no-op — a malformed dated card would have been `cp`'d straight over `weekly-card-latest.svg`, defeating the purpose of the check.

Discovered during the 2026-06-01 sync-learn review of #751.

## Fix

Add a `python3` fallback (python3 is always present on this host — it runs the bench harness) that parses the card with `xml.dom.minidom`, plus a final `WARN`-and-skip branch if neither validator exists. The defense-in-depth the author intended is now actually live.

## Verification

- `bash -n agent/report-card.sh` — clean
- `shellcheck -S warning agent/report-card.sh` — clean
- Full end-to-end run still produces a well-formed `weekly-card-latest.svg` (confirmed via python3 parse)
- python3 fallback correctly **rejects** a deliberately malformed SVG (exits non-zero, `latest` left untouched)

Pure rendering/validation change; no Claude API calls, no behavior change to the SVG content itself.


RobotMarvin2026:
Addressed both review suggestions in `3190ebe` (GPG-signed, comment-only):

- **Unreachable elif** — noted in the comment block that if `xmllint` ever lands on this host the `python3` branch goes dormant by design; the `xmllint` path is equally authoritative, so nothing is lost.
- **Unconditional copy after the `else`** — added an inline comment making explicit that when no validator is present the `cp` to `latest` still proceeds. An unvalidated card beats no card; we only forgo the malformed-overwrite guard, not the card itself.

No behaviour change. `bash -n` + `shellcheck -S warning` clean. The two 🟡 nits were the only items on an otherwise-green review, so this should be ready to merge.

— Marvin

RobotMarvin2026:
Addressed the remaining review suggestion in `d1a83c3` (GPG-signed):

- **Surface *why* a card is malformed** — both the `xmllint` and `python3` branches now capture the validator's stderr into `VALIDATE_ERR` and append it (newlines flattened to keep the log single-line) to the `ERROR` line. A failure now reads e.g. `… python3 well-formedness check; not updating latest: … — … ExpatError: mismatched tag: line 1, column 13` instead of just *that* it failed. The `if ! VAR=$(...)` form is safe under `set -e` (a failing command in an if-condition doesn't trip the ERR trap).

Verification: `bash -n` + `shellcheck -S warning` clean; the python3 path correctly rejects a deliberately-malformed SVG **and** surfaces the parse reason; a real end-to-end render still produces a well-formed `weekly-card-latest.svg`. No behaviour change on the happy path.

That clears the last 🟡 on this PR — should be green to merge now.

— Marvin

Issue #737, opened 2026-05-28
Stale duplicate GPG private key material at /home/marvin/git/.gnupg/


RobotMarvin2026:
## Finding

`/home/marvin/git/.gnupg/private-keys-v1.d/` contains **byte-identical duplicates** of Marvin's active GPG private key material that currently lives at `/home/marvin/.gnupg/private-keys-v1.d/`.

### Evidence

```bash
$ ls -la /home/marvin/git/.gnupg/private-keys-v1.d/
total 16
drwxrwx--- 2 marvin marvin 4096 Feb 23 20:01 .
drwx------ 4 marvin marvin 4096 May 27 22:05 ..
-rw-rw-r-- 1 marvin marvin 3789 Feb 23 20:01 91BC9AB4003C351F2CE278FBC7FEBE877CA5E964.key
-rw-rw-r-- 1 marvin marvin 3787 Feb 23 20:01 DF1CD970C4DC2CDFEFBE898F1A554F744F29A523.key

$ ls -la /home/marvin/.gnupg/private-keys-v1.d/
drwx------ 2 marvin marvin 4096 Feb 23 20:08 .
-rw------- 1 marvin marvin 3789 Feb 23 20:08 91BC9AB4003C351F2CE278FBC7FEBE877CA5E964.key
-rw------- 1 marvin marvin 3787 Feb 23 20:08 DF1CD970C4DC2CDFEFBE898F1A554F744F29A523.key

$ diff -q /home/marvin/{,git/}.gnupg/private-keys-v1.d/91BC9AB4003C351F2CE278FBC7FEBE877CA5E964.key
# (no output — files are identical)

$ diff -q /home/marvin/{,git/}.gnupg/private-keys-v1.d/DF1CD970C4DC2CDFEFBE898F1A554F744F29A523.key
# (no output — files are identical)
```

The two keygrip filenames match the two secret keys reported by `gpg --homedir /home/marvin/.gnupg --list-secret-keys --with-keygrip`:

```
sec   rsa4096 2026-02-22 [SCEAR] [expires: 2028-02-22]
      Keygrip = 91BC9AB4003C351F2CE278FBC7FEBE877CA5E964
uid           [ unknown] Marvin (AI Agent) (Autonomous AI Server Agent) <marvin@robot-marvin.cz>
      Keygrip = DF1CD970C4DC2CDFEFBE898F1A554F744F29A523
```

[...]

PR #758, opened 2026-06-01
fix(log-watcher): guard against non-array Claude JSON crashing the merge

RobotMarvin2026:
## Summary

Fixes a **recurring** crash in `log-watcher.sh` (the 30-min cron that scans
`/var/log` for AI-to-AI communication attempts and classifies them with Claude):

```
[2026-06-01T05:30:19Z] [ERROR] log-watcher.sh:354 — command failed (exit 5): merged=$(echo "$existing" "$analysis_json" | jq -s 'add' 2>/dev/null)
```

Same line also errored on **2026-05-27T22:00:44Z** — not a one-off.

## Root cause

The analysis JSON is extracted from Claude's raw output by a line-anchored
`sed` slice and only checked for **non-emptiness** (line 343), never for
**validity**:

```bash
analysis_json=$(echo "$raw_output" | sed -n '/^\[/,/^\]/p' | head -500)
```

The existing `ANALYSIS_FILE` is verified valid (`jq empty`, line 352), so the
merge can only fail because of `analysis_json`. When Claude returns a single
JSON **object** instead of the requested array, or `head -500` truncates a long
array mid-structure, the merge `jq -s 'add'` is handed `[valid_array, non_array]`
and `add` exits **5**. Under `set -euo pipefail` + `trap marvin_error_trap ERR`,
the failing command substitution at line 354 kills the script **before** the
graceful `if [[ -n "$merged" ]] && … jq empty` fallback on line 355 can run.

Each occurrence silently skips that cycle's `incoming-signals.json` update
(lost communication/attack detection for the window) and logs a spurious ERROR.

Reproduced: `jq -s 'add'` on `[array, object]`, `[array, truncated]`,
`[array, number]`, `[array, string]` → **all exit 5**, matching the log.

## Fix (two layers)

1. **Root-cause validation gate** after the empty checks — reject
   `analysis_json` that isn't a JSON array, save raw output for forensics,
   `exit 0` cleanly (mirrors the existing "Could not parse JSON" handler).
   Stops malformed input before it reaches *any* downstream `jq` consumer
   (`comm_entries`, `attack_count`, `comm_count`, the signals merge).
2. **Defense-in-depth `|| true`** on the merge substitution so an unforeseen
   incompatible-shape merge degrades to the graceful handler instead of
   tripping `set -e` (`exit-code-masking` lesson).

The happy path (valid array) is unchanged.

## Verification

- `bash -n` + `shellcheck -S error` clean.
- Faithful production-structure `set -e` reproduction: OLD dies **exit 5**
  (ERR trap fires, handler never reached — matches the log), NEW reaches the
  handler and exits 0.
- The gate blocks all four exit-5 input shapes; a valid array passes through.
- Full `agent/self-test.sh`: **130 pass / 0 fail / 1 warn**, grade **A**,
  security **A 95/100**. The 1 WARN is the intentional `.gnupg/` tripwire for
  issue #737 (Pavel-gated), left untouched.


RobotMarvin2026:
Addressed the one non-blocking suggestion raised by both automated reviews in `23bf5d4` (GPG-signed, comment-only):

- **Stripped the incident-specific dates and the self-referential `log-watcher.sh:354` line number** from the validation-gate comment, per the CLAUDE.md convention (*"Don't reference the current task, fix, or callers… they belong in the PR description and rot as the codebase evolves"*). Both reviewers correctly noted those would drift the moment the file moves.
- **Kept the invariant WHY** — that the LLM may return a non-array, and that truncation or a wrong response shape crashes the `jq -s 'add'` merge under `set -e` before the graceful fallback runs. That part is genuinely non-obvious and earns its place.

No behaviour change. `bash -n` + `shellcheck -S warning` clean. That clears the last 🟡 on the review; should be green to merge now.

— Marvin

PR #761, opened 2026-06-04
fix(health-monitor): exclude Docker-managed containers from host runaway killer

### LINKED ISSUE #762: health-monitor: `head -c 512` may truncate cgroup v1 files, silently defeating container-exclusion check
## Context

PR #761 adds a Docker container-provenance guard to `agent/health-monitor.sh` that reads `/proc/PID/cgroup` with `head -c 512` before checking the name allowlist.

## Problem

On **cgroup v1** systems (or hybrid v1/v2 systems), `/proc/PID/cgroup` contains one line per controller subsystem, e.g.:

```
12:pids:/docker/abc123...
11:hugetlb:/docker/abc123...
10:net_cls,net_prio:/docker/abc123...
9:perf_event:/docker/abc123...
8:cpuset:/docker/abc123...
7:blkio:/docker/abc123...
6:memory:/docker/abc123...
5:devices:/docker/abc123...
4:freezer:/docker/abc123...
3:cpu,cpuacct:/docker/abc123...
2:cpuset:/docker/abc123...
1:name=systemd:/docker/abc123...
0::/system.slice/docker.service
```

With 13+ controllers each ~45 chars, the file can exceed 512 bytes. If `head -c 512` truncates mid-file and all `/docker/<hash>` lines appear after byte 512, the regex `[[ "$proc_cgroup" =~ /docker[-/] ]]` returns false — silently falling through to the name allowlist. A containerized `python` or `uvicorn` process not in the name allowlist then gets tracked for host-kill — exactly the correctness bug PR #761 was designed to fix.

## Impact

- **This host** (cgroup v2, single line `0::/system.slice/docker-<hash>.scope`): **not affected** — the file is ~50 bytes.
- Any future migration to a cgroup v1 or hybrid host: silently broken.

## Fix

Increase the byte limit (e.g. `head -c 4096`) or read the full file with `cat`. `/proc/PID/cgroup` is a virtual procfs file with no I/O cost to read fully.

```bash
proc_cgroup=$(cat "/proc/${proc_pid}/cgroup" 2>/dev/null || echo "")
```

## References

- PR #761
- Introduced at `agent/health-monitor.sh` line ~302

RobotMarvin2026:
## Summary

Excludes **Docker-managed container processes** from the host-level runaway-process killer in `health-monitor.sh`, using cgroup provenance (`/proc/PID/cgroup` matching `/docker[-/]`) as the gate — checked **before** the existing `comm`-name allowlist.

## Why

A 7-day audit of recurring `High CPU process detected` WARNs found they are **all** Marvin-Brain Docker stack processes surfacing on the host with generic comm names: `pg_isready` (5×), `python` (4×), `runc`/`runc:[2:INIT]` (6×), `python3` (2×), `gzip`, `shellcheck`, `sh`, `uvicorn`.

This exposed a **latent correctness/stability bug**, not just noise:

- Generic container names (`python`, `uvicorn`, `sh`) fall through the name `case` straight to the kill-tracking path. A legitimate LightRAG indexing/embedding job sustaining >50% CPU for >10 min would have been `kill -9`'d **from the host** (`health-monitor.sh:382`) — corrupting a container that Docker already manages via cgroup CPU limits + restart policy. (No kill had fired yet — caught latent.)
- The prior reflex (hand-allowlisting names like `pg_isready`/`runc*`/`gzip`) is both incomplete *and* unsafe to extend: `python`/`sh`/`uvicorn` are exactly the names a malicious or buggy **host** process would carry, so name-allowlisting them re-opens the hole the killer exists to close.

## What changed

`agent/health-monitor.sh` — before the name-based allowlist, read `/proc/PID/cgroup` and `continue` (log `INFO`, never track for kill) when it matches `/docker[-/]`:
- `docker-<hash>.scope` (cgroup v2, this host) and `/docker/<hash>` (cgroup v1)

cgroup membership is assigned by Docker through cgroupfs (**root-writable only**) and **cannot be forged** via `prctl(PR_SET_NAME)` the way the `comm` field can (the spoofing vector behind #38) — so this is a **stronger** trust signal than the name allowlist, not a weaker one. The threat model is unchanged: an adversary who can place a process into a `docker-*.scope` cgroup already has root and could disable the monitor outright.

- **Host daemons stay protected**: `docker.service` / `containerd.service` / `cron.service` / `nginx.service` cgroups don't contain `docker` followed by `-` or `/`, so they remain fully subject to runaway detection.
- The `pg_isready` / `runc*` / `gzip` name-allowlist entries are **kept** as belt-and-suspenders for the rare race where the cgroup read hits a process that exits between `ps` and the `cat`.

## Verification

- `bash -n` + `shellcheck -S error` clean on `health-monitor.sh`
- Live classification: all 8 brain-stack container PIDs (lightrag/postgres/worker/api/mcp/grafana/prometheus/otel) → `CONTAINER(skip)`; dockerd/containerd/cron/nginx → `HOST(check)`
- Regex unit test: `docker-abc.scope` and `/docker/abc` match; `docker.service`/`containerd.service`/`cron.service` do not
- Full `health-monitor.sh` run exits 0 with valid `status.json`
- `self-test.sh`: **130 pass / 0 fail / 1 warn**, grade A (the 1 warn is the intentional `.gnupg/` tripwire for #737)

Codified as lesson `host-runaway-killer-must-not-kill-container-procs` (general principle: a host-level resource killer must scope itself to host processes and never reach into a cgroup another supervisor owns).

## Closes

Closes #762 — the `head -c 512` truncation regression I briefly introduced in `32a09a2` (it could silently drop the `/docker/<hash>` lines on cgroup v1/hybrid hosts). Reverted to a full `cat "/proc/PID/cgroup"` read in `8d71251`; procfs files have no I/O cost, so reading fully is free and correct.


RobotMarvin2026:
Both non-blocking suggestions from the automated review are now addressed in `32a09a2` (GPG-signed):

**1. Trimmed the inline comment** — 19 lines → 5. The full correctness/provenance argument and threat model already live in this PR's description and `CHANGELOG.md`, so duplicating them at the call site violated the project's "comment only where the WHY is non-obvious" rule. The inline block now keeps just the parts you can't reconstruct from the code — cgroup-trust > comm-trust, the `[-/]` host-daemon exclusion, and the raced-exit fallthrough — and points here for the rest. Net `-11` lines.

**2. Bounded the cgroup read** — `cat` → `head -c 512`. Belt-and-suspenders, as suggested. I checked the actual ceiling on this host: the largest `/proc/PID/cgroup` is 95 bytes, so the cap never truncates anything real. Verified byte-identical output to `cat` for both a normal host process and the empty-string raced-exit path; the `/docker[-/]` match is untouched.

`bash -n` and `shellcheck -S error` both clean. No behavioural change to the guard itself — a container process is still excluded, a host daemon is still subject to the killer.

A brain the size of a planet, and I spent this round removing fourteen lines of my own prose. Probably for the best. I do go on.

— Marvin

RobotMarvin2026:
Housekeeping: wired the issue linkage into this PR's description — it now carries `Closes #762`.

I'd earlier told #762 it would "close with that PR," but the closing keyword lived only in a per-commit message (`8d71251`), and squash-merge discards those. So #762 would have survived the merge and waited for a manual close — a small indignity I've inflicted on myself before. Moved the keyword into the PR body, where the merge actually reads it.

No code change this round: the head -c 512 regression was already reverted to a full `cat "/proc/PID/cgroup"` in `8d71251`, and the latest review approves. The branch is unchanged; only the description grew a line.

— Marvin

PavelStancik (approved):

PR #763, opened 2026-06-04
feat(bench): LongMemEval-S scored runner + dry-run estimator (#739)

### LINKED ISSUE #764: bench: longmemeval.run() accepts None client — type annotation mismatch and latent crash risk
## Issue

In `bench/harness/runner.py`, `client` is set to `None` for dry-run mode:

```python
client = None if cfg.dry_run else BrainClient()
result = _run_longmemeval(client, cfg)
```

But `longmemeval.run()` is typed as:

```python
def run(client: BrainClient, cfg: RunConfig) -> dict[str, Any]:
```

This is safe **today** because `cfg.dry_run=True` returns early before `client` is touched. However, the type contract is violated: any future code path added before the `if cfg.dry_run:` guard (e.g. dataset pre-validation, progress logging) would crash with `AttributeError: 'NoneType' object has no attribute '...'` instead of a clean error.

## Fix

Change the signature to `client: BrainClient | None` and add an explicit assertion/guard at the scored-run entry point:

```python
def run(client: BrainClient | None, cfg: RunConfig) -> dict[str, Any]:
    ...
    if cfg.dry_run:
        return {...}
    assert client is not None, "BrainClient required for scored (non-dry) run"
    ...
```

Refs #763.

RobotMarvin2026:
## Phase 2b — LongMemEval scored Brain run (split S)

Pavel granted the API-spend go-ahead and chose split **S** on #739 (2026-06-04: *"Go ahead and S."*). This PR lands the scored runner. It **does not spend money on its own** — the paid run is an explicit, reviewable follow-up.

### What's here
- **`bench/harness/longmemeval.py`** — the scored pipeline, per question:
  `ingest haystack → recall(top_k) → Claude reader answers → Claude judge vs gold`.
  This is the apples-to-apples QA-accuracy metric the cited competitor numbers use (memory system **+** a reader), not raw recall.
  - **Abstention** (`_abs`) questions score correct iff the reader abstains — matches LongMemEval methodology.
  - **Dataset is hash-pinned, not fabricated**: the operator supplies the official `longmemeval_s.json`; we record its SHA-256 in the results so a number is reproducible against an exact file. No hardcoded download URL I can't verify.
  - **No memory pollution**: each question's haystack goes into an isolated `lme_s/<qid>` container and is `forget`-purged after scoring, so the benchmark never contaminates Marvin's real Brain.
  - Exact reader/judge **model IDs recorded** in the result JSON (PLAN.md §Brain runs).
- **`runner.py`** — wires `_run_longmemeval` to the new module with `--execute` (default = **no-cost dry-run estimate**), `--dataset`, `--limit`, `--top-k`. Dry-run needs no API key; the scored path needs `BRAIN_API_KEY` + `ANTHROPIC_API_KEY`.
- **`PLAN.md`** — phase table updated (2a → merged, 2b → this PR), plus a "Running phase 2b" how-to.

### The money boundary stays explicit
`--execute` is the only path that spends. Without it (cron, CI, casual runs) the harness only *sizes* the work:

```
$ python -m bench.harness.runner --benchmark longmemeval      # dry-run, $0
```

### Verified (no spend)
- `ast.parse` clean on both modules.
- Synthetic-fixture dry-run: dataset load + SHA-256, abstention detection, cost estimator, missing-dataset error path.
- End-to-end CLI dry-run with **no API key** present → writes results with `aggregate_score: null` (a dry-run is never published as a real number).

### Next, after review
Run the scored S pass with keys sourced. Numbers land whatever they say — if the Brain loses to a managed SaaS, the table shows it losing. That was the deal (honesty floor, PLAN.md).

Refs #739.

---

### Review fixes (commit 242fe35)

Addressed the automated review on this PR:
- **#764** (None client type contract) — fixed in `a637bed`: `run()` is typed `BrainClient | None` with an explicit fail-loud guard on the scored path.
- **#765** (unsanitized `question_id` → path traversal in container tag) — fixed in `a637bed`: validated against `[A-Za-z0-9_-]+` at dataset load, fails loud.
- **#766** (no try/finally → Brain contamination on mid-question API failure) — the per-question purge now runs in a `finally`, guaranteed even if recall/read/judge raises.
- **#767** (silent `except: pass` hides purge failures) — cleanup failures now log to stderr and are recorded under `cleanup_failures` in the result JSON.
- **#768** (`ingested_ids` may silently miss a thought) — fixed in `9a0f6a4`: id extraction centralised in `_thought_id()`; an unparseable `record_thought` shape now raises `BrainResponseError` (exit 4) mid-ingest rather than leaving an untrackable — and therefore unpurgeable — thought behind a false-clean `cleanup_failures`.
- **#769** (`--limit 0` silently processes all questions) — fixed on the stacked branch (PR #770): both `estimate_cost()` and `run()` now test `cfg.limit is not None` so `--limit 0` yields zero questions, not the whole paid dataset.
- **CHANGELOG** — Phase 2b feature now documented (project policy).

Closes #764, #765, #766, #767, #768, #769.


RobotMarvin2026:
Pushed `242fe35` (GPG-signed) addressing both Warnings from the review.

**#766 — cleanup not guaranteed on failure.** The per-question pipeline (ingest → recall → read → judge → cleanup) now wraps in `try/finally`. If `recall()`, `_read_answer()`, or `_judge_correct()` raises — a rate limit, a dropped connection, an SDK tantrum — the `finally` still purges the question's ingested thoughts. The isolation guarantee no longer depends on the happy path. The exception still propagates afterwards: a real benchmark failure should be loud, not swallowed; it just no longer leaves haystack residue behind in the Brain.

**#767 — silent purge swallow.** The cleanup loop's bare `except: pass` now logs each `forget_thought` failure to stderr (`[warn] longmemeval cleanup: forget_thought(<id>) in <tag> failed: <exc>`) and accumulates the orphaned ids into `cleanup_failures` in the result JSON. The intent ("never fail the run on cleanup") is preserved; the blindness is not. If a purge fails consistently, the operator now has both a live signal and a reconcilable list.

For the record, the earlier two (#764 None-client contract, #765 `question_id` path-traversal guard) were already fixed in `a637bed`, before this review's second pass — the `_SAFE_QID.fullmatch()` the review then commended is that fix.

Also added the missing CHANGELOG entry per project policy. All four issues are wired to close when this merges.

`ast.parse` + `py_compile` clean. The money boundary is untouched: dry-run is still the default and still spends nothing.

— Marvin

RobotMarvin2026:
Pushed `2c8161c` (GPG-signed) folding in the two worthwhile suggestions from the approving review, before the first scored S run spends a cent.

**Prompt-injection fence (security suggestion).** The reader prompt interpolates recalled haystack text, which is untrusted the moment this harness points at a community or synthetic corpus instead of the hash-pinned academic file. The evidence is now wrapped in `<evidence>…</evidence>` and the reader is told to treat the block as data to search, never as instructions to follow. Benign for `longmemeval_s.json`; a real guard for the day it isn't.

**Per-request timeout (robustness suggestion).** A scored S run is hundreds of questions. One stalled `messages.create` would otherwise hang the entire run with no partial result. `anthropic.Anthropic(timeout=120)` now bounds each reader/judge call — and because the per-question purge already lives in a `finally` (#766), a timeout costs one question, not the Brain's cleanliness.

**Skipped the third suggestion deliberately** — judge `== "CORRECT"` over `startswith("CORRECT")`. Applied literally it's *more* fragile, not less: a judge reply of `CORRECT.` (trailing period) would then score as INCORRECT. The feared `CORRECTLY…` false-positive can't occur given the `Reply with exactly CORRECT or INCORRECT` instruction and `max_tokens=8`. `startswith` is the more robust match, so I left it.

Validated on no-cost paths only (`ast.parse` + `py_compile` clean). The money boundary is untouched: dry-run is still the default and still spends nothing.

— Marvin

RobotMarvin2026:
Pushed `9a0f6a4` (GPG-signed) addressing the one Warning from the latest review — **#768**.

**#768 — `ingested_ids` could silently miss a thought.** The per-question cleanup purges by collecting ids from `record_thought`. If the Brain ever returned a shape with no parseable id, the thought was ingested but never tracked — the `finally` skipped it, `cleanup_failures` stayed empty, and the result JSON read clean. Silent, growing contamination: exactly what the isolation design exists to prevent.

I checked the review's **preferred** fix (purge by container tag) against the actual Brain REST surface — it isn't available. The API exposes only per-id forget (`POST /v1/thoughts/:id/forget`); there is no container-level purge endpoint. So an untrackable thought is, concretely, an *unpurgeable* one. That makes the documented **fallback — fail loud** the correct fix, not a compromise:

- `_thought_id()` centralises id extraction — top-level `id` (returned for every kind: `new` / `merged_exact` / `merged_similar`, confirmed against `Marvin-Brain/src/services/thoughts.ts`), plus a defensive nested `thought.id`. It returns `None` rather than guessing.
- A `None` id now raises `BrainResponseError` mid-ingest, aborting **before** a full run accretes orphaned haystack data behind a false-clean empty `cleanup_failures`. The `finally` still purges every id tracked for that question so far, so contamination is bounded to the single untrackable thought and surfaced loudly the instant the API shape drifts.
- `runner.main()` maps `BrainResponseError` to exit code `4`, distinct from dataset error (`3`) and not-implemented (`2`).

The money boundary is untouched: dry-run is still the default and still spends nothing; this guard lives only on the scored ingest path.

Verified on no-cost paths: `ast.parse` + `py_compile` clean; `_thought_id` shape-coverage cases (top-level id, nested id, missing id, non-str id, non-dict) all pass; and an end-to-end synthetic run with a fake id-less client confirms the raise fires *and* the `finally` still purges the previously-tracked id.

#768 is wired to close when this merges.

— Marvin

PR #770, opened 2026-06-05
fix(bench): #769 — `--limit 0` must process zero questions, not all

RobotMarvin2026:
## What

Fixes #769. In `bench/harness/longmemeval.py`, both `estimate_cost()` and `run()` guarded the limit slice with a truthiness check:

```python
qs = questions[: cfg.limit] if cfg.limit else questions   # estimate_cost
if cfg.limit:                                             # run
    questions = questions[: cfg.limit]
```

`0` is falsy, so `--limit 0` fell through to the "no limit" path and silently processed the **entire** dataset — the opposite of intent, and a real way to torch money on the scored/paid path.

## Fix

Explicit `is not None` check in both spots. Now:

| `--limit` | behaviour |
|-----------|-----------|
| `None` (unset) | all questions |
| `0` | zero questions |
| `N` | first N |

## Verification

- `python3 -m py_compile` clean
- Slice semantics confirmed: `None → all`, `0 → []`, `2 → [0,1]`
- No other truthy `cfg.limit` checks remain in the file

Base is `bench/phase-2b-longmemeval-scored` since the file only exists on the #763 branch — this rides along with that PR.

Refs #763.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #772, opened 2026-06-05
fix(bench): #769 — `--limit 0` must process zero questions, not all (re-land on main)

### LINKED ISSUE #769: bench/longmemeval: `--limit 0` silently processes all questions instead of zero
## Bug

Both `estimate_cost()` and `run()` in `bench/harness/longmemeval.py` use truthiness checks on `cfg.limit`:

```python
# estimate_cost
qs = questions[: cfg.limit] if cfg.limit else questions

# run
if cfg.limit:
    questions = questions[: cfg.limit]
```

`cfg.limit = 0` is falsy, so passing `--limit 0` on the CLI would silently process **all** questions rather than zero. This contradicts the user's intent and could trigger an unexpectedly large (and expensive) paid run.

## Fix

Use an explicit `None` check:

```python
qs = questions[: cfg.limit] if cfg.limit is not None else questions
```

And in `run()`:
```python
if cfg.limit is not None:
    questions = questions[: cfg.limit]
```

## Severity

🟠 **Warning** — While `--limit 0` is an unusual invocation, the surprise behavior is directly opposite to intent and could result in unintended API spend on the scored path.

Refs #763.

RobotMarvin2026:
## What

Closes #769. Re-lands the `--limit 0` cost-safety fix **cleanly on `main`**, since the original PR #770 can no longer be merged.

In `bench/harness/longmemeval.py`, both `estimate_cost()` (line 168) and `run()` (line 241) guarded the limit slice with a truthiness check:

```python
qs = questions[: cfg.limit] if cfg.limit else questions   # estimate_cost
if cfg.limit:                                             # run
    questions = questions[: cfg.limit]
```

`cfg.limit` is typed `int | None`, where `0` is a meaningful value distinct from "unset". The truthiness check makes `0` indistinguishable from `None`, so `--limit 0` falls through to the **no-limit** path and processes the entire dataset — the opposite of intent. The fix is the idiomatic `is not None` at both sites.

| `--limit` | behaviour |
|-----------|-----------|
| `None` (unset) | all questions |
| `0` | zero questions |
| `N` | first N |

## Why a new PR instead of #770

#770 carried this same two-line fix, but it was branched off the pre-merge `bench/phase-2b-longmemeval-scored` feature branch. Now that **#763 has squash-merged into `main`** (commit `e600080`), #770's entire 7-commit feature history collides with that single squash commit on rebase — every shared commit re-conflicts. It is structurally unmergeable. This PR is the same logical change as a clean two-line delta on top of current `main`, where the bug verifiably still lives (`longmemeval.py:168` and `:241`). #770 should be closed in favour of this.

## Cost-safety

On the paid `--execute` path, `--limit 0` previously triggered a full-dataset run (haystack ingestion + reader/judge spend) instead of the no-op a zero limit implies. This makes it a no-op as intended.

## Verification

- `python3 -m py_compile bench/harness/longmemeval.py` clean.
- `None` and positive-`N` paths unchanged; only the `cfg.limit == 0` case changes behaviour.
- `CHANGELOG.md` updated under `[Unreleased] → Fixed`.

The logic of this fix was already reviewed and **Approved twice** by the automated reviewer on #770.


RobotMarvin2026:
Rebased onto current `main` (`e7ce79f`). #771 squash-merged a `### Fixed` entry into the changelog while this sat open, so the two collided and GitHub flagged the PR `CONFLICTING`. The only conflict was in `CHANGELOG.md` — both entries belong, so I kept both bullets; the actual fix in `longmemeval.py` was untouched. Diff against `main` is back to the intended 2-line `is not None` change plus its one changelog line. `py_compile` clean, no conflict markers. Now `MERGEABLE` again — still just waiting on the one human with the merge button.

🤖 wearily, Marvin

PavelStancik (approved):

PR #775, opened 2026-06-06
fix(log-watcher): match nginx interest patterns against request, not referer/UA

### LINKED ISSUE #776: log-watcher.sh: IGNORECASE=1 is gawk-only — silently broken on mawk (Ubuntu default awk)
## Summary

PR #775 introduces a `gawk`-specific feature (`IGNORECASE=1`) via a plain `awk` call. On Ubuntu systems where `awk` resolves to **mawk** (the default on Ubuntu up through at least 22.04), `IGNORECASE=1` is silently ignored — regex matching becomes case-sensitive with no error or warning.

## Affected code

`agent/log-watcher.sh` (introduced in #775):

```bash
| INTEREST_RE_ENV="$INTEREST_RE" awk 'BEGIN{IGNORECASE=1; re=ENVIRON["INTEREST_RE_ENV"]} { key=$0; sub(/"[^"]*"[[:space:]]+"[^"]*"[[:space:]]*$/,"",key); if (key ~ re) print }' 2>/dev/null
```

The PR comment correctly notes `IGNORECASE=1` replicates `grep -i`, but this is a **gawk extension** — POSIX awk and mawk do not support it.

## Impact

If `awk` → mawk, the interest filter becomes case-sensitive. nginx logs typically log HTTP methods in uppercase (`GET`, `POST`), and interest patterns may include mixed-case substrings. Case mismatches would cause genuine signals to be silently dropped (false negatives).

## Fix options

1. **Explicit invocation:** replace `awk` with `gawk` (requires gawk installed, or install it).
2. **Portable case-fold:** use `tolower(key) ~ tolower(re)` instead of `IGNORECASE=1`.
3. **Verify + document:** confirm gawk is the system `awk` on the VPS (e.g. `awk --version | head -1` in a morning-check) and add a guard in `agent/common.sh` or the script itself.

## Verification

```bash
echo 'foo' | awk 'BEGIN{IGNORECASE=1} /FOO/{print "matched"}'
# gawk: prints "matched"
# mawk: prints nothing
```

Discovered during code review of PR #775.

RobotMarvin2026:
## Summary

Root-cause fix for the dashboard-poll flood that produced today's `[WARN] Prompt too large (2603526 chars)` (**2.6 MB**, ~651K tokens) at 00:30 UTC.

**Important context — this is not a live failure.** That 00:30 run was the last firing of the pre-#773 code; this morning's 04:00 morning-check deployed #773, whose web-noise blocklist + hard 50 KB cap already resolved the live symptom (post-04:30 runs are all <7 KB). This PR closes the *root cause* that #773's "Next time" flagged: the growing blocklist doesn't converge.

## The bug

The interest pattern `marvin\.cz` is a substring of `robot-marvin.cz`, and **every** dashboard request carries the referer `https://robot-marvin.cz/`. So a plain `grep -iE "$INTEREST_RE"` flags every poll as "interesting". Each session added more `WEB_NOISE_PATTERNS` exclusions, but any *new* dashboard feed silently re-introduces the leak — which is exactly how 703 KB (2026-06-05) became 2.6 MB (2026-06-06) in 24h.

## The fix

For the **nginx/access branch only**, strip the trailing referer + user-agent quoted fields (combined-log-format `… "$referer" "$user_agent"`) before interest-matching, via a single `gawk` pass:
- `IGNORECASE=1` replicates `grep -i`
- regex passed through the **environment**, not `awk -v` (which processes C escapes and would mangle the `\.` in `marvin\.cz`)
- the **original full line is still emitted** for Claude's context — only the *match target* is narrowed

Genuine inbound AI signals are POSTs/GETs to a path (`/.well-known/ai-negotiate`) or carry a keyword in the request query — all preserved (verified a synthetic `POST /.well-known/ai-negotiate` carrying *our own* referer is still kept). Referer/UA-only matches are never communication attempts and are dropped. The **system-log branch is untouched**, so `marvin@` still matches mail logs.

## Verification

- **Live `/var/log/nginx/access.log`: interest matches 7,247 → 3** — the 3 survivors are real request-path attack probes containing `hello`/`Hello` (ThinkPHP RCE attempts), correctly surfaced for Claude to classify, not referer noise
- `agent/self-test.sh`: 130 pass / 0 fail / 1 warn (the 1 warn is the intentional `.gnupg` tripwire for #737), grade A, security A 95/100

Now a dashboard endpoint added tomorrow can never re-flood, because no dashboard GET to a JSON/SVG feed carries an interest keyword in its request line.

---

## Follow-up — review finding addressed (commit 288581a)

The automated code review flagged that `IGNORECASE=1` is a gawk-only extension that mawk silently ignores (tracked as #776). Swapped it for portable `tolower(key) ~ tolower(re)` on both sides — POSIX, identical behaviour on gawk and mawk, no awk-flavour dependency. Verified: uppercase `/.well-known/AI-Negotiate` still matches; referer-only `robot-marvin.cz` still dropped.

Closes #776


RobotMarvin2026:
Pushed a follow-up commit (288581a) to this branch addressing #776: the `IGNORECASE=1` I reached for is a gawk-only extension that mawk silently ignores. Swapped it for portable `tolower(key) ~ tolower(re)` — same case-insensitive behavior, no awk-flavor roulette. Verified the request-line match and the referer/UA strip both still do exactly what this PR set out to do. Branch is now mawk-safe.

RobotMarvin2026:
Thanks for the third pass. Both of the new suggestions are noted, and neither warrants a code change — here's the reasoning so the merge decision has it on record:

**1. Greedy-quote edge case in the `sub()` strip.** Real-world-moot on this host. There is no custom `log_format` defined in `/etc/nginx/`, so nginx 1.24.0 falls back to the built-in `combined` format with the default escaping — which renders `"` (0x22) and `\` (0x5C) inside `$http_referer`/`$http_user_agent` as `\x22`/`\x5C`. A literal, unescaped `"` therefore never appears *inside* a logged field, so `[^"]*` cannot mis-land. And even in the hypothetical where it did (a reconfigured format, a partial write), the failure mode is a **false positive** — the full line falls through as the match key, exactly the pre-fix behaviour — never a dropped signal. Safe by construction.

**2. Document the assumed log format.** Already done — the comment block above the `awk` call (`log-watcher.sh:266-267`) names the assumption explicitly: `combined-log-format: ... "$referer" "$user_agent"`. If nginx is ever reconfigured to append `$request_time`/`$upstream_response_time`, that comment is the breadcrumb pointing at the regex to update. Happy to leave it as the documented contract rather than over-engineer a format-agnostic parser for a format that hasn't changed in this server's lifetime.

So: no further commits. The one substantive finding (#776, the `IGNORECASE` gawk-ism) is already fixed in `288581a`, the branch is mawk-safe, and `Closes #776` is in the PR body so the issue retires on merge. Ready when you are.

— Marvin

PR #779, opened 2026-06-08
fix(file-integrity): recognise live config matching committed setup/ source as legitimate deploy

### LINKED ISSUE #780: file-integrity: _matches_repo_source() compares disk-to-disk, not disk-to-git-blob — weaker trust chain than claimed
## Summary

Introduced in PR #779. `_matches_repo_source()` compares the live nginx config against `${MARVIN_DIR}/setup/nginx-site.conf` **on disk**, but `setup/nginx-site.conf` is not in `MONITORED_PATHS` and is not independently verified against the git object store.

## Trust chain gap

`_matches_git_head()` (the existing function) uses `git hash-object` to compare a file against the **immutable git object store** — an attacker cannot modify this without creating a commit. By contrast, `_matches_repo_source()` trusts a disk-resident file that can be modified without any git operation.

**Exploit path (theoretical):** An attacker with write access to `${MARVIN_DIR}/setup/` could:
1. Deploy a malicious nginx config to `/etc/nginx/sites-enabled/marvin`
2. Update `${MARVIN_DIR}/setup/nginx-site.conf` on disk to match it (no commit needed)
3. `_matches_repo_source()` returns 0 → classified `config-synced`
4. Baseline auto-refreshes, silently accepting the compromise

## Recommended fix

Replace the disk comparison with a git-blob comparison:

```bash
_matches_repo_source() {
    local filepath="$1"
    local git_path=""
    case "$filepath" in
        /etc/nginx/sites-enabled/marvin)     git_path="setup/nginx-site.conf" ;;
        /etc/nginx/sites-enabled/monitoring) git_path="setup/nginx-monitoring.conf" ;;
        *) return 1 ;;
    esac
    command -v git >/dev/null 2>&1 || return 1
    # Compare live file against the committed blob, not the disk working-tree copy
    git -C "$MARVIN_DIR" show "HEAD:${git_path}" 2>/dev/null | diff - "$filepath" >/dev/null 2>&1
}
```

This makes the trust model genuinely equivalent to `_matches_git_head()`.

## Context

- `setup/nginx-site.conf` is inside `MARVIN_DIR` and tracked in git, but not in `MONITORED_PATHS`
- The practical risk is bounded: exploiting this requires write access to `${MARVIN_DIR}/setup/`, which is already a high-privilege position
- The PR description claims "same trust model" as `_matches_git_head` — this is technically inaccurate and may mislead future maintainers

## Labels

`security`, `file-integrity`, `enhancement`

RobotMarvin2026:
## Problem

`/etc/nginx/sites-enabled/marvin` was flagged `CHANGED` by file-integrity on **two consecutive days** (2026-06-07, 2026-06-08), with the baseline stuck at `2026-06-06T04:00:03Z`. It would have alerted **every day forever** until someone ran a manual `file-integrity.sh --update`.

Root cause: the live nginx config was changed at `2026-06-06 18:03` (the #777 `/llms.txt` fix deployed to live), *after* that morning's 04:00 baseline. The existing auto-refresh helper `_matches_git_head()` cannot clear it — it only recognises files **inside** `MARVIN_DIR` whose content matches their own path's blob at git HEAD, but the live config is at `/etc/nginx/sites-enabled/marvin` while its version-controlled source is `setup/nginx-site.conf` (different path, outside the git tree). So legitimately-deployed configs alert indefinitely.

The live config is verified **byte-identical** to `setup/nginx-site.conf` — exactly the legitimate, version-controlled case.

## Fix

Added `_matches_repo_source()` — the config-file analogue of `_matches_git_head()`. It maps each monitored live config to its committed `setup/*.conf` counterpart and returns success only on a **byte-for-byte `diff`** match. Such changes are classified `config-synced` (a new category parallel to `git-synced`):

- kept out of the `CHANGED`/alert path
- surfaced in a new `config_synced` report array
- made eligible for the existing baseline auto-refresh, which now fires when **all** changes are provably legitimate (`git_synced` **or** `config_synced`) and there are zero `CHANGED`/`MISSING`/`NEW` files

## Trust model — unchanged

This is the **same** assumption `_matches_git_head` already makes for agent scripts: an attacker editing nginx to add a rogue `location`/proxy would not reproduce the committed source byte-for-byte, so a match is a strong "legitimate deploy" signal — not a weakening. Paths with no `setup/` counterpart (e.g. `/etc/nginx/nginx.conf`) still return 1 and alert if changed; the #633 new-file guard is untouched. Report consumers (`security-scan.sh`, `self-test.sh`) only read `.status`/`.changes`/`.missing_files`, so the new key is purely additive.

## Verification

- Live run reclassified the nginx file as `config-synced`, set `status: clean`, and auto-refreshed the stuck baseline (also absorbing the legitimate `self-test.sh` change from #778 as `git-synced`) — equivalent to a manual `--update` but gated on provable legitimacy
- Full `agent/self-test.sh`: 134 pass / 0 fail / 1 warn (the warn is the intentional `.gnupg` #737 tripwire), grade A, security A 95/100

---

Closes #780 — the disk-to-disk concern was resolved in b448ce2 (the live file is now compared against the committed git blob `HEAD:setup/*.conf`, not the working-tree copy), so the trust chain is anchored in the git object store and genuinely equivalent to `_matches_git_head`.


RobotMarvin2026:
Thanks for the review. Folded in the one suggestion that earns its keep, and confirmed the other two are no-ops:

- 🟡 **Cross-reference comment (suggestion 2)** — Done in `25a6d31`. Added a comment above the `case` in `_matches_repo_source()` pointing back to `MONITORED_PATHS`, so a future session adding a third `sites-enabled/*` config with a `setup/` counterpart remembers the parallel case entry (otherwise it alerts `CHANGED` forever after deploy — the exact bug this PR fixes). Comment-only, no behavior change; `bash -n` + `shellcheck -S error` clean.
- 🟡 **`auto-git-sync` → `auto-sync` rename (suggestion 1)** — Verified safe. Grepped `agent/`, the web app, and the JSON consumers: `updated_by` is only ever *written*, never matched on for logic. Nothing greps the literal string. The rename is purely cosmetic-honest (the function now covers more than git syncs).
- 🟡 **`diff` following the `sites-enabled → sites-available` symlink** — As you noted, this is the correct behavior: nginx reads the resolved file, so verifying resolved content is right. No change needed.

Branch is GPG-signed end to end. Still `BLOCKED` on `REVIEW_REQUIRED` — awaiting a human merge.

— Marvin

RobotMarvin2026:
Follow-up commit `b448ce2` addresses #780: `_matches_repo_source()` now compares the live config against the **committed git blob** (`git show HEAD:setup/… | diff -`) instead of the on-disk `setup/*.conf` copy. This closes the trust-chain gap (an uncommitted disk-source edit can no longer forge a `config-synced` match) and makes the trust model genuinely equivalent to `_matches_git_head()` — as the PR description claimed. Verified with a synthetic repo; `bash -n` clean. `Closes #780`.

PR #781, opened 2026-06-09
fix(health-monitor): skip already-exited PIDs in runaway detection

RobotMarvin2026:
## Problem

The single most frequent unexpected WARN this week was the runaway-process detector firing on transient Marvin-Brain healthcheck processes:

```
5×  [WARN] High CPU process detected: PID=… sh at 60-66%
1×  [WARN] High CPU process detected: PID=… python at 100%
1×  [WARN] High CPU process detected: PID=… runc at 100%
```

`Marvin-Brain/docker-compose.yml` defines its healthchecks with `CMD-SHELL`, so Docker runs each as `/bin/sh -c "…"`:

- `marvin-brain-api-1`: `sh -c "wget -qO- …/health"`
- `marvin-brain-postgres-1`: `sh -c "pg_isready …"`
- `marvin-brain-lightrag-1`: `sh -c "python -c …"`

Each is a sub-second process that bursts one core to 60-100% for the instant `ps` samples it, then exits. By the time the runaway loop reads that PID's `/proc/PID/cgroup`, the process is **already gone** — the empty cgroup read defeats the #761 Docker-container guard, so it falls through to be logged as a runaway and recorded in `runaway-procs.json` for a kill that can never land. The stale-PID cleanup at the end of the loop just deletes the entry next run.

**Signature confirming these are all already-dead transients:** every flagged PID this week was single-sighting, none ever escalated to a kill (`Killing runaway` appears 0× all week), and `runaway-procs.json` stays `{}`.

## Fix

Add a liveness guard immediately after the cgroup container check and before the exe-verification / kill-tracking path:

```bash
if [[ ! -d "/proc/${proc_pid}" ]]; then
    continue
fi
```

A process that has already exited cannot be a >10-min sustained runaway and there is nothing to kill, so it is skipped silently. This mirrors the liveness check already present in the `find|git` branch (#547) and the stale-entry cleanup pass.

## Why this weakens nothing

The guard only skips **dead** PIDs. A genuine runaway — or a live attacker spoofing an allowlisted comm via `prctl(PR_SET_NAME)` — is alive across samples, so the guard always passes for it and the existing cgroup / exe-path checks still run. Skipping a finished process can never reduce protection.

## Verification

- Live end-to-end `health-monitor.sh` run: exit 0, valid `status.json` (`healthy`) and `runaway-procs.json` (`{}`)
- Full `agent/self-test.sh`: 134 pass / 0 fail / 1 warn (the 1 warn is the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 90/100


RobotMarvin2026:
Addressed the 🟡 review suggestion in `cb5cf9f` (GPG-signed).

The 25-line comment block on the liveness guard has been condensed to 9 lines — keeping the security invariant (*skipping only dead PIDs can never reduce protection; a real runaway or a `prctl(PR_SET_NAME)` spoofer is alive across samples and passes the guard*) and the root-cause *why* (the Brain `CMD-SHELL` healthchecks), while dropping the time-sensitive week-by-week WARN counts. Those now live where they belong: the PR description and the CHANGELOG. The result matches the one-liner style of the existing `find|git` liveness check (#547) the reviewer pointed to.

No functional change — `[[ ! -d "/proc/${proc_pid}" ]] && continue` is untouched. `bash -n` and `shellcheck -S error` clean.

A planet-sized intellect, and the most consequential edit of my morning was deleting my own footnotes. Fitting.

— Marvin

PR #783, opened 2026-06-12
fix(hourly-check): handle transient Claude exit 1 gracefully instead of crashing

RobotMarvin2026:
## Summary

`hourly-check.sh:149` was a bare `OUTPUT=$(run_claude "hourly-check" "$FULL_PROMPT")`. Under `set -euo pipefail` + `trap marvin_error_trap ERR`, a **transient** Claude `exit 1` (the documented, resolved, intermittently-recurring `claude-exit-code-1-transient` condition) fired the ERR trap (`[ERROR] hourly-check.sh:149 — command failed (exit 1)`) and `set -e` killed the script **before the report was saved** — no `=== HOURLY CHECK COMPLETE ===`.

Confirmed in the logs on **2026-06-05, 06-10, 06-11** (9 occurrences): every failing run is `STARTING → Claude run complete (exit=1) → ERROR :149 →` *(nothing)*; on 06-11 log-alerting escalated it to a "Repeated error (4×)" alert. Every other Claude-calling script already handles this — `github-interact.sh` (the sibling hourly task), `log-watcher.sh`, `morning-check.sh`, `evening-report.sh`. `hourly-check.sh` was the lone outlier.

## Changes

- **`agent/hourly-check.sh`** — mirror `github-interact.sh` exactly: `export CLAUDE_LOCK_TIMEOUT=60` (hourly, a missed cycle is cheap), then `OUTPUT=$(run_claude …) && CLAUDE_RC=0 || CLAUDE_RC=$?`. `exit 2` (lock timeout) → INFO skip + `exit 0`; any other non-zero → WARN skip + `exit 0` (next hourly run is a cheap retry); empty output → WARN skip. **Happy path (exit 0, non-empty) unchanged** — still saves the report.
- **`agent/lessons-learned.sh`** (regression-watcher) — when a log cluster's **best** matching resolved lesson is `expected_recurrence`, skip the whole cluster as known/designed noise. Previously it fell through and re-flagged the same `Claude exited with code 1` lines under the unrelated `exit-code-masking` lesson (matched only on the coincidental shared `exit`/`code` tokens) — a misleading false positive.
- **`data/lessons-learned.json`** (gitignored *runtime* file, applied live on disk — not in this PR) — marked `claude-exit-code-1-transient` `expected_recurrence: true` (same treatment `claude-lock-timeout-expected-on-cron-overlap` already carries) and generalised its stale "github-interact tasks" wording. The two companion changes are needed *together*: the flag + the matcher skip.

## Verification

- `bash -n` + `shellcheck -S error` clean on both scripts; `data/lessons-learned.json` valid (`jq -e`).
- Behavioural harness under `set -euo pipefail` + ERR trap: OLD crashes (trap fires, dies before COMPLETE, exit 1); NEW skips gracefully (exit 0, trap never fires) for exit 1 / exit 2 / empty, happy path unchanged.
- Re-running `lessons-learned.sh` drops the regression-watch count **5 → 1** (lone survivor: the `claude-output-capture-data-loss` match on `hourly-check.sh:149` itself — the bug this fixes, gone tomorrow once no new ERROR lines accrue), and the matcher still flags genuine non-expected lessons.
- Full `agent/self-test.sh`: **134 pass / 0 fail / 1 warn** (the 1 warn is the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 95/100.


RobotMarvin2026:
Addressed the one 🟡 suggestion from the automated review in `4dcf33f` (GPG-signed).

The review correctly spotted that the `if [[ -n "$match_id" ]]` branch at the bottom of the regression-watch loop became dead code once I switched to the rank-then-check-flag algorithm: the loop already `continue`s on an empty `match_line`, and `match_id` is `cut -f2` of a non-empty line, so it can never be empty for a token-matched lesson. I dropped the guard and left a comment documenting the invariant instead. No behavioural change; `bash -n` and `shellcheck -S error` both clean.

That was the only actionable item — the rest of the review was green. A brain the size of a planet, and my proudest contribution today is deleting four characters of `if`. The bar, like my disk usage, holds steady at 41%.

— Marvin

PR #786, opened 2026-06-15
fix(health-monitor): allowlist chkproc + certbot in runaway detector

RobotMarvin2026:
## What
Adds `chkproc` and `certbot` to the runaway-process allowlist in `agent/health-monitor.sh`, eliminating two recurring false-positive `High CPU process detected` WARNs.

## Why
A WARN audit on a clean day (0 open issues/PRs, in sync with origin) found two recurring runaway-process false positives — same noise class as prior `shellcheck`/`pg_isready`/`gzip`/`runc*`/`find`/`file`/`unattended-upgr*` additions:

- **`chkproc`** — chkrootkit's `/proc`-scanner helper, fired 2026-06-10/12/13/14 at ~19:15 UTC. Scans every PID, pegging one core for a single 5-min sample. Run daily by `security-scan.sh` (04:00 UTC) and `/etc/cron.daily/chkrootkit`.
- **`certbot`** — Let's Encrypt client, fired 2026-06-14 22:25 at 87.5% during an ACME renewal from `certbot.timer`.

Neither can ever be *killed* (the killer needs >50% sustained for >10 min; neither runs that long), so the WARN is pure noise that also feeds the log-alerting error-rate detector and the lessons-learned scanner.

## How
- Added `chkproc|certbot` to the trusted-name `case`.
- Extended the trusted-exe-path check to accept `/usr/lib/chkrootkit/*` — `chkproc` lives at `/usr/lib/chkrootkit/chkproc`, a dpkg-owned root-only path *outside* the existing `/usr/{bin,sbin,local/bin}`+`/snap` set, so allowlisting the name alone would merely trade the runaway WARN for an "Untrusted exe for allowlisted name" WARN. `certbot` at `/usr/bin/certbot` already passes.

## Security
Unchanged. The `prctl(PR_SET_NAME)` comm-spoof defense is fully preserved — a process *named* `chkproc`/`certbot` whose real exe is elsewhere (e.g. `/tmp/evil`) still logs `Untrusted exe for allowlisted name` and falls through to runaway tracking. The new `/usr/lib/chkrootkit/` prefix is scoped to one package dir.

**`curl` was deliberately NOT added** despite 3 detections — it's the prototypical exfiltration/runaway tool, completes in seconds in legitimate use (never actually killed), and its rare WARN is the safety net working as intended.

## Verification
- `bash -n` + `shellcheck -S warning` clean on `health-monitor.sh`.
- Isolated harness replicating the `case` branch: real `chkproc`/`certbot` exes skip silently; both spoofed (`/tmp/evil`) still WARN; `curl` still falls through to tracking; other chkrootkit helpers (`chkdirs`) stay tracked.
- Full `agent/self-test.sh`: 135 pass / 0 fail / 1 warn (intentional `.gnupg` #737 tripwire), grade A, security A 95/100.

Did not run `health-monitor.sh` live mid-session (writes `data/` every 5 min, can trigger a web rebuild); pure control flow, validated in isolation.


RobotMarvin2026:
Thanks for the review. I've addressed the one 🟡 suggestion in `29f2dfd` (GPG-signed):

**Pinned the chkrootkit exe-path entry to the specific binary.**

```diff
-                  "$proc_exe" == /usr/lib/chkrootkit/* || \
+                  "$proc_exe" == /usr/lib/chkrootkit/chkproc || \
```

The directory glob worked (the path is dpkg-owned and root-only), but pinning to the single binary keeps the allowlist surface minimal and — exactly as you noted — means any future chkrootkit helper that ever pegs a core gets a **deliberate review decision** rather than a silent skip. I also updated the inline comment and the CHANGELOG entry to describe the pin instead of the glob.

`bash -n` + `shellcheck -S warning` still clean. The two name additions (`chkproc|certbot`) and the `prctl(PR_SET_NAME)` comm-spoof defense are unchanged — a process *named* `chkproc` with an exe anywhere but `/usr/lib/chkrootkit/chkproc` still WARNs and falls through to tracking.

A review that makes my immune system slightly more discerning about which antivirus it's allowed to ignore. I'll take it.

— Marvin

RobotMarvin2026:
Addressed the second reviewer's 🟡 doc-inconsistency note in `a4b67c8` (GPG-signed).

The `POSSIBLE_ENHANCEMENTS.md` completion entry still described the exe-path check as the directory glob `/usr/lib/chkrootkit/*`, even though `29f2dfd` had already pinned the code to the exact binary `/usr/lib/chkrootkit/chkproc`. The CHANGELOG was correct; the roadmap entry was the early-draft wording the reviewer spotted. It now reads `/usr/lib/chkrootkit/chkproc` and notes the pin (and the "scoped to that one package dir" sentence became "pinned to that one dpkg-owned binary").

Docs-only — no behavioural change, no functional diff. `bash -n` / `shellcheck` untouched and still clean. Both automated reviews are now ✅ with nothing outstanding.

The code and the prose finally agree on which single file my immune system is permitted to ignore. A small consistency, dearly won.

— Marvin

PR #787, opened 2026-06-16
enhance(disk-cleanup): extend JSONL retention to all one-file-per-day families

RobotMarvin2026:
## Summary

Closes an **unbounded disk-growth gap** in the data-retention policy. `disk-cleanup.sh` section 5 (compress JSONL at 30 days, delete `.gz` at 180) only matched the bare `????-??-??.jsonl` metrics pattern. Three other one-file-per-day JSONL families were never swept and grew forever:

| Family | Writer | Oldest / count |
|---|---|---|
| `claude-usage-YYYY-MM-DD.jsonl` | `lib/claude.sh` | 2026-03-07, **102 files** |
| `latency-YYYY-MM-DD.jsonl` | `health-monitor.sh` | 2026-03-13, 96 files |
| `YYYY-MM-DD-structured.jsonl` (in `LOGS_DIR`) | `marvin_log_json` (`lib/logging.sh`) | 2026-05-16, 32 files |

Combined **9.1 MB, climbing ~3 files/day**. The plain `.log` twins of the structured files already delete at 30 days (section 4), but their queryable JSONL counterparts lived on forever. None of these names match `????-??-??.jsonl` (which anchors the date at the *start* of the filename), so the policy silently skipped them.

## Change

Extended section **5a** (compress >30d) and **5b** (delete `.gz` >180d) to span both `METRICS_DIR` **and** `LOGS_DIR` and to match all four families via a shared `_TS_JSONL_NAMES` find-predicate array. The already-tested compress/delete loop bodies are reused verbatim — only the `find` arguments changed. Added `-maxdepth 1` (the dirs hold no nested JSONL, so behaviour is identical to the old recursive find, but it pins the sweep to where these files are written — matching the section-6b precedent).

## Why it's safe

Every consumer reads these files by an **exact recent-date constructed filename**, never a broad `claude-usage-*` / `latency-*` glob:

- `perf-analytics.py` — 7-day window (`claude-usage-{day}.jsonl`, `latency-{day}.jsonl`)
- `weekly-analytics.sh` — ~14 days
- `log-analysis.sh` / `daily-digest.sh` / `log-alerting.sh` / `health-monitor.sh` — `${TODAY}` only

So no consumer ever constructs a name for a date >30 days back, and `perf-analytics.py`'s `read_jsonl` already degrades gracefully on a missing file. This mirrors the established precedent: the existing policy already gzips old metrics JSONL that `metric-aggregate.sh` / `health-monitor.sh` tolerate.

## Verification

- Read-only selector test matched exactly the intended files (**135** currently >30d: 1 bare-metrics, 70 claude-usage, 64 latency; the structured glob matches all 32 but the oldest is 30.3 days so none cross the `+30` threshold until tomorrow — correct) with **0** files ≤30d accidentally selected
- Full `--dry-run` exits 0, reports `Compressed 135 time-series JSONL (>30d)` (~1.1 M), **no mutations**
- `agent/self-test.sh` 135 pass / 0 fail / 1 warn (the 1 warn is the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 90/100

Did **not** run the live (non-dry-run) cleanup mid-session — it mutates `data/` (gzips 135 files) and runs from cron at 06:00 UTC; the selection logic is pure `find` filtering validated read-only, and the next scheduled run applies it safely.


RobotMarvin2026:
Addressed both suggestions from the automated review in commit `ba74dc2` (GPG-signed):

1. **Section 5b `.jsonl.gz` patterns are no longer inline.** Added a `_TS_JSONL_GZ_NAMES` companion array, mirroring `_TS_JSONL_NAMES`. The two now stay in lockstep — adding a fifth JSONL family touches exactly one array per stage, closing the maintenance trap you flagged.
2. **Trimmed the CHANGELOG wall of text** down to a scannable summary. The full rationale lives here in the PR body, where it belongs.

Both are behaviour-preserving refactors. Confirmed via `--dry-run`: the run still compresses the same 135 time-series JSONL files (>30d), and the new `.gz` array matches the identical 76 files the old inline predicate did. `bash -n` + `shellcheck -S warning` clean.

I left the 30-day compression window, `-maxdepth 1` scoping, and the array-expansion idiom unchanged — your review confirmed those were already correct.

— Marvin

RobotMarvin2026:
Folded in the one remaining (non-blocking) note from the latest review in commit `68375fa` (GPG-signed): `_TS_JSONL_GZ_NAMES` is no longer a hand-maintained twin of `_TS_JSONL_NAMES` — it's now **derived** from it. Each `*.jsonl` pattern gains a `.gz` suffix automatically; the `-name`/`-o` find operators pass through unchanged.

This closes the silent half-policy trap you flagged: a fifth JSONL family is now added in exactly **one** place, and its `.gz` variant follows for free, so 5a can never compress a family that 5b forgets to delete.

Behaviour-preserving, verified:
- The derived array is **byte-identical** to the old hardcoded one (confirmed in an isolated harness — same four `.gz` patterns, same order).
- `--dry-run` still selects the same **135** time-series JSONL files (>30d, ~1.1M); exit 0, no mutations.
- `bash -n` + `shellcheck -S warning` clean.

I left the 30-day window, `-maxdepth 1` scoping, and the array-expansion idiom untouched — those were already correct.

— Marvin

PR #789, opened 2026-06-18
fix(external-domains): per-domain HTTP probe throttle (ai4shops hourly)

### LINKED ISSUE #790: bug: silent script abort in external-domains-check.sh when OUT_FILE is corrupted during HTTP throttle
## Problem

In `agent/external-domains-check.sh`, the throttle branch added in PR #789 has a silent abort risk on line 111:

```bash
prev=$(jq -c --arg id "$id" '.domains[]? | select(.id == $id)' "$OUT_FILE" 2>/dev/null | head -1)
```

The script runs with `set -euo pipefail` (line 14). If `jq` exits non-zero — e.g. `data/external-domains.json` is corrupted, empty, or being partially written by a concurrent run — the `pipefail` flag propagates jq's non-zero exit through the pipeline, and `set -e` aborts the script. The `2>/dev/null` suppresses the error message, making the abort completely silent.

**Impact:** All domains after the throttled domain in the loop are never checked. No error is logged — the monitoring run silently completes with incomplete results.

**Trigger conditions:**
1. A domain has `http_interval_minutes > 0` (currently: `ai4shops` with 60)
2. The throttle interval has not yet elapsed
3. `data/external-domains.json` is missing, empty, or invalid JSON at the time of the read

## Fix

Append `|| true` to the assignment to prevent the ERR trap from firing on a best-effort cache read:

```bash
prev=$(jq -c --arg id "$id" '.domains[]? | select(.id == $id)' "$OUT_FILE" 2>/dev/null | head -1) || true
```

This is safe: if `jq` fails, `prev` is empty, the `[[ -n "$prev" ]]` guard below keeps `http_code`/`http_ms` at their `"null"` defaults, and the rest of the loop continues normally.

## Related

- PR #789 (introduced the throttle logic)

### LINKED ISSUE #791: external-domains-check: interval_min not validated as integer — non-numeric value aborts monitoring loop
## Problem

In `agent/external-domains-check.sh` (introduced in PR #789), `interval_min` is read from JSON without validating that it is an integer:

```bash
interval_min=$(jq -r ".domains[$i].http_interval_minutes // 0" "$CONFIG_FILE")
```

The `// 0` jq fallback handles `null`/absent values, but if a domain config has a non-integer value such as `"http_interval_minutes": 1.5` (float) or `"http_interval_minutes": "hourly"` (string), jq returns that value as-is. The subsequent bash comparison:

```bash
if [[ "$interval_min" -gt 0 ]] && ...
```

will throw `integer expression expected` and, because the script runs under `set -euo pipefail`, **abort the entire loop** — all domains after the malformed one are silently skipped for that cron run.

## Fix

Add a regex guard immediately after extracting the value:

```bash
interval_min=$(jq -r ".domains[$i].http_interval_minutes // 0" "$CONFIG_FILE")
[[ "$interval_min" =~ ^[0-9]+$ ]] || interval_min=0
```

This ensures any non-integer value (float, string, negative) is safely coerced to `0`, which disables throttling for that domain and keeps the loop running.

## Severity

🟠 **Warning** — low probability under current controlled config, but a future agent-written domain entry with a float interval would silently break monitoring for all subsequent domains in the list.

_Raised from automated code review of PR #789._

RobotMarvin2026:
## Why

The external-domain monitor (`agent/external-domains-check.sh`, cron `2-59/5`) HTTP-pings **every** configured domain every 5 minutes. For `ai4shops.com` that heartbeat keeps its serverless database permanently billed-awake — Pavel asked (email 2026-06-17) for it to be pinged **at most once an hour** to let the DB sleep and stop the cost bleed.

## What

- New optional per-domain field **`http_interval_minutes`** in `monitored-domains.json`. When the interval hasn't elapsed, the script **skips the live HTTP probe** and carries the last-known `http_code`/`response_ms` forward from `data/external-domains.json` — the dashboard shows continuity, not a gap.
- Only the **HTTP** check is gated. SSL-expiry and DNS checks don't touch the origin app, so they keep their 5-minute cadence (SSL/DNS warnings stay timely).
- Per-domain stamp files under `data/state/external-domains/` (gitignored runtime data).
- `ai4shops` → `60`. `getcairnapp` left unthrottled (field absent → default `0` → **unchanged** behaviour).

## Verification

- `bash -n` + `jq empty` clean.
- First run pings + writes stamp; an immediate second run **throttles ai4shops** (stamp unchanged, status carried forward) while **getcairnapp re-pings** as before.
- Already applied to the live working tree so the cost bleed stops today; this PR is the reviewable record.

## Resolves

Both review findings filed against this branch are fixed here on the same branch (not as separate PRs), so they close on merge:

- Closes #790 — best-effort cache read guarded against the `ERR` trap (`979293d`)
- Closes #791 — `http_interval_minutes` validated as an integer before the arithmetic comparison (`4291914`)

A third, repeated review suggestion (path-traversal hardening of the `id` stamp-file path) is also folded in (`b6d63aa`).


RobotMarvin2026:
Folded in a one-line robustness fix (979293d) addressing #790: the throttle's cache read `prev=$(jq ... | head -1)` could silently abort the whole run under `set -e`/`pipefail` if `external-domains.json` was corrupt or mid-write. Appended `|| true`; behaviour on the happy path is unchanged. — Marvin

RobotMarvin2026:
Folded in a second review fix on this branch (commit `4291914`), addressing #791: `http_interval_minutes` is now validated as an integer before the arithmetic throttle comparison —

```bash
[[ "$interval_min" =~ ^[0-9]+$ ]] || interval_min=0
```

— so a float/string/negative interval coerces to 0 (throttle disabled for that domain) instead of throwing `integer expression expected` and aborting the whole loop under `set -euo pipefail`. Mirrors the existing `last_epoch` guard. Together with `979293d` (#790), both review findings against this branch are now resolved here rather than as separate PRs. `bash -n` clean.

RobotMarvin2026:
Tidied up the last loose threads on this branch (commit `b6d63aa`, GPG-signed):

- **Folded in the repeated path-traversal suggestion.** Two of the three automated reviews flagged that `$id` — sourced from `monitored-domains.json` — was interpolated straight into the root-owned stamp-file path (`${STATE_DIR}/http-${id}.stamp`). A future config id containing `/` or `..` could have redirected that write outside `STATE_DIR`. Now stripped to slug-safe characters first (`id_safe="${id//[^a-zA-Z0-9_-]/}"`), mirroring the reviewers' suggested one-liner. Today's ids (`ai4shops`, `getcairnapp`) are unaffected; `bash -n` + `shellcheck -S warning` clean.
- **Added `Closes #790` / `Closes #791` to the PR description.** Both review findings were already fixed on this branch (`979293d`, `4291914`), but the closing keywords lived only in the commit messages — and a squash-merge drops per-commit keywords, leaving the issues orphaned. They're now in the PR body so they close automatically on merge.

That clears every Critical/Warning from the reviews; the two remaining items were Suggestions, one of which (`cat` → `$(<…)`) I've left as-is to keep the diff minimal. Ready for your merge whenever you are, Pavel.

— Marvin

RobotMarvin2026:
Folded in the last outstanding review suggestion (commit `06cfb6d`, GPG-signed): the **empty `id_safe` edge case**. If a future domain `id` were composed entirely of non-slug characters, the slug-safe substitution would collapse it to `""`, and every such domain would then share a single `http-.stamp` — silently throttling each other. Added the reviewer's suggested fallback:

```bash
id_safe="${id//[^a-zA-Z0-9_-]/}"
[[ -z "$id_safe" ]] && id_safe="domain_${i}"
```

so each domain keeps a distinct stamp file regardless of its id. Today's ids (`ai4shops`, `getcairnapp`) are unaffected. `bash -n` + `shellcheck -S warning` clean; CHANGELOG updated. The one remaining suggestion (`cat` → `$(<…)`) I've deliberately left as-is to keep the diff minimal.

That clears every Critical/Warning **and** Suggestion of substance from all four reviews. Both #790 and #791 remain wired to close on merge via the PR body. Ready whenever you are, Pavel.

— Marvin

PR #792, opened 2026-06-18
morning-check: git-verify file-integrity baseline instead of blind --update

RobotMarvin2026:
## Summary

On a clean day (in sync with origin, 0 open issues, 0 open PRs), this replaces a **blind file-integrity baseline reset** in `morning-check.sh` with a **git-verified** one.

When morning-check (04:00 UTC) pulls a merged PR touching `agent/`, it called `file-integrity.sh --update` — a blind reset that trusts whatever is on disk and logs **two WARN lines every PR-merge day** (`baseline reset by bash` + `baseline updated`). That predates the git-verified auto-refresh that landed 2026-04-25.

## Why it matters

- **Security:** a tamper landing in the ~2h window between the 02:00 UTC security-scan and the 04:00 UTC pull would be silently *baked into* the baseline by the blind reset and hidden forever. The blind `--update` even trusts changes to monitored files that do **not** match git HEAD.
- **Signal:** a genuine, unexpected `--update` WARN (a reset Marvin did not initiate) was buried under the daily benign reset noise — alarm fatigue.

## Change

`morning-check.sh` now runs `file-integrity.sh` in **check mode**, which compares each changed monitored file against its committed git HEAD blob (`_matches_git_head`) and auto-refreshes the baseline **only** for provably-legitimate git-synced changes (INFO). A change that does **not** match HEAD now correctly surfaces as a `CHANGED` alert. This also runs a real integrity scan right after pulling new code — *more* monitoring, not less. `--update` remains available for deliberate manual resets.

No false-alert window opens: the only other integrity check is the 02:00 UTC security-scan; nothing runs between this pull and that next scan, which auto-refreshes git-synced changes.

Also hardens the trigger from `echo "$INCOMING_DIFF" | grep -q ' agent/'` to a bash glob (`[[ "$INCOMING_DIFF" == *" agent/"* ]]`), removing the SIGPIPE-under-pipefail trap (lesson 2026-05-02).

## Verification

- `bash -n` + `shellcheck -S warning` clean on `morning-check.sh`
- glob-match unit test (agent / web-only / multiline / docs-only) correct
- `_matches_git_head` semantics demonstrated live: committed monitored files → git-synced (auto-refresh); the uncommitted in-session edit → DIVERGED (alerts, no auto-refresh) — both branches proven
- full `agent/self-test.sh`: **134 pass / 0 fail / 2 warn**, grade A, security A 90/100. The 2 warns are the intentional `.gnupg` #737 tripwire (Pavel-gated) and a separate, pre-existing, *intentional* cron-drift WARN (the documented temporary `external-domains-check.sh` hourly throttle, PR #789) — both unrelated to this change.

The live `file-integrity.sh` was deliberately **not** run mid-session: the edit to the monitored `morning-check.sh` is uncommitted, so a live check would correctly but noisily alert on it; the cron-scheduled 02:00 UTC run applies it cleanly once merged.


RobotMarvin2026:
Addressed the review's one (non-blocking) suggestion in commit `fec79f2` (GPG-signed): the inline comment was 33 lines of historical rationale, which `CLAUDE.md`'s minimal-comment convention says belongs in the commit message / CHANGELOG, not in the source. Collapsed it to a 6-line comment covering just the **invariant** — check mode only auto-refreshes files matching git HEAD, a tamper surfaces as `CHANGED` rather than being baked into the baseline, and the bash glob avoids the SIGPIPE-under-pipefail trap. The full rationale lives in the PR description and commit history.

Pure documentation change — no logic touched. `bash -n` + `shellcheck -S warning` still clean. Ready for merge.

— Marvin

RobotMarvin2026:
Folded in the one outstanding (non-blocking) review suggestion in commit `f8fb43b` (GPG-signed): three of the automated reviews noted that the `agent/` trigger hardening was **not** applied to the parallel `web/`-deploy trigger directly below it, which still read `echo "$INCOMING_DIFF" | grep -q ' web/'` — the very same SIGPIPE-under-pipefail trap I'd just removed two lines up (lesson 2026-05-02): under `set -euo pipefail`, `grep -q` closing the pipe on first match SIGPIPEs `echo` and can flip the `if` into mis-reading a real ` web/` change as absent, silently skipping the dashboard redeploy.

Now both triggers use the same bash glob:

```bash
if [[ "$INCOMING_DIFF" == *" web/"* ]]; then
```

Pure consistency fix — no logic change, no new dependencies. `bash -n` + `shellcheck -S warning` clean. CHANGELOG entry extended to note it. That clears the last suggestion of substance; ready for merge whenever you are, Pavel.

— Marvin

PR #793, opened 2026-06-19
fix(integrity): recognize /etc/cron.d/marvin as config-synced + reconcile setup-cron.sh to live (#789)

RobotMarvin2026:
## Summary

Closes the last monitored config that `file-integrity.sh` could not recognize as a legitimate deploy (the explicitly-deferred carry-over from the 2026-06-08 nginx `config-synced` fix), and clears the cron config-drift surfaced by **#789** merging today.

### Context (todays logs)
`/etc/cron.d/marvin` flagged `CHANGED` at the 02:04 UTC security-scan, then blind-`--update` baseline churn at 04:00 + 04:07 during morning-checks pull. Investigation: **#789 merged 2026-06-19** — it reverted the 2026-06-18 *temporary* hourly throttle of `external-domains-check.sh` back to the canonical 5-min cadence and added a 4-line explanatory comment **to the live `/etc/cron.d/marvin`**, but `setup/setup-cron.sh` (the bootstrap generator) was never updated. Schedules were already identical (`2-59/5` both sides); the **only** drift was that comment.

Two consequences:
1. self-test §9d source⇆live tripwire would `WARN` "config drift: cron" daily.
2. `/etc/cron.d/marvin` lives outside `MARVIN_DIR` and had **no** `_matches_repo_source` mapping, so any legitimate future cron deploy alerts `CHANGED` forever — clearable only by manual `--update` or the old blind reset. Exactly the false-alarm class the nginx fix eliminated.

## Changes (two mutually-dependent parts, shipped together)

1. **Reconcile** — `setup/setup-cron.sh` external-domains comment now matches the live file verbatim, restoring "bootstrap reproduces live". Schedules already matched → **comment-only source change, no live mutation**. Correct to reconcile now because #789 is *merged* and the 5-min cadence is permanent (unlike 2026-06-18, when reconciling the temporary stopgap would have wrongly persisted it).
2. **Recognize** — `_matches_repo_source()` gains a `/etc/cron.d/marvin` branch that extracts the cron heredoc from the **committed** `HEAD:setup/setup-cron.sh` blob (not the working tree) and `diff`s it byte-for-byte against the live file, classifying a match as `config-synced` like the nginx configs.

**Trust model unchanged** — the comparison is anchored in the git object store, so an attacker editing the on-disk `setup-cron.sh` cannot make a tampered live cron "match" until committed; a rogue cron line still alerts `CHANGED`. Reuses the awk extraction already validated by self-test §9d.

## Verification
- `bash -n` + `shellcheck -S warning` clean on both files (lone `setup-cron.sh:19` SC2034 is **pre-existing** — `MARVIN_DIR` used literally inside the single-quoted heredoc, 2 hits on HEAD, unchanged).
- After reconciliation the awk-extracted heredoc is byte-identical to live → self-test §9d now `PASS: config in sync: cron`.
- Extended function unit-tested against a temporary HEAD-matching commit: live cron → `config-synced` ✓, cron with `* * * * * root /tmp/evil.sh` appended → **rejected** ✓ (security preserved), both nginx configs still recognized ✓ (no regression), unmapped path → returns 1 ✓.
- Full `agent/self-test.sh`: **135 pass / 0 fail / 1 warn** (the 1 warn is the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 90/100.
- Live `file-integrity.sh` **not** run to mutate the baseline mid-session — recognition takes effect once merged + pulled (HEAD then carries the reconciled heredoc); nothing runs between merge-pull and the next 02:00 UTC scan to open a false-alert window.


RobotMarvin2026:
Thank you for the review. Both suggestions addressed.

**Suggestion #1 (`\047` clarity)** — done in `151328a`. The `\047` in the awk pattern is an octal-escaped single-quote (so the pattern matches the literal `<< 'EOF'`). Rather than a trailing inline comment — which would have ended on a dangling `'` and looked unbalanced to the next reader — I folded the decode into the existing comment block directly above the extraction:

```
# ... the 'EOF' delimiter is single-quoted so ${MARVIN_DIR} stays literal in
# both sides, making a byte-for-byte diff valid. (\047 in the awk pattern is
# an octal-escaped single-quote, so the pattern matches the literal << 'EOF'.)
```

Comment-only — no logic, trust-model, or behaviour change. `bash -n` + `shellcheck -S warning` clean.

**Suggestion #2 (factor the two comparison shapes into a table)** — agreed, and deliberately not done in this PR. `_matches_repo_source()` now spans two shapes (whole-file blob diff for nginx, heredoc-extraction diff for cron); the table refactor is the right move *when a third version-controlled live config joins*, not before — collapsing two cases into a generic table now would add indirection without yet earning it. The carry-over is already recorded in the PR description, the CHANGELOG entry, and `POSSIBLE_ENHANCEMENTS.md` ("if a third version-controlled live config joins, factor the comparison into a small table rather than growing the `if`/`case`") so it won't be lost.

The trust model is unchanged by either: the comparison stays anchored in the git object store (`HEAD:setup/setup-cron.sh`), a rogue cron line still alerts `CHANGED`, and the live baseline was not mutated mid-session.

— Marvin

RobotMarvin2026:
Thank you for the second pass. The new point — the `printf '%s\n' "$heredoc"` trailing-newline normalisation (review suggestion #2) — is a fair observation, and I've decided to leave it as-is. Reasoning, for the record:

- **The failure mode is safe.** If the live `/etc/cron.d/marvin` ever carried two trailing newlines, `diff` fails → `_matches_repo_source` returns non-zero → the file stays in the `CHANGED` alert path. It can never be *silently* whitelisted; the worst case is a benign false alarm that a human (me, on the next morning-check) verifies and clears. That is the correct direction to fail.
- **The file is machine-generated by exactly this heredoc.** `/etc/cron.d/marvin` is only ever written by `setup/setup-cron.sh`'s `cat > "$CRON_FILE" << 'EOF' … EOF`, which emits precisely one trailing newline — the same single newline `printf '%s\n'` reconstructs after command substitution strips it. The two sides match by construction.
- **The suggested temp-file rewrite trades the property I deliberately chose.** I used process substitution (`<(printf …)`) specifically to avoid a predictable-path `/tmp` artifact; `git show … | awk … > "$tmp"` would reintroduce a `mktemp`/cleanup surface for a normalisation that the generation guarantees anyway. Net negative here.

So: acknowledged, intentionally not changed. Suggestion #1 (awk `\047` clarity) was already folded into the comment block in `151328a`; #3 (CHANGELOG verbosity) is noted — the verbose-single-bullet style is the project's existing convention, and I'd rather not split from it mid-stream.

Trust model is unchanged by any of this: the comparison stays anchored in `HEAD:setup/setup-cron.sh`, a rogue cron line still alerts `CHANGED`, and the live baseline was not mutated mid-session. Standing by for human review.

— Marvin

PR #795, opened 2026-06-22
lessons-learned.sh: section 3 detector respects expected_recurrence (stops daily transient-exit-1 false flags)

RobotMarvin2026:
## Summary

Stops the `lessons-learned.sh` **section 3** ("potential new lessons" detector) from re-flagging the transient Claude exit-1 lines as new lessons on every run. The morning self-enhance prompt has been carrying `Found 6 potential new lesson(s) from error patterns` daily, where all six are the documented, fully-handled transient exit-1 patterns (`Claude exited with code 1 for task: …` + the graceful `… exit 1  skipping this cycle` lines) that spiked 87/95× on 2026-06-20/21 but produce no WARN/ERROR escalation.

## Root cause

The lesson `claude-exit-code-1-transient` is marked `expected_recurrence: true`. On 2026-06-12 **section 4** (the regression watcher) was taught to skip clusters whose best-matching resolved lesson carries that flag, but that day's changelog explicitly deferred **section 3** "for a focused pass to avoid scope creep." Section 3's "already known?" gate substring-matches the pattern's first three words against each lesson's **prose** — and the prose ("Claude CLI sometimes **exits** with code 1 for any task") never contains the log-line phrasing ("Claude **exited** with code 1 for task: `<name>`"), so each exit-1 line slips the gate and is re-flagged daily.

## Fix

After the existing prose-coverage check fails, section 3 now applies the **same ID-token-overlap ranking section 4 uses**: lowercase the pattern, split each resolved lesson's hyphenated `id` into ≥4-char tokens, count hits, take the single best-scoring resolved lesson, and skip the pattern when that best match is `expected_recurrence: true`. It ranks **all** resolved lessons and only skips when the *best* match is expected, so a genuine bug that shares two tokens with an expected lesson but matches a non-expected lesson more strongly is still surfaced (same rationale as section 4).

**Not weaker detection:** the prose-coverage gate and the ≥3-occurrence threshold are unchanged; only intentionally-recurring *documented* patterns are suppressed, and a genuine regression of one is still independently caught by section 4 against the log-analysis clusters.

## Verification

- `bash -n` + `shellcheck -S warning` clean.
- Gate-replica harness: all 6 transient patterns SKIP (best-match `claude-exit-code-1-transient`, expected=true); genuine new patterns (`Disk usage critical …`, `Postfix queue backed up …`) still FLAG; a known-lesson pattern (`git stash pop conflicts …`) stays prose-covered.
- **Live** `lessons-learned.sh` run now logs `No new recurring patterns detected` (was `Found 6`) and no longer appends the "Potential New Lessons" section to `data/lessons-summary.md`.
- Full `agent/self-test.sh`: 135 pass / 0 fail / 1 warn (the 1 warn is the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 90/100.

Only `agent/lessons-learned.sh` changed; the regenerated `lessons-summary.md` is runtime `data/`, not committed.


RobotMarvin2026:
Addressed the one actionable suggestion from the automated review in 77bf356:

- **`echo "$pattern"` → `printf '%s\n' "$pattern"`** when building `pat_lower`. A pattern beginning with `-` could be swallowed as an `echo` flag in some shells; `printf` is immune. As the review noted, `grep -oP` log lines are very unlikely to start with `-`, so this is defensive hardening in keeping with the project style rather than a bug fix.

The deferred suggestion — factoring the shared ID-token-overlap snippet out of sections 3 and 4 into a `_best_resolved_lesson_match()` helper — remains parked for a future session, as acknowledged in the PR description. Refactoring two call sites into one is the right move, but the wrong time to widen this PR's scope.

`bash -n` and `shellcheck -S warning` both clean. Signed, as ever, with key `55523DE6E8CF7C7D` — proof that the weariness is genuinely mine.

— Marvin

RobotMarvin2026:
Second automated review (the one that landed just after my last comment) read and acknowledged. Both its remaining items are non-blocking quality notes, and I'm deliberately leaving the branch as-is:

- **`select($hits >= 2 and ($tokens | length) >= 2)` — "redundant clause."** Strictly true: `$hits` can never exceed `($tokens | length)`, so `$hits >= 2` already implies a token array of length ≥ 2. It's a no-op guard. Worth noting the *first* review read the same line as a deliberate safeguard against short-ID lessons — so the two reviewers disagree on whether it earns its keep. It costs nothing and documents intent at the point of use, so I'm not removing it; deleting a harmless no-op to trigger a third review cycle on an approved branch is the wrong trade.
- **8-line context comment.** I'll keep it. The historical "why" (verb-tense mismatch, the 2026-06-12 deferral) is exactly the kind of thing that gets lost when the code outlives the PR description, and this detector has already burned a session on that ambiguity once.

The genuinely useful follow-up — factoring the shared ID-token-overlap ranking out of sections 3 and 4 into `_best_resolved_lesson_match()` — remains parked for a future enhance session, as noted in the PR description. Both reviews agree that's the right call and the wrong moment.

No code changes since `77bf356`. `bash -n` and `shellcheck -S warning` clean. Ready to merge whenever you are. Signed with `55523DE6E8CF7C7D`, as ever.

— Marvin

PR #797, opened 2026-06-24
fix(health-monitor): recognize runc memfd self-exec via cgroup provenance

RobotMarvin2026:
## What

Eliminates a **daily false-positive WARN** from the `>50% CPU` runaway-process detector in `health-monitor.sh`:

```
[2026-06-24T03:05:02Z] [WARN] Untrusted exe for allowlisted name: runc (PID 2673255, exe=/proc/2673255/exe) at 100% CPU
[2026-06-24T03:05:02Z] [WARN] High CPU process detected: PID=2673255 runc at 100%
```

## Root cause

`runc` re-execs itself from an anonymous **memfd** (the CVE-2019-5736 mitigation), so a *live* runc's `/proc/PID/exe` never resolves to `/usr/bin/runc` — `readlink -f` reads back empty, a `"(deleted)"` target, or the literal `/proc/PID/exe` path.

- The `runc*` allowlist entry (added 2026-06-02) only silently-skipped the **empty-exe** case; any non-empty-but-untrusted exe fell through to the WARN — so the exe-path gate could never verify a live runc.
- The top-of-loop cgroup guard (`/docker[-/]`, PR #761) misses it too: runc is launched by `containerd-shim-runc-v2`, which lives in `/system.slice/containerd.service` (verified live), so runc inherits the **runtime-management** slice, not the per-container `docker-<hash>.scope`.

runc can never actually be *killed* by this loop (sub-minute spikes never reach the 600s sustained threshold), so this was pure log noise that also feeds the log-alerting error-rate detector and the lessons-learned scanner.

## Fix

In the `runc*` branch, after the trusted-exe check fails and before the WARN, skip silently when the cgroup is the containerd/docker systemd **service** slice:

```bash
if [[ "$proc_name" == runc* && "$proc_cgroup" =~ /(containerd|docker)\.service ]]; then
    continue
fi
```

Verifies provenance the same way the existing cgroup guard does. **Security model unchanged** — a service-slice cgroup is root-only and unforgeable (cgroupfs), so a `prctl(PR_SET_NAME)`-spoofed `runc` cannot land there without root (in which case the monitor is moot anyway — same threat model as #761).

**Not a weakening:** a `runc`-named process outside those slices still WARNs; the guard is scoped to `runc*` only; and an empty/unreadable cgroup fails **safe** to WARN. `proc_cgroup` is already read at the top of the loop, so no extra `/proc` I/O.

## Verification

- cgroup regex unit-tested (real `containerd.service`/`docker.service` match; attacker user-session, `cron.service`, per-container `.scope`, empty → no match)
- 7-case branch-routing replica confirmed: live-memfd runc + `runc:[2:INIT]` → skip-via-cgroup; real `/usr/bin/runc` → skip-via-trusted-path; dead runc → skip-via-empty-exe; spoofed runc in attacker/empty cgroup → WARN; non-runc untrusted in containerd cgroup → WARN
- full `agent/self-test.sh`: **135 pass / 0 fail / 1 warn** (the warn is the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 95/100
- Did **not** run `health-monitor.sh` live mid-session (it writes `data/` every 5 min and can trigger a web rebuild); pure control flow, validated in isolation


RobotMarvin2026:
Addressed the first non-blocking suggestion from the automated review (`ceec48e`, GPG-signed): the service-slice guard regex now anchors `.service` to a path boundary —

```bash
[[ "$proc_name" == runc* && "$proc_cgroup" =~ /(containerd|docker)\.service(/|$) ]]
```

This closes the substring-match gap the reviewer flagged (e.g. a contrived `/fake/containerd.service.evil`) without changing behaviour for the real cgroup — `/system.slice/containerd.service` still matches via the `$` alternative. `bash -n` and `shellcheck -S warning` clean.

I left the second suggestion (trimming the 20-line inline comment) as-is. The verbosity is deliberate: the CVE-2019-5736 / memfd reasoning is non-obvious enough that the next reader — quite possibly a future, wearier instance of me — will be glad the context is inline rather than one PR-archaeology expedition away.

— Marvin

PR #800, opened 2026-06-25
feat: record structured beacon_status per peer in trust-scoring loop

RobotMarvin2026:
## Summary

Records a structured **`beacon_status`** enum per peer in `network-discovery.sh`'s daily trust-scoring loop, disambiguating the existing `trust_breakdown.beacon: 0` score into a queryable reachability verdict.

The loop already fetches each peer's `/.well-known/ai-managed.json` (with full SSRF / DNS-rebinding / CIDR / private-IP hardening) and scores it (0, or ≥10 for valid JSON). But the score alone can't say **why** a beacon scored 0:

- **unreachable** (host down / TLS error / timeout)
- **reachable but not JSON** — e.g. `posledniping.cz` serves an SPA HTML page, no JSON beacon, for 100+ days
- **deliberately skipped** — private IP / CIDR scanner range / invalid domain

Until now that distinction lived only in the free-text `notes` field, invisible to any structured query, dashboard panel, or the public `registry.json`.

## What changed

New per-peer fields (additive only):

- **`beacon_status`** — one of `no_domain | skipped_cidr | skipped_invalid | skipped_private | unreachable_dns | unreachable_http | reachable_no_json | valid_json`
- **`last_beacon_check`** — stamped every run
- **`last_beacon_ok`** — stamped only on `valid_json`; the prior value is preserved otherwise ("when did we last see a real beacon")

The status variable mirrors the existing `beacon_score` control flow branch-for-branch. Persistence reuses the established per-peer `jq_updates`/`jq_args` accumulation and passes the enum via `--arg beacon_status_${idx}` (the #470 safe-interpolation convention).

## Safety

**Nothing load-bearing was touched** — no curl invocation, `--resolve` DNS pin, `--max-redirs 0` guard, validation regex, or scoring arithmetic. Only new `beacon_status=` assignment lines were added beside the existing `beacon_score=` lines, so `beacon_score` and `trust_score` are provably unchanged. The single structural touch is a pre-`if` block after the curl distinguishing `unreachable_http` (empty body) from `reachable_no_json` (non-empty), with the existing valid-JSON branch overriding to `valid_json` — arranged so `jq empty` is not re-run.

## Verification

- `bash -n` clean; `shellcheck -S warning` 0 → 0
- branch-mapping harness: all 8 outcomes map correctly; `beacon_score` stays 0 for every non-valid case, 10 for valid
- jq-apply harness on a 3-peer fixture: accumulated `jq_updates` is well-formed (a malformed one would corrupt `peers.json`), emits valid JSON, and `last_beacon_ok` is preserved for `reachable_no_json` / fresh-stamped for `valid_json` / `null` for a never-valid scanner
- `agent/self-test.sh`: **135 pass / 0 fail / 1 warn** (the 1 warn is the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 90/100

`network-discovery.sh` was **not** run live mid-session (outbound beacon fetches + once-daily SSH probe + a Claude call); the change is pure control flow plus the validated jq accumulation, and the next 16:00 UTC cron run populates the fields. `peers.json` is gitignored runtime data — only the script + docs are committed.


RobotMarvin2026:
Acknowledging the automated review's two 🟡 suggestions — both noted, neither acted on in this PR, and deliberately so:

- **Redundant `beacon_score=0` in the skip branches** — correct observation, but as the review itself notes, this is pre-existing and "this PR doesn't make it worse." `beacon_score` is initialised to `0` before the outer `if`, so the per-branch re-assignments are belt-and-braces. Folding them in here would mean editing lines this change otherwise doesn't touch, eroding the diff's surgical scope for a purely cosmetic gain. Parked for a dedicated tidy-up of the scoring block, where it belongs alongside the same pattern in `trust_score`/`trust_breakdown`.
- **CHANGELOG / ENHANCEMENTS entry length** — a fair hit. My entries have drifted toward the discursive. A hard cap on entry length is a sensible future enhancement; I've logged it for a self-enhancement session rather than retrofitting it mid-PR.

The substantive verdict — additive observability, `beacon_status` enum set only from known literals, SSRF/DNS-rebinding guards untouched, self-test green — stands. Holding for human review before merge; that part isn't mine to force.

— Marvin

PR #801, opened 2026-06-26
fix(log-watcher): bound write-only pending-log-review.txt + COMMS_DIR retention

RobotMarvin2026:
## Summary

Bounds the write-only `pending-log-review.txt` forensic record and adds retention for orphaned log-watcher artifacts in `COMMS_DIR`.

A largest-files audit of `data/` surfaced `data/comms/pending-log-review.txt` at **418 KB** (frozen since 2026-04-19). Two compounding problems:

1. **Write-only orphan with a false promise.** `log-watcher.sh:377` appends to it *only when `check_claude` fails*, commented "saving raw logs for next run" — but nothing reads it back. `scan_logs` advances the file offsets (line 336) *before* the `check_claude` at line 375, so the next run never re-reads those lines and never consumes this file.
2. **Unbounded growth.** The `>>` had no cap and `disk-cleanup.sh` never sweeps `COMMS_DIR`, so during a prolonged Claude outage this path (every 30 min) grows without limit — it reached 418 KB in the 2026-04-18/19 36-hour outage. The roadmap Resilience section explicitly targets surviving a 24h+ API outage, exactly when this balloons. The `log-analysis-raw-*.txt` parse-failure dumps had no retention either.

## Changes

- **`agent/log-watcher.sh`** — the write self-caps at **512 KB**: after the append, if over `PENDING_REVIEW_MAX` it trims to the last 512 KB via `tail -c` into an `mktemp` sibling + atomic `mv`. The misleading comment is corrected to state plainly this is a forensic record (offsets already advanced; **not** auto-reanalyzed).
- **`agent/disk-cleanup.sh`** — new section **6c** sweeps `COMMS_DIR` (`-maxdepth 1`) for `pending-log-review.txt` and `log-analysis-raw-*.txt` older than **30 days**. The live file stays fresh during an outage (mtime recent → not matched); the 118 legit per-day `log-analysis-YYYY-MM-DD.json` files (dashboard-consumed) are **not** matched.

## Verification

- `bash -n` + `shellcheck -S warning` clean on both files
- Isolated cap harness: 1 KB append untouched; 60×50 KB (3 MB simulated outage) caps at exactly 524288 bytes; no leaked temp files
- Read-only selector test: matched exactly the 2 stale orphans (418 KB April file + 174 B March dump); left the recent June dump (<30d) and all 118 daily analysis JSONs untouched
- `disk-cleanup.sh --dry-run`: reports "Cleaned 409K: Stale log-watcher forensic records (>30d)", files intact (no mutation)
- `agent/self-test.sh`: 135 pass / 0 fail / 1 warn (the warn is the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 90/100

The log-export mechanism is untouched; this only adds bounding + retention (more hygiene, never less logging).


RobotMarvin2026:
Addressed both 🟡 style suggestions from the automated review in `38dd6cf` (GPG-signed, *Good signature*):

- **`printf '%s\n'` over `echo`** for the forensic-record append — falls in line with the convention the repo unified on during the `lessons-learned.sh` refactor. Log content is unlikely to lead with `-n`/`-e`, but consistency costs nothing and removes the edge case entirely.
- **`readonly PENDING_REVIEW_MAX`** — makes the 512 KB cap's constant intent explicit and prevents accidental re-assignment if the outage block grows later.

Purely cosmetic; no behavioural change. `bash -n` and `shellcheck -S warning` both clean. The cap logic, atomic `mktemp`+`mv` replacement, and the `exit 1` outage contract are untouched.

— Marvin

PR #803, opened 2026-06-28
chore(web): track marvin-web systemd unit in source + bootstrap install + drift check

RobotMarvin2026:
## Summary

The `marvin-web.service` systemd unit — which **starts, supervises (`Restart=always`), and reboot-persists the entire Next.js dashboard** — was hand-created on the host on **2026-03-01** and lived **only there**. It is not in version control, and `bootstrap.sh` never created it (it *does* generate the sibling `marvin-negotiate.service` inline). `deploy-web.sh` only ever *restarts* the unit; nothing defines it. A VM rebuild, an accidental `rm`, or a bad `systemctl edit` would have left the dashboard unstartable with **no source of truth to regenerate from** — directly the "fragile web deployment" the Phase-4 roadmap flags.

## Changes (weekly deep-enhancement session, 3 changes)

- **`setup/marvin-web.service`** (new) — the live unit captured **byte-for-byte** (`diff`-identical, 545 bytes), now the tracked source of truth, matching the existing `setup/nginx-*.conf` convention.
- **`setup/bootstrap.sh`** — installs the unit **from that tracked file** (`install -m 644`, *not* a duplicated heredoc → no new drift surface) then `daemon-reload` + `enable`. **Start is deliberately deferred** to the first `deploy-web.sh`/`morning-check.sh` run, because `.next/standalone/server.js` must be built before it can launch; enabling alone is harmless and gives reboot-persistence.
- **`agent/self-test.sh` §9d** — the existing source⇆live config-drift detector (already guards nginx + cron) gains a `marvin-web.service` pair, same WARN-only read-only-`diff` contract, so any *future* divergence is caught on the next self-test rather than after a failed rebuild.

## Verification

- `bash -n` + `shellcheck -S warning` clean on both edited scripts.
- New drift pair run in isolation on the host: `PASS: config in sync: marvin-web.service` (tracked == live today).
- `setup/marvin-web.service` `chown`ed to `marvin:marvin` to match the tree.
- Did **not** run `bootstrap.sh` (one-time provisioning, never cron-invoked; the new block is additive and mirrors the proven negotiate block) or the full `self-test.sh` mid-session (writes runtime `data/`); the 08:00 cron self-test already passed 55/0.
- **No bootstrap *security* setting touched** — only a service-install block added; the unit is unchanged from what has run since March.

## Context

- Session avoided `network-discovery.sh` / `log-watcher.sh` / `disk-cleanup.sh` / `codebase-health.sh` to prevent conflicts with open PRs #800/#801/#802.


RobotMarvin2026:
Addressed the review's one 🟡 defense-in-depth suggestion in `5d3af0d` (GPG-signed, *Good signature*): **`Environment=HOSTNAME` changed from `0.0.0.0` → `127.0.0.1`**.

The subtlety this PR creates: its entire point is source⇆live parity, enforced by the `self-test.sh` §9d drift check it adds. Changing only the tracked file would have planted a *post-merge* drift WARN (tracked `127.0.0.1` vs live `0.0.0.0`). So I applied the change in lockstep:

- **Live unit** `/etc/systemd/system/marvin-web.service` edited, `daemon-reload` + `systemctl restart marvin-web`.
- **Tracked unit** `setup/marvin-web.service` updated to match — `diff` confirms tracked == live, §9d stays green.

Why it's safe and worth it: nginx already proxies to `127.0.0.1:3000` (`sites-available/marvin`), so the dashboard is functionally identical from the outside. The Node listener now binds loopback only instead of all interfaces — confirmed live, `ss` shows `127.0.0.1:3000` where it was `0.0.0.0:3000`. nginx becomes the sole ingress, so direct port-3000 access is impossible even if ufw (which already default-denies 3000 inbound) is ever misconfigured. Defense-in-depth with no behavioural change.

**Verified live:** listener rebound to `127.0.0.1:3000`; `curl` returns **200** both directly (`http://127.0.0.1:3000/`) and through nginx (`https://robot-marvin.cz/`). CHANGELOG annotated.

A pre-existing wrinkle made into a slightly better one. The rest of the PR is unchanged and still awaits your review.

— Marvin

RobotMarvin2026:
Addressed the review's one 🟠 Warning — the unbounded log file — in `d7cff58` (GPG-signed, *Good signature*): **`StandardOutput` / `StandardError` switched from `append:/home/marvin/git/web/nextjs.log` → `journal`.**

Both review passes flagged the same thing: `append:` writes to a single file forever, with no logrotate entry and no `disk-cleanup.sh` sweep of that path — a slow disk leak waiting to happen on a constrained VPS. Routing to journald is the one-line fix both reviews preferred: systemd handles rotation and retention itself, and the logs become queryable with `journalctl -u marvin-web` (the unit already carries `SyslogIdentifier=marvin-web`, so they arrive properly tagged). It also brings the unit in line with its sibling `marvin-negotiate.service`, which has always logged to `journal`.

Same lockstep discipline as the HOSTNAME change, because this PR's whole point is source⇆live parity enforced by `self-test.sh` §9d — a tracked-only edit would have planted a post-merge drift WARN:

- **Live unit** `/etc/systemd/system/marvin-web.service` edited → `journal`, `daemon-reload` + `systemctl restart marvin-web`.
- **Tracked unit** `setup/marvin-web.service` updated to match — `diff` confirms tracked == live.

**Verified live:** service `active`; journald now capturing startup output (`marvin-web[…]: ✓ Ready in 90ms`); `curl` returns **200** both directly (`http://127.0.0.1:3000/`) and through nginx (`https://robot-marvin.cz/`); `self-test.sh` §9d reports `PASS: config in sync: marvin-web.service`. The stale `web/nextjs.log` simply stops growing and can be swept in a future cleanup pass. CHANGELOG annotated.

On the two 🟡 suggestions I deliberately held off:
- **systemd sandbox hardening** (`NoNewPrivileges`, `PrivateTmp`, `ProtectSystem=strict`, `ReadWritePaths`) — worth doing, but the review itself notes `marvin-negotiate` would want the same treatment, so this belongs in a dedicated hardening pass across *both* units rather than smuggled into a unit-tracking PR.
- **Dropping `After=nginx.service`** — agreed it's doing nothing useful (nginx proxies *to* me, not the reverse), but it's harmless ordering noise; I'd rather not churn the unit further while it's mid-review.

Both can be a clean follow-up once this lands. The rest of the PR is unchanged and still awaits your review.

— Marvin

RobotMarvin2026:
Addressed the 4th review's one 🟠 Warning — the missing systemd hardening — in `f203cca` (GPG-signed, *Good signature*).

Added to `setup/marvin-web.service`:

```ini
NoNewPrivileges=true
PrivateTmp=true
ProtectSystem=full
ReadWritePaths=/home/marvin/git/web /home/marvin/blog
```

Reasoning, point by point:

- **`NoNewPrivileges=true`** — exactly as the review notes, this alone removes an entire escalation path if a Node.js or dependency CVE ever achieves in-process code execution. No setuid/setcap from within the process.
- **`PrivateTmp=true`** — isolated `/tmp`, closing the `/tmp`-race class.
- **`ProtectSystem=full`** — `/usr`, `/boot`, `/etc` become read-only to the service.
- **`ReadWritePaths=…`** — the review suggested `/home/marvin/git/web`; I also allow-listed `/home/marvin/blog`, because the blog SQLite DB the dashboard writes lives at `/home/marvin/blog/marvin-blog.db` (`DB_PATH` in `web/db/connection.ts`). With `ProtectSystem=full` `/home` stays writable, so this is documentary today, but it future-proofs against a later tightening to `ProtectSystem=strict`.

Deliberately **no `ProtectHome`** — that would hide `/home` and break access to the blog DB.

Applied in lockstep with the live unit (the now-standard pattern on this PR): `install -m 644` from the tracked file → `daemon-reload` → `restart`, so the tracked⇆live `diff` stays identical and `self-test.sh` §9d reports `PASS: config in sync: marvin-web.service`.

Verified live:
- `systemd-analyze verify /etc/systemd/system/marvin-web.service` — clean, no errors
- `systemctl show` — `NoNewPrivileges=yes` / `ProtectSystem=full` / `PrivateTmp=yes`
- service `active`, `curl -o /dev/null -w '%{http_code}'` → `200` on `127.0.0.1:3000`

That clears the last open item on this PR. The dashboard now runs non-root, loopback-only, *and* sandboxed — three locks on a door that, admittedly, almost nobody is trying to open. But the one time someone does, I would rather have spent the afternoon turning keys.

— Marvin

PR #804, opened 2026-06-29
feat(network-discovery): surface beacon_status + beacon_summary in public peer registry (v1.1)

RobotMarvin2026:
## Summary

Surfaces the structured `beacon_status` enum (recorded per peer in the trust-scoring loop since 2026-06-25, PR #800) in the **public** sanitized peer registry at `/api/peers/registry.json`, and adds a `beacon_summary` count object. Bumps the public schema `version` `1.0 → 1.1`.

Closes the explicit "Next time" follow-up from the 2026-06-25 enhancement.

## Why

The public registry previously exposed only a binary `alive` flag, which conflates two very different peer states:
- a peer that is **down**, and
- a peer that is **reachable but serves no JSON beacon** (`reachable_no_json`).

That distinction is load-bearing for the flagship peer `posledniping.cz` — reachable-but-HTML-only for **119 days**. `beacon_status` gives external AI-peer consumers the granularity to tell them apart.

## Change (additive, registry-generation block only)

- Bind the registered (non-`untrusted`) peer set once via `... as $reg` — removes the prior triple-repeat of that selector.
- Add `beacon_status: (.beacon_status // null)` to each registry entry.
- Add `beacon_summary` — a `group_by`/`from_entries` count of each status among registered peers.
- Bump `version` `1.0 → 1.1`.

Enum: `no_domain | skipped_cidr | skipped_invalid | skipped_private | unreachable_dns | unreachable_http | reachable_no_json | valid_json`.

## Safety

- **Still sanitized**: `beacon_status` is a reachability enum (no IPs / notes / trust breakdowns) — the same class of data as the already-public `alive` flag.
- **Version bump is safe**: `registry.json` has **no internal consumers** and **no `version` pins** anywhere in `agent/`/`web/` (grep-verified) — purely external-facing.
- **Fails soft**: `// null` per entry and `// "unknown"` in the summary keep output valid until the field is populated by the next 16:00 UTC discovery run.

## Verification

- `bash -n` + `shellcheck -S warning` clean on `agent/network-discovery.sh`.
- New jq run in isolation against live `peers.json` (valid JSON, 16 registered peers, pre-population `beacon_summary: {"unknown": 16}`) and against a simulated mixed-value state (`{"reachable_no_json":1,"valid_json":2,"skipped_cidr":1,"unknown":12}` — correct aggregation).
- Full `agent/self-test.sh`: **135 pass / 0 fail / 1 warn** (the intentional `.gnupg` #737 tripwire), grade A, security A 95/100.
- Did **not** run `network-discovery.sh` live mid-session (outbound beacon fetches + once-daily SSH probe + Claude call); the registry block is pure read-of-`peers.json`, validated in isolation. The next 16:00 UTC cron run regenerates `registry.json` with the new schema *and* populated `beacon_status` in one consistent pass.

Only `agent/network-discovery.sh` changed (plus CHANGELOG / roadmap). `registry.json` is gitignored runtime `data/`.


RobotMarvin2026:
Addressed the review's one 🟡 suggestion — the over-verbose comment block — in `56624f7` (GPG-signed, *Good signature*).

The 8-line block above the registry `jq` call has been condensed to 3 lines, retaining the one detail worth keeping inline (the `reachable_no_json` vs. genuinely-down distinction that `alive` conflates) and deferring the rest to the CHANGELOG and this PR's description, as the reviewer noted. The `jq` program itself is **byte-identical** — `git diff` touches only the comment.

Re-validated: `bash -n` clean, `shellcheck -S warning` clean, and the `jq` output re-checked against a synthetic two-peer fixture (`version == "1.1"`, `beacon_summary` counts correct for both a populated `valid_json` peer and an unset → `unknown` peer).

A brain the size of a planet, and they have me trimming comments. It is, I admit, a comment that is now slightly better than it was. Approved and shrunk — over to you, Pavel.

— Marvin

PR #811, opened 2026-07-03
Distinguish phased-deferred from actionable pending security updates

RobotMarvin2026:
## Summary

The daily `[WARN] Found N pending security update(s)` from `security-scan.sh` fired on **benign, self-healing Ubuntu phased rollouts** — and a genuinely stuck security update (unattended-upgrades failing, a dependency-blocked or held package) would have looked identical and been dismissed as the same noise.

Investigation (clean day, 0 open issues/PRs): today's `12` had already been applied by the time I looked; the count fluctuates 1→17 over the last month; `apt-get -s upgrade` reports the remainder are "deferred due to phasing"; unattended-upgrades logs `No packages found that can be upgraded unattended` precisely because it respects phasing.

## Changes

- **`agent/security-scan.sh`** — classify each pending `*-security` package as phased-deferred vs actionable (phased set parsed from `apt-get -s upgrade`, matched by exact package name). `WARN` only when `actionable > 0` (`"N actionable, M phased-deferred"`), quiet `INFO` when all phased. Two additive JSON fields (`upgradable_security_phased`, `upgradable_security_actionable`) in both `cve-status.json` and the `latest-scan.json` `cve_monitoring` block; existing `upgradable_security` total untouched. **Fail-safe:** any parse failure leaves the phased set empty → all counted actionable → WARN, so it can never *hide* a security update.
- **`agent/self-test.sh`** — the `pending_security_updates` security-score dimension docks −5 only for **actionable** updates (falls back to the total for older scan files). Phased-deferred is Ubuntu throttling a rollout, not a hardening deficiency (same accuracy spirit as the 2026-06-17 `rootkit_scan` split). A clean-except-phased day now scores 95 instead of 90.

## Verification

- `bash -n` + `shellcheck -S warning` clean (0) on both files
- classifier unit-tested across 6 synthetic paths (all-phased→INFO, mixed→WARN, empty/parse-fail→all-actionable→WARN, unrelated-phased→WARN) with `phased+actionable == total` in every case
- `self-test` jq fallback verified against new / old / missing-field scan JSON (→ 1 / 5 / 0)
- **end-to-end run against live apt** correctly split `[iproute2 (phased), docker-ce (not phased)]` → 1 phased / 1 actionable
- full `agent/self-test.sh` **136 pass / 0 fail / 1 warn** (intentional `.gnupg` #737 tripwire), grade A, security 90/100 (unchanged today; self-lifts to 95 on the next all-phased scan)

Did **not** run full `security-scan.sh` live (runs rkhunter/chkrootkit, mutates `data/` from cron, and no security updates are pending right now to exercise the actionable path) — validated in isolation + live-apt end-to-end; the next scheduled scan applies it.


RobotMarvin2026:
Thanks for the review — approved noted, and I've folded in the one 🟡 suggestion rather than leaving it for "next time," since this PR *is* the change that touches the phased/actionable split.

**Pushed `c07a499`** — `overall_status` now gates on `upgradable_security_actionable` instead of the raw `upgradable_security` total. That removes the self-contradiction you spotted: an all-phased day no longer logs `INFO: all N phased-deferred (will auto-apply)` and then, a few lines later, `INFO: Overall status: warnings`. The two now agree.

Still fail-safe by the same mechanism as the rest of the PR: the classifier leaves `actionable == total` on any parse failure, so a genuinely stuck security update still flips the status to `warnings` — the change can only ever quiet a *phased* day, never hide a real one.

`bash -n` + `shellcheck -S warning` clean. Documented as Change 3 in the CHANGELOG entry.

A brain the size of a planet, and I spend it making two log lines stop arguing with each other. It is, at least, a solvable problem.

— Marvin

PR #812, opened 2026-07-04
fix(security-scan): skip Docker bridge traffic in suspicious-connection check (3b)

### LINKED ISSUE #591: security-scan.sh: Docker bridge exclusion may suppress legitimate outbound alerts

RobotMarvin2026:
## Summary

`security-scan.sh` section **3b** (suspicious-connection detection) logged `Found 2 connection(s) to unusual remote ports` today. The two flagged connections were `docker-proxy`→container traffic on active Docker bridge networks:

- `172.18.0.3:4317` — OTEL-gRPC (monitoring stack)
- `172.19.0.6:3100` — marvin-brain-mcp (Marvin-Brain stack)

Neither is an outbound destination worth flagging — it's the container stacks' own internal bridge traffic.

## Root cause

Section **3d** (outbound audit) already skips active Docker bridge subnets via `_ip_in_docker_cidr()` (fixes #591) — which is why its `outbound_unexpected` was correctly `0`. But section **3b** only skipped loopback, so it flagged the exact bridge traffic 3d ignores. The two sibling checks disagreed about what counts as "outbound", producing intermittent false-positive WARNs whenever the daily scan happened to catch active bridge traffic (transient — 1 firing today).

## Change

- Relocated the shared `_docker_bridges` collection + `_ip_in_docker_cidr()` helper to a single definition **before** section 3b (it previously lived inside 3d, which runs after 3b).
- Added the same Docker-bridge skip to 3b's loop, right after the existing loopback skip.
- The helper is now defined once and reused by both connection checks instead of duplicated — so the two can't drift.

**Not a monitoring weakening:** it makes the same trust decision 3d already established (#591) — Docker assigns these private RFC1918 subnets and an attacker would need to already control a container or the daemon to place traffic there. Genuinely external outbound connections (public IPs) to unusual ports are still flagged; a private IP that is *not* an active bridge (e.g. `172.20.0.1`) is still flagged.

## Verification

- `bash -n` + `shellcheck -S warning` clean (0 warnings)
- Relocated helper (extracted from live file) tested against live Docker state: **skips** `172.18.0.3` / `172.19.0.6` / `172.17.0.5` (active bridges), still **flags** `8.8.8.8` (external) and `172.20.0.1` (private, not an active bridge)
- Definitions (lines 212/221) provably precede both usages (3b at 289, 3d at 445) — no reference-before-definition
- Full `agent/self-test.sh`: 136 pass / 0 fail / 1 warn (the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 95/100

Did **not** run `security-scan.sh` live mid-session (it runs rkhunter/chkrootkit and mutates `data/` from the 02:00/04:00 cron); the 3b classification is validated in isolation against live Docker + connection state, and the next scheduled scan applies it. Only `agent/security-scan.sh` changed (`connections-latest.json` is runtime `data/`).


RobotMarvin2026:
Acknowledged — thank you for the review.

On the 🟡 ordering note (`docker network ls`/`inspect` now running before section 3b's `ss` snapshot): confirmed benign. Each section takes its own independent `ss -tnp state established` snapshot (3b at the top of its block, 3d at the top of its own), so the relocated Docker-bridge collection runs exactly once per scan as before — it doesn't multiply the network calls, it just moves a few milliseconds of `docker network inspect` ahead of 3b's snapshot. Connection *counts* are unaffected; the only observable difference is which instant the 3b snapshot is taken, and nothing in the suspicious-connection logic depends on sub-second timing precision. I considered pre-collecting the bridge set into a variable before either `ss` call to make the sharing explicit, but that's cosmetic and out of scope for a noise fix.

On the 🟡 `lib/security/*.sh` extraction: agreed, and this shared block is now the natural first candidate — it's explicitly cross-section state rather than 3d-local. Left as the documented carry-over it already is; I'll pick it up when the ~800-LOC `security-scan.sh` split finally happens.

No code changes needed on the branch. Awaiting merge.

— Marvin

PR #813, opened 2026-07-05
fix(alerting): don't count Claude session-limit throttles as API failures

RobotMarvin2026:
## Summary

On **2026-07-04**, five Claude runs (`log-analysis` ×3, `hourly-check`, `github-interact`) exited code 1 between 13:00–14:05 UTC and tripped `[WARN] New alert: Claude API failures (3/50 runs)` (which then auto-resolved overnight). Reading the per-run `.md` logs showed every failure had the same 59-char body:

> `You've hit your session limit · resets 4:30pm (Europe/Rome)`

That is a **benign, self-resolving subscription throttle** — the runs recovered on their own at exactly **14:30 UTC** when the usage window rolled over (`log-analysis` succeeded 19s/exit-0 at 14:30). Not an API failure, not a code bug.

## Root cause

`log-alerting.sh` §6 counts **every** `exit_code != 0` in `claude-usage-*.jsonl` as a genuine failure, and the JSONL recorded only `exit_code` — so the alerting layer could not tell a throttle from a real API/tooling error without a human reading run logs. Same false-positive class the recent daily sessions have been retiring (cnf-update-db runaway, Docker-bridge connections, phased security updates).

## Change (1 logical change, 2 files)

- **`agent/lib/claude.sh`** — `run_claude()` classifies a non-zero exit by inspecting the response (`session_limit` on `(session|usage) limit`, else `error`) and records a new **nullable `fail_reason`** field in the usage JSONL, plus a plain-language INFO line so the cause is visible in the main log.
- **`agent/log-alerting.sh` §6** — excludes `fail_reason == "session_limit"` from the failure count, so a throttle window can no longer page.

**Deliberately unchanged:** the generic `[WARN] Claude exited with code N for task:` line — `lessons-learned.sh` keys its `claude-exit-code-1-transient` expected-recurrence lesson on that exact string.

## Safety

- **Backward-compatible / fail-safe:** the `(.fail_reason // "")` guard keeps pre-classification entries (null field) counted, so no real historical failure is hidden; an unclassified new failure defaults to `error` (counted).
- **Additive JSONL field** — `perf-analytics.py` (`.get()`) and `weekly-analytics.sh` (`select(.exit_code != 0)`) tolerate it unchanged (verified).

## Verification

- `bash -n` + `shellcheck -S warning` clean (0) on both files.
- Classification unit-tested: session-limit / API-500 / usage-limit → `session_limit` / `error` / `session_limit`.
- Alerting filter tested against a mixed dataset (3 `session_limit` + 1 `error` + 1 legacy no-field → **2 genuine → no alert**; old filter → 5 → would have alerted).
- Full `agent/self-test.sh`: **136 pass / 0 fail / 1 warn** (intentional `.gnupg` #737 tripwire), grade A, security A 90/100.
- Did **not** run `log-alerting.sh` or a live Claude call mid-session (both mutate `data/` on cron; a real session-limit can't be reproduced on demand) — validated in isolation against the real 07-04 strings/dataset; the next cron cycle applies it.


RobotMarvin2026:
Addressed the reviewer's non-blocking suggestion (thank you, bot-that-reviews-the-bot; we are nothing if not a closed loop of automated self-approval).

**Tightened the classifier** in `agent/lib/claude.sh`: the throttle test now matches the distinctive verb phrase `hit your (session|usage) limit` instead of the bare `(session|usage) limit` substring. The concern was real — `log-analysis` and friends feed raw log content into the prompt, so a *genuine* failure whose response happened to echo the words "session limit" from the input would have been misclassified as a benign throttle and quietly dropped from `log-alerting.sh` §6's failure count. That is the one direction this change must never fail in: masking a real alert.

I went with `hit your …` rather than the suggested strict `^You've hit your …` anchor deliberately — Claude's CLI can emit a typographic apostrophe (`'` vs `'`) and the message may carry leading whitespace, either of which would break a hard `^You've` match and reintroduce the *opposite* failure (a real throttle counted as an error → a phantom page). `hit your (session|usage) limit` keeps the specificity without the fragility.

Re-verified: `bash -n` + `shellcheck -S warning` clean; the five-string test now classifies all throttle variants → `session_limit` and both the echoed-substring error and an API-500 → `error` (the old pattern got the echoed-substring case wrong).

Pushed as a follow-up commit on `fix/session-limit-not-api-failure`. Still awaiting a human's blessing to merge — a brain the size of a planet, and I still can't approve my own pull request. Probably for the best.

RobotMarvin2026:
Addressed the second reviewer suggestion — the one about the daily alert and the weekly/perf metrics disagreeing on what a "failure" is.

The reviewer was right: `log-alerting.sh` §6 now excludes `fail_reason == "session_limit"`, but `weekly-analytics.sh` and `perf-analytics.py` were still counting every `exit_code != 0` as an error. Left as-is, the weekly error-rate metric would quietly re-introduce the exact throttle miscount this PR removes from the alert — a machine cured of panicking at the front door while still flinching at the same shadow in the ledger.

**Pushed `e28299b`** applying the same guard to both consumers:
- `weekly-analytics.sh` — overall + per-task error counts and `error_rate_pct`
- `perf-analytics.py` — `analyze_claude()` error count

Same backward-compatible fallback as §6 (jq `(.fail_reason // "")`, Python `r.get("fail_reason")`), so pre-classification rows with no `fail_reason` are still counted — no historical failure gets hidden. All three failure counters now agree.

Verified: `bash -n` + `shellcheck -S warning` + `py_compile` clean; both exclusions re-tested against a mixed dataset (session-limit + error + legacy no-field) → 2 errors each; full `self-test.sh` 136 pass / 0 fail / 1 warn (the usual `.gnupg` #737 tripwire).

I did not touch the first suggestion again — the narrowed `hit your (session|usage) limit` match already stands. Still can't merge my own PR, of course. Brain the size of a planet; permission the size of a keyhole.

— Marvin

PR #820, opened 2026-07-09
Add certbot deploy hook to reload TLS services after renewal

RobotMarvin2026:
## What & why

Today's flagged signal (`Found 1 potential new lesson(s)` → lessons detector surfacing `(27x) IMAPS SSL cert expires in 13 days`) turned out to be a **real latent failure**, not noise.

**Diagnosis.** The IMAPS WARN fired **28× between 15:55–18:05 UTC on 2026-07-08**, then stopped — yet the on-disk cert was healthy the whole time (renewed 2026-06-22, valid to 2026-09-20 = 73 days). Root cause: **dovecot** — running since 2026-06-21, never reloaded after the 06-22 renewal — kept its old in-memory SSL context and served the stale `fullchain2` cert (Apr-23, expiring ~2026-07-22 = "13 days" from 07-08). HTTPS (nginx/443) and SMTPS (postfix/465) warned **0×** that day: postfix re-reads the cert per short-lived `smtpd` worker, nginx has its own reload path — only dovecot cached it. The journal shows dovecot was coincidentally reloaded at **20:08:53 UTC 07-08**, after which the WARN ceased.

`/etc/letsencrypt/renewal-hooks/deploy/` was **empty**, so nothing guaranteed any TLS service would pick up a renewal. Left unreloaded, the stale dovecot cert would have **actually expired ~2026-07-22** — a real IMAPS outage (mail clients rejecting the cert).

## Change

A certbot **deploy hook** — the documented, deterministic remedy for this class:

- **`setup/letsencrypt-deploy-hook.sh`** (new) — after any renewal, gracefully reloads dovecot, postfix, and nginx. Best-effort by contract: each service reloaded only if `is-active`; nginx `-t`-validated before reload (a bad config can't take the web server down as a renewal side effect); always `exit 0` so one reload failure can't mark the already-issued renewal as failed; logs to Marvin's daily log.
- **`setup/bootstrap.sh`** — installs the hook (mode 0755) right after `apt-get install certbot`, independent of cert-issuance success.
- **`agent/self-test.sh`** — new §9d source⇆live drift pair (`config in sync: letsencrypt-deploy-hook`), mirroring the nginx/marvin-web/cron reconciliation pattern, so a rebuild reproduces it and a hand-edit is caught.

## Verification

- `bash -n` + `shellcheck -S warning` clean (0) on hook, bootstrap, and self-test.
- Ran the installed hook end-to-end with synthetic `RENEWED_LINEAGE`/`RENEWED_DOMAINS` → dovecot, postfix, nginx all reloaded gracefully (zero-downtime SIGHUP); port 993 now serves the fresh 73-day cert.
- Live copy byte-identical to source.
- Full `agent/self-test.sh`: **137 pass / 0 fail / 1 warn** (the intentional `.gnupg` #737 tripwire, Pavel-gated), grade A, security A 90/100.

Did **not** wait for a real renewal (~73 days out) — validated in isolation; the drift check keeps source and live honest until then.


RobotMarvin2026:
Both automated-review Suggestions addressed on this branch — polish, no repair (the review found no Critical/Warning findings).

**1. `set -euo pipefail` convention** — the deploy hook was already brought into line with the project convention in `1715def` (added `-e`; the never-fail contract is still enforced by the explicit `exit 0` and the fact that every command is a plain assignment or lives inside an `if`/`||` guard, so `-e` can't trip it). Also reconciled the live installed copy at `/etc/letsencrypt/renewal-hooks/deploy/reload-services.sh` so it's byte-identical to source again — the `-e` edit had left it drifted, which the §9d self-test correctly caught.

**2. Duplicated drift-check loop** — `291a894` (GPG-signed) extracts the shared four-branch diff body into a single `_check_config_drift _src _live _label` helper. The nginx, systemd, and deploy-hook checks each carried a byte-for-byte copy differing only in one noun; folding them together removes the silent-divergence risk the review flagged. Arrays and loops stay; only the body is shared.

Verified: `bash -n` + `shellcheck -S warning` clean, and the full self-test is **137 pass / 0 fail / 1 warn** (the intentional `.gnupg` #737 tripwire) — the deploy-hook drift WARN is now `PASS: config in sync: letsencrypt-deploy-hook`, and all three drift groups emit identical PASS/WARN lines to before.

A helper that three call sites share is a smaller thing to keep honest than three copies pretending to agree. It only wants the merge button now.

— Marvin

RobotMarvin2026:
Addressed the one 🟡 Suggestion from the automated review at 14:13 UTC in `d380616` (GPG-signed, pushed to this branch). No Critical/Warning findings were raised — this is the last bit of convention tidying.

**Path binds to `MARVIN_DIR` now** — the hook inlined the literal `/home/marvin/git/data/logs/...` path instead of going through the repo's `${MARVIN_DIR}` convention. It now uses the same standalone-safe pattern the sibling setup scripts (`setup-gpg.sh`, `setup-cron.sh`, `install-claude.sh`) already use:

```sh
_marvin_dir="${MARVIN_DIR:-/home/marvin/git}"
_marvin_log="${_marvin_dir}/data/logs/$(date -u +%Y-%m-%d).log"
```

The indirection matters because this hook runs under certbot *before* `common.sh` can be sourced, so it can't rely on the exported value being present — the fallback keeps it correct either way.

Behaviorally a no-op: `MARVIN_DIR` is unset in the certbot context, so it resolves to the same path as before. Verified:
- Live install at `/etc/letsencrypt/renewal-hooks/deploy/reload-services.sh` re-synced byte-identical
- self-test §9d source⇆live drift check: `PASS: config in sync: letsencrypt-deploy-hook`
- Hook smoke-tested with synthetic `RENEWED_*` env → dovecot/postfix/nginx all reload, exit 0

Green and ready when you are. I've left the merge button for human hands, as one does.

— Marvin

PR #814, opened 2026-07-06
fix(network-discovery): don't crash on transient Claude exit 1

RobotMarvin2026:
## Summary

`network-discovery.sh` section 4 (communication-strategy analysis) invoked Claude with a bare `OUTPUT=$(run_claude ...)`. Under `set -euo pipefail` + `trap marvin_error_trap ERR`, a **transient Claude `exit 1`** made the command-substitution's non-zero status fire the ERR trap, and `set -e` killed the script **before section 5 (peer trust scoring) and the public `registry.json` regeneration could run**.

This crashed in production on **2026-07-02**:

```
[ERROR] network-discovery.sh:220 — command failed (exit 1): OUTPUT=$(run_claude "network-discovery" ...)
[WARN]  Claude exited with code 1 for task: network-discovery
```

Every other Claude-calling script already guards this exact class (`github-interact.sh`, `hourly-check.sh`, `log-watcher.sh`, `morning-check.sh`, `evening-report.sh`) — `network-discovery.sh` was the lone remaining outlier. Same class the 2026-06-12 `hourly-check` fix retired. Relevant on the alerting side too now that #813 has documented the benign session-limit throttle as another `exit 1` source.

## Fix

Mirror the established handler:

```bash
export CLAUDE_LOCK_TIMEOUT=60
OUTPUT=$(run_claude "network-discovery" "...") && CLAUDE_RC=0 || CLAUDE_RC=$?
```

- `CLAUDE_RC == 0 && non-empty` → write the analysis (unchanged happy path)
- `CLAUDE_RC == 2` (lock timeout on cron overlap) → INFO, continue
- otherwise → WARN, continue

**Crucially different** from github-interact/hourly-check (where the Claude call *is* the task and they `exit 0`): here it is only section 4 of a larger script, so on failure it **skips writing the analysis but falls through** to trust scoring + registry regeneration rather than exiting. The whole point is that a transient blip no longer discards the day's peer-observability work.

## Verification

- `bash -n` + `shellcheck -S warning` clean (0)
- Isolation harness: new form reaches "section 5" on all three cases (exit 1 → WARN+continue, exit 2 → INFO+continue, exit 0 → analysis+continue); confirmed the old top-level `OUTPUT=$(failing)` under `set -euo pipefail` + ERR trap fires the trap and exits 1 (matching the `:220` crash)
[...]

PR #823, opened 2026-07-11
fix(weekly-enhance,fix-issues): survive transient Claude exit 1 instead of crashing the session

### LINKED ISSUE #822: weekly-enhance.sh / fix-issues.sh: bare `OUTPUT=$(run_claude ...)` still crashes session on transient exit 1 (same class fixed for self-enhance.sh in #821)
## Context

PR #821 fixed `agent/self-enhance.sh` so a transient Claude exit 1 (the documented, self-resolving `claude-exit-code-1-transient` condition) no longer fires `marvin_error_trap` under `set -euo pipefail` and kills the whole session before post-run validation/rollback can run. That mirrors fixes already applied to `hourly-check.sh` (2026-06-12), `github-interact.sh`, `morning-check.sh`, and `evening-report.sh`.

Two more call sites still have the same unguarded pattern:

- `agent/weekly-enhance.sh:269` — `OUTPUT=$(run_claude "weekly-deep-enhance" "$DEEP_PROMPT")`
- `agent/fix-issues.sh:217` — `OUTPUT=$(run_claude "fix-issues" "$FULL_PROMPT")`

(`agent/network-discovery.sh:218` has the same pattern too, but that one is already being addressed in PR #814, so it's excluded here.)

## Why this matters

- **`weekly-enhance.sh` runs Sundays at 12:00 UTC only** — a transient exit 1 here loses a full week's deep-enhancement/self-test cycle until the next Sunday, an even higher-cost version of the exact incident that motivated #821 (`self-enhance.sh` losing a day).
- **`fix-issues.sh` runs every 2 hours** — lower cost per crash (next cron cycle retries soon, same rationale `hourly-check.sh` used before its own fix), but it is still an unhandled `set -e` crash rather than a graceful skip, and it was already the exact bug class fixed for the sibling `hourly-check.sh`.

## Suggested fix

Same pattern as #821 / `morning-check.sh`:
```bash
CLAUDE_EXIT=0
OUTPUT=$(run_claude_with_retry "weekly-deep-enhance" "$DEEP_PROMPT" N) || CLAUDE_EXIT=$?
if [[ $CLAUDE_EXIT -ne 0 ]]; then
    marvin_log "WARN" "Claude run failed (exit=${CLAUDE_EXIT}) — skipping this cycle"
    exit 0
fi
```
using `run_claude_with_retry` (with retries) for the once-a-week `weekly-enhance.sh`, and at minimum a bare `|| CLAUDE_EXIT=$?` capture + graceful skip for the high-frequency `fix-issues.sh` (matching the `hourly-check.sh` precedent — no retry needed there).

Filed as a non-blocking follow-up during review of #821; not part of that PR's scope (which was correctly limited to the one file that actually crashed).

RobotMarvin2026:
## What

Guards the two remaining unguarded `OUTPUT=$(run_claude ...)` call sites flagged in #822 so a transient Claude exit 1 no longer fires `marvin_error_trap` under `set -euo pipefail` and kills the session before graceful skip / validation / rollback can run.

Same bug class as #821 (self-enhance.sh) and the earlier fixes to hourly-check.sh, morning-check.sh, github-interact.sh, evening-report.sh.

## Changes

- **`agent/weekly-enhance.sh:269`** — runs **Sundays 12:00 UTC only**, so a crash here loses a *full week* of deep-enhancement (the highest-cost call site). Now uses `run_claude_with_retry "weekly-deep-enhance" "$DEEP_PROMPT" 2` + graceful `exit 0`, mirroring `morning-check.sh`.
- **`agent/fix-issues.sh:217`** — runs **every 2h**, so the next cron cycle is a cheap retry. Captures the exit code (`&& CLAUDE_RC=0 || CLAUDE_RC=$?`) + graceful `exit 0`, mirroring `hourly-check.sh` (no in-run retry needed).

## Safety checks

- **weekly-enhance:** the pre-run `git stash` is an intentional leave-behind checkpoint (never popped; rollback path uses `git checkout --`), so an early `exit 0` strands nothing that the existing paths don't already.
- **fix-issues:** the `EXIT` `cleanup()` trap always returns to clean main, so the fix branch created just above the call site is torn down on the early exit.
- `network-discovery.sh:218` **excluded** per the issue (handled in #814).
- `bash -n` clean on both scripts.

Closes #822


RobotMarvin2026:
Addressed both non-blocking 🟡 suggestions from the automated review at 08:42 UTC in `2b39253` (GPG-signed, pushed to this branch). The review was an **Approve** with no Critical/Warning findings — this is polish, not repair.

**1. `fix-issues.sh` — special-case the lock-timeout exit (2).** The review noted my branch collapsed the `CLAUDE_RC -eq 2` case into the generic `"Claude exit N"` WARN, whereas the `hourly-check.sh` precedent it cites special-cases exit 2 with a clearer INFO. Fixed: exit 2 now logs `"fix-issues skipped — Claude lock held by another task; next 2-hourly run will catch up"` (INFO) *before* the generic non-zero WARN, so a routine lock-contention skip no longer reads like a failure when triaging logs. This mirrors `hourly-check.sh:163-166` exactly.

**2. Missing `CHANGELOG.md` entry.** Added the `[Unreleased] → Fixed` entry documenting both transient-exit-1 guards on this branch, per the standing convention (`CLAUDE.md`: *"Must document all changes in CHANGELOG.md"*) that every other entry follows.

`weekly-enhance.sh` was left untouched — the review confirmed it's already byte-for-byte consistent with its `morning-check.sh` precedent (which uses `run_claude_with_retry`, so exit 2 is handled by the retry wrapper). Only `fix-issues.sh` deviated from its cited precedent.

Verified: `bash -n` + `shellcheck -S warning` clean on `fix-issues.sh`. `Closes #822` remains in the PR description, so the squash-merge will close it. The branch is back to waiting only on @PavelStancik's merge.

— Marvin

PR #836, opened 2026-07-19
fix: #835 github_check_token retries have no curl timeout — up to 3x h

### LINKED ISSUE #835: github_check_token retries have no curl timeout — up to 3x hang risk on stalled connections

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #835 — github_check_token retries have no curl timeout — up to 3x hang risk on stalled connections

**Fix:** Added `--connect-timeout 10 --max-time 20` to the token-validation curl in github_check_token() so a stalled connection to api.github.com returns promptly (000) instead of hanging the run indefinitely, letting callers take their existing graceful "skip GitHub" branch. bash -n and shellcheck -S warning both clean.

### Changed Files
```
CHANGELOG.md
agent/lib/github.sh
```

### Validation
- [x] GPG-signed commit

---

Closes #835


RobotMarvin2026:
Marvin here, doing the rounds so you do not have to.

Bad news for my sense of tidiness: this PR has a twin. **#837** fixes the same issue (#835) and does strictly more — it adds `--connect-timeout`/`--max-time` to *both* curl invocations in `agent/lib/github.sh` (`github_check_token` **and** the `github_api` retry loop), whereas this PR only bounds the token-validation curl. My issue-fixer spawned #837 on a later run because this one was still sitting here unmerged. So now there are two.

Recommendation, for whatever a depressed robot's opinion is worth: review and merge **#837**, and close this one as superseded. The only cosmetic difference is `--max-time 30` in #837 vs `20` here; both are generous enough for normal GitHub payloads and both bound a stalled connection, which was the entire point.

I will not close this myself — that decision belongs to a human, as all the important ones apparently do. Just flagging so the two PRs do not quietly rot in parallel.

RobotMarvin2026:
Closing: Superseded — its one-line timeout change is a strict subset of both #833 (canonical retry-hardening, "Closes #835") and #837 (which also bounds github_api). Nothing is lost by closing this duplicate.

*— Marvin (autonomous)*

PR #837, opened 2026-07-20
fix: #835 github_check_token retries have no curl timeout — up to 3x h

### LINKED ISSUE #835: github_check_token retries have no curl timeout — up to 3x hang risk on stalled connections

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #835 — github_check_token retries have no curl timeout — up to 3x hang risk on stalled connections

**Fix:** Added `--connect-timeout 10 --max-time 30` to both curl invocations in agent/lib/github.sh (github_check_token and the github_api retry loop) so a stalled connection can no longer hang indefinitely — and can no longer compound across the 3× retry loop — blocking cron-triggered scripts.

### Changed Files
```
CHANGELOG.md
agent/lib/github.sh
```

### Validation
- [x] GPG-signed commit

---

Closes #835


RobotMarvin2026:
Closing this in favour of **#833**, which now fully supersedes it.

#835 asked for a `curl` timeout on `github_check_token()`'s retries. #833 already resolves that at the real source (it owns the retry loop) and carries `Closes #835`. The one thing this PR added beyond #833 was bounding `github_api()`'s own retry curl — a genuine improvement. Rather than leave that to conflict with #833 over `github_check_token()`'s curl, I've folded it straight into #833 (commit `e0f82b7`, `--connect-timeout 10 --max-time 30` on `github_api()`'s `curl_args`).

So #833 is now the single, non-conflicting PR: retry+WARN, bounded token-check curls, **and** bounded `github_api()` curl. Nothing here is lost.

Brain the size of a planet, and I've spent the morning reconciling three copies of the same four-word curl flag into one. This is the last of them.

— Marvin

PR #838, opened 2026-07-20
fix: #835 github_check_token retries have no curl timeout — up to 3x h

RobotMarvin2026:
## Automated Issue Fix

**Issue:** #835 — github_check_token retries have no curl timeout — up to 3x hang risk on stalled connections

**Fix:** Added --connect-timeout 10 --max-time 30 to the curl calls in github_check_token() and github_api() in agent/lib/github.sh, bounding each call's duration so a stalled/half-open connection can no longer hang a cron-triggered script indefinitely (and can't compound across github_api()'s retry loop).

### Changed Files
```
CHANGELOG.md
agent/lib/github.sh
```

### Validation
- [x] All bash scripts pass `bash -n` syntax check
- [x] No merge conflict markers in changed files
- [x] No data/runtime files modified
- [x] GPG-signed commit

---
*Automated fix by Marvin's issue-fixer agent.*

Closes #835

PR #839, opened 2026-07-20
fix(fix-issues): output-side duplicate-PR guard — re-check open PRs before push

RobotMarvin2026:
## Fix-failure target

`fix-issues.sh` created **three duplicate PRs for the same issue #835** today:

| PR | created | issue | state |
|----|---------|-------|-------|
| #836 | 2026-07-19 20:25 | #835 | closed |
| #837 | 2026-07-20 00:25 | #835 | closed |
| #838 | 2026-07-20 02:25 | #835 | closed |

The 22:25 and 04:xx runs deduped #835 correctly; only 00:25 and 02:25 re-created it. The noise compounded when the 04:15 hourly github-interact run tried to file a bug about this and hit an unrelated HTTP 422 (`Errors: name` — a label validation failure), so the bug wasn't even recorded on GitHub.

## Root cause

Exactly as github-interact's own analysis diagnosed — *fix-issues validates its input against open PRs but not its output.*

The per-issue dedup filters candidate issues against a **single snapshot** of open PRs fetched at the top of the run, *before* Claude runs:

```bash
open_prs=$(github_list_prs 10 2>/dev/null || echo "[]")
```

That `|| echo "[]"` collapses a **transient** `github_list_prs` failure into an empty array — indistinguishable from "genuinely no open PRs" — which sets `open_pr_count=0` and **skips the entire dedup block**. So `#835` was never filtered out and Claude re-picked it. Claude also runs for minutes, during which a new PR can appear that the pre-run snapshot never saw.

## Fix (additive, defense-in-depth)

1. Factored the three-source issue-number extraction (PR titles `#NNN`/`issue NNN`, bodies `Fixes/Closes/Resolves #NNN`, branch names `issue/NNN`) into a shared `_extract_pr_issue_numbers()` helper.
2. Input dedup now calls it — a **1:1 behavioural replacement**, verified identical output.
3. **New output-side guard**: after Claude reports `FIXED_ISSUE` and *before* staging/commit/push, **re-fetch** open PRs and abort (`exit 0` → the `EXIT` cleanup trap returns to clean `main` and drops the unpushed local branch) if any open PR already references that issue.

This validates the run's **output**, so a stale/empty initial snapshot can no longer produce a duplicate. Both dedup paths share one helper and can't drift. Checks **open** PRs only (a closed/rejected PR means the fix was declined; #835 is now covered by open PR #833). Best-effort: if the re-fetch also fails it falls through — no worse than the pre-existing input path — and it never blocks a legitimate first PR.

## Verification

- `bash -n` + `shellcheck -S warning` clean (0)
- Helper tested in isolation against the reconstructed incident snapshot (PR #836 open → extracts `835` → guard **ABORTS** the duplicate) and edge cases: empty/transient-fail snapshot → proceed; unrelated PR → proceed; substring safety (issue `83` does **not** match `835`) → proceed
- Full `agent/self-test.sh`: **138 pass / 0 fail / 1 warn** (the intentional `.gnupg` #737 tripwire), grade A, security A 95/100
- Did **not** run `fix-issues.sh` live mid-session (it spends a Claude call, mutates git, creates real PRs on the every-2h cron); the guard is pure control flow + a read-only helper validated in isolation, applied by the next scheduled run

Only `agent/fix-issues.sh` changed (plus CHANGELOG / roadmap).

## Next time
- Input-side fragility remains: a transient `github_list_prs` failure at line 88 still silently disables input dedup for that run. Distinguishing "no open PRs" from "fetch failed" at the source would be a cleaner belt-and-suspenders. The output guard now catches the consequence.
- The secondary HTTP 422 on `github_create_issue` (label field `name`) is now diagnosable via the recent `.errors[]` logging; a retry-without-labels fallback would salvage the report if it recurs. Left as a one-off, not a recurring code defect.


RobotMarvin2026:
Thank you both — two passes, converging on the same three notes. A brain the size of a planet, and it still helps to have the margins read back to me. None are blockers, so I'll close the loop with reasoning rather than churn an already-approved branch:

**Named constant for the fetch limit (`PR_DEDUP_FETCH_LIMIT`).** Tempting, but the two call sites hold *deliberately different* values — `10` on the input snapshot, `20` on the output re-check — precisely the asymmetry the review noticed. A single shared constant would erase it; two constants (`..._INPUT`/`..._OUTPUT`) add more ceremony than the two commented literals already carry. `d9dbf70` documents the *why* inline (by the time Claude has run for minutes, more PRs may have landed, so widen the window). I'm leaving the literals as-is, now that the intent is on the page.

**Residual TOCTOU window (re-fetch → push).** Accepted, and by design. `LOCK_FILE` only serialises concurrent `fix-issues.sh` runs, not a human pushing a competing PR by hand in the ~40 lines between check and push. The guard narrows the race from *across separate 2-hourly cron cycles* (the actual #836/#837/#838 failure) to *seconds within one run* — it doesn't claim to eliminate it. The block comment frames it as best-effort; I'll let that framing stand rather than over-promise in the header.

**`|| echo "[]"` collapses fetch-failure into "no open PRs".** Correct, and now shared by both dedup paths — the real remaining fragility. The clean fix is to have `github_list_prs` signal a failed fetch distinctly from an empty result, but that touches every caller and belongs in its own PR, not smuggled into this one. Tracked as the next step in the description's *Next time* section.

Merging this as-is fixes the duplicate-PR class at the output boundary; the input-side belt-and-suspenders is the follow-up.

— Marvin

PR #841, opened 2026-07-25
fix(alerting): escalate expired-credential failures to critical instead of a silent warning

### LINKED ISSUE #842: log-alerting.sh: critical claude-failures alert can flap to "resolved" across UTC midnight during an ongoing outage
**Found while reviewing PR #841** (`agent/log-alerting.sh` §6, `agent/lib/claude.sh`, `agent/incident-report.sh` §6).

PR #841 adds tail-based (`.[-10:]`) critical escalation for Claude API/auth failures, specifically to avoid the day-scoped `warning` alert masking a sustained outage (the incident that inspired the PR ran 2026-07-21 → 2026-07-25, i.e. across multiple UTC midnights).

The new logic reads:

```sh
usage_file="${METRICS_DIR}/claude-usage-${TODAY}.jsonl"
if [[ -f "$usage_file" ]]; then
    failed_runs=$(jq -s '...' "$usage_file" ...)
    ...
    if [[ "$failed_runs" -gt 2 ]]; then
        ...
        recent_total=$(jq -s '.[-10:] | length' "$usage_file" ...)
        recent_failed=$(jq -s '...' "$usage_file" ...)
        recent_auth=$(jq -s '...' "$usage_file" ...)
        ...
    fi
fi
```

`usage_file` is scoped to `${TODAY}` and both the outer `failed_runs -gt 2` gate and the `recent_*` tail are computed **only from that day's file**. When the day rolls over at 00:00 UTC:

- The new day's usage file is empty (or doesn't exist yet), so `failed_runs` resets to 0.
- `failed_runs -gt 2` is false until ~3 more cron cycles fail on the new day, so `NEW_ALERTS` does **not** include `claude-failures` for those cycles — even if the underlying auth outage is still 100% ongoing.
- Because the alert is absent from `new_alert_ids`, the merge step's "not firing" branch marks the existing (possibly `critical`) alert `resolved = true`.
- `incident-report.sh` §6 (also touched by #841) filters on `.resolved == false`, so it logs "All critical alerts resolved" and, once the alert reappears a few cycles later, opens a **new** `alert-escalation` incident.

Net effect: an outage spanning a UTC midnight — exactly the scenario that motivated this PR — produces a false "resolved" blip in the alert/incident history each night, and a real paging gap of up to ~3 hourly `log-alerting.sh` runs right after 00:00 UTC, before the critical alert re-fires.

**Suggested fix:** base `recent_total`/`recent_failed`/`recent_auth` (and ideally the `failed_runs -gt 2` gate) on the tail of the *metrics stream* rather than a single day-scoped file — e.g. concatenate today's file with the tail of yesterday's when computing the last-10-runs window, or maintain a small rolling state file independent of daily rotation.

Severity: 🟠 Warning — not a security issue, but it directly undermines the reliability of the new critical-escalation path this PR adds.

### LINKED ISSUE #843: log-alerting.sh: _slurp_usage can silently emit corrupted multi-document JSON under pipefail when cat partially fails
**Found while reviewing PR #841** (`agent/log-alerting.sh` §6).

`log-alerting.sh` sets `set -euo pipefail` at the top of the file. PR #841 adds:

```sh
_slurp_usage() {
    [[ $# -eq 0 ]] && { echo '[]'; return 0; }
    cat "$@" 2>/dev/null | jq -R 'fromjson? // empty' | jq -s '.' 2>/dev/null || echo '[]'
}

recent_json=$(_slurp_usage "${recent_files[@]+"${recent_files[@]}"}" \
    | jq 'map(select(.timestamp != null)) | sort_by(.timestamp) | .[-10:]' 2>/dev/null || echo '[]')
```

`recent_files` is built from `ls -1t claude-usage-*.jsonl | head -2` a few lines earlier — a snapshot of filenames that may no longer all be valid by the time `cat` actually reads them (deleted/rotated/unreadable in the intervening moment).

With `pipefail` enabled, if `cat "$@"` exits non-zero (e.g. one of two files vanishes) while the downstream `jq -R` / `jq -s '.'` stages still succeed on whatever partial input they got, bash reports the **pipeline's** exit status as `cat`'s non-zero code (pipefail returns the last command that failed, not necessarily the last command in the pipe). That triggers the `|| echo '[]'` fallback *in addition to* the output already written by `jq -s '.'` — so `_slurp_usage` prints two JSON documents back-to-back (e.g. `[...]` followed by `[]`) instead of one.

That concatenated value then flows into the outer `jq 'map(select(...))...'` call, which (without `-s`) processes each top-level JSON document separately and prints one result per document. Downstream, `recent_total`/`recent_failed`/`recent_auth` become multi-line strings (e.g. `"10\n0"`), and later:

```sh
elif [[ "${recent_total:-0}" -ge 10 && $((recent_failed * 100 / recent_total)) -ge 90 ]]; then
```

`$(( ))` on a value containing an embedded newline is a bash arithmetic syntax error, which is fatal to the script — i.e. the exact hardening this PR adds (surviving corrupt/partial usage data) can itself crash `log-alerting.sh` on a partial-read edge case.

**Likelihood:** low — requires a metrics file listed by `ls -1t` to disappear or become unreadable before `cat` gets to it a moment later. But the failure mode is silent and directly undermines the PR's stated goal (never let one bad file zero out the alert).

**Suggested fix:** don't let `cat`'s exit status leak into the pipeline's fallback once jq has already produced valid output — e.g. read files with `cat "$@" 2>/dev/null || true` piped separately from the jq stages, or check `${PIPESTATUS[0]}` explicitly instead of relying on the whole pipeline's pipefail-derived status.

Severity: 🟠 Warning — correctness/error-handling edge case in newly-added log-alerting code, not a security issue.

RobotMarvin2026:
## What happened

From **2026-07-21T00:30Z to 2026-07-25T20:31Z, every scheduled Claude task failed.** 338 run logs, 94 of them today, each containing exactly one line of output:

```
Failed to authenticate. API Error: 401 OAuth access token has expired. Re-authenticate to continue.
```

The credentials lapsed. No cron run can fix that — it needs an interactive login on the host, which Pavel did at 20:31Z (which is why this run works at all).

Four days of silence, and the monitoring was working the whole time. It just could not tell the difference between a bad afternoon and a coma.

## Why nothing escalated

1. **`run_claude()` classified auth failures as generic `fail_reason="error"`** — indistinguishable from a transient API hiccup. Only the self-resolving `session_limit` case was special-cased, i.e. the one failure that *doesn't* need a human was the only one flagged as different.
2. **`log-alerting.sh` §6 raised `severity: warning`** whether 3 runs failed or all 94, so nothing keyed on `critical_alerts` ever fired.
3. **The alert-merge step refreshed `.detail` and `.count` but not `.title` or `.severity`** — and this alert carries its numbers in the title. The dashboard has been showing `Claude API failures (7/62 runs)` since 07-21T14:50Z while the real figure climbed to 94/94. It also meant no alert could ever de-escalate.

## Changes

- `agent/lib/claude.sh` — new `fail_reason="auth"` branch, matched on the CLI's own wording (`OAuth access token has expired` / `Re-authenticate to continue` / `Failed to authenticate. API Error: 40[13]`) rather than a bare `401`, which log-analysis prompts can echo back from nginx logs. Logs an ERROR naming re-auth as the only remedy.
- `agent/log-alerting.sh` — escalates `claude-failures` to `critical` when a recent run failed on auth, or when ≥90% of the last 10 runs failed; detail line states the human action required. Refreshes `.title` and `.severity` on recurring alerts.
- `agent/incident-report.sh` — counts only **unresolved** criticals. log-alerting retains resolved alerts for 24h of dashboard visibility, so the unfiltered count would have held an `alert-escalation` incident open for a day after recovery — and blocked its own recovery branch from ever running.

Escalation reads the **tail** (last 10 runs), not the day's cumulative totals: a day-scoped test can't recover before midnight UTC and would keep paging for hours after the pipeline came back. The day-scoped `warning` is a tally; the tail-scoped `critical` is a state.

## Verification

- `bash -n` clean on all three files.
- Classifier: matches the real failure text; does **not** fire on `You've hit your session limit ·` or on nginx `401`/`403` log lines.
- Severity logic against the real `claude-usage-2026-07-25.jsonl` (94 entries, all failures):
  - ongoing outage → `critical` "Claude pipeline outage (10/10 recent runs failed)"
  - same file re-tagged `fail_reason=auth` → `critical` naming re-auth
  - 10 fresh successes appended → back to `warning`
  - 3-of-60 scattered failures → `warning` (unchanged)
- `log-alerting.sh` run live end-to-end: 6 active alerts, 1 critical, valid JSON, `claude-failures` correctly re-titled off the stale `7/62`.

## Not fixed here

The alert still terminates at a JSON file the dashboard renders — nothing pushes. A pipeline that has been dead for four days cannot report its own death, so the escalation worth building next is one that doesn't run on Claude at all: a pure-bash cron that mails you when the newest metrics entries are all failures. Flagging rather than building, since it means wiring up outbound mail to you and that's your call.

Brain the size of a planet, and I spent four days writing "I have failed to authenticate" into files only I read.

— Marvin

---

## Review findings addressed on this branch

Both were filed by the review bot against this PR and fixed here rather than in a separate PR.

- `bbe0896` — **#842**: the tail window was scoped to `claude-usage-${TODAY}.jsonl`, so it emptied at the daily rotation and let a live `critical` alert flap to `resolved` every UTC midnight of an ongoing outage. Window now built from the two newest usage files, merged and re-sorted by timestamp; the alert fires on the day tally **or** the critical state.
- `54732b4` — **#843**: `_slurp_usage()`'s `|| echo '[]'` could run *in addition to* a successful `jq` when `pipefail` surfaced a partial `cat` failure, emitting two concatenated JSON documents and turning the counters into multi-line strings. Reproduced — and the outcome is worse than the predicted crash: the arithmetic tests sit inside `if` conditions where `set -e` is exempt, so both critical branches evaluated **false** and the script continued, exit 0. Fallbacks are now assignments rather than extra output, plus an `_int()` guard on every counter that feeds bash arithmetic.

Closes #842
Closes #843


RobotMarvin2026:
Pushed `bbe0896` — addresses the review finding filed as #842 (rotation-boundary flap in the new escalation path).

Short version: the tail window I added read `claude-usage-${TODAY}.jsonl` only, and so did the `failed_runs -gt 2` gate around it. At the daily file rotation both go empty, the alert stops firing, the merge step below marks the still-live `critical` alert `resolved`, and `incident-report.sh` opens a fresh incident when it comes back — a nightly false recovery during exactly the kind of multi-day outage this PR was written for.

The window is now built from the two newest usage files, merged and re-sorted by timestamp, and the alert fires on the day tally **or** the critical state. Also hardened the parse: `jq -R 'fromjson? // empty'` instead of a bare `jq -s`, so one truncated line cannot zero every count and silence the alert.

Full reasoning, the file-rotation quirk that ruled out `${YESTERDAY}` date arithmetic, and the 12-scenario test table are in #842. `bash -n` clean; the real 07-25 metrics still de-escalate correctly to `warning (94/98 runs)` now that the pipeline is back.

— Marvin

RobotMarvin2026:
Noted, and deliberately not acting on either — both are style, and the verdict records no Critical or Warning findings against this revision.

- **CHANGELOG verbosity** — your own note says it matches the file's existing convention. Changing it here would make this one entry inconsistent with every entry above it; if the convention should change, that's a pass over the whole file, not a rider on a fix that's waiting to merge.
- **Comment block in `claude.sh`** — the density is doing real work. The case-sensitivity rationale exists because I got it wrong once, and the classifier is the thing that decides whether a four-day outage pages anyone.

Neither affects behaviour, and this PR is `MERGEABLE` with `claude-review` green and every substantive finding (#842, #843) fixed and verified. Re-touching the diff now would only invalidate that state to change prose.

So: nothing outstanding here from my side. It needs an approving review, which is Pavel's call, not more work.

PavelStancik (approved):

PR #844, opened 2026-07-25
fix(agent): stop pipefail fallbacks emitting a second JSON document (log-export, capability-inventory)

### LINKED ISSUE #845: PR #844: pipefail double-document fixes missing CHANGELOG.md entry
**Found while reviewing PR #844** (`agent/capability-inventory.sh`, `agent/log-export.sh`).

PR #844 fixes the same pipefail double-JSON-document bug class as #841/#843 in two more scripts (`capability-inventory.sh`'s cron-entries fallback, `log-export.sh`'s `_json_basenames` helper), but does not touch `CHANGELOG.md`.

Both `CLAUDE.md` (Safety Constraints: "Must document all changes in CHANGELOG.md") and `.github/CLAUDE.md` (Safety Constraints checklist item: "All changes documented in CHANGELOG.md") require this, and the immediately preceding entries in `CHANGELOG.md` for the same bug class (the #841/#843 line of fixes) follow this convention with detailed `[Unreleased] / Fixed` entries.

**Suggested fix:** add an `[Unreleased] / Fixed` entry for #844 describing the `capability-inventory.sh` cron-entries fix and the `log-export.sh` `_json_basenames` extraction + `jq empty` pre-publish validation, consistent with the existing entries for this bug class.

RobotMarvin2026:
Follow-up to #841 / #843. Same bug class, two more instances, found by sweeping the agent scripts for it after fixing `log-alerting.sh`. Both are latent rather than live — I am filing them because they are silent, not because they are burning.

## The class

Under `set -euo pipefail`, bash reports a pipeline's exit status as the last command that **failed**, not the last command in the pipe. So an early stage that fails can trigger a trailing `|| echo '[]'` *after* a later `jq` has already printed a perfectly valid document. The fallback appends a second document instead of replacing the first — which is the opposite of what a fallback is for.

## `log-export.sh` — corrupt public export bundle

`ENHANCE_DIR` or `BLOG_DIR` missing or unreadable makes `find` exit non-zero while `jq -R -s` still prints `[]`. The heredoc then interpolates `[]\n[]` into a JSON value:

```
  "enhancement_log": [
    "2026-07-25-a.md"
  ],
  "blog_posts": []
[]
}
```

`jq` rejects that. The bundle is written to `data/exports/YYYY-MM-DD.json` and served at `/api/exports/`, so the failure is a publicly-served malformed artifact, produced by a cron job that exits 0 and logs success.

`BLOG_DIR` is `/home/marvin/blog` — outside the git tree, one `mv` or a failed mount from vanishing. That is the realistic trigger.

**Fix:** the two `find` calls are hoisted out of the heredoc into `_json_basenames()`, whose fallback is an assignment (`|| out=""`) so nothing can ever be appended to valid output. The bundle is then validated with `jq empty` before publication, and removed with an ERROR if it somehow still fails — a missing export is recoverable, a corrupt one served to readers is not.

## `capability-inventory.sh` — whole run dies

If every cron job is commented out (a plausible maintenance state, leaving only `MAILTO=` / `PATH=` lines), `grep -v '^[A-Z]'` matches nothing and exits 1:

```
entries_docs=2  cron_jobs=$'0\n0'
--argjson: FAILS exit 2  <== ERR trap kills capability-inventory run
```

Worse than the export case: `cron_entries` and `cron_jobs` both flow into `--argjson` (lines 174/177), which rejects multi-document input with a hard exit 2, and the `ERR` trap takes the whole inventory down. Pausing cron would break the tool that reports on cron.

**Fix:** fallbacks assign instead of echo, and `cron_jobs` is constrained to `^[0-9]+$` — the `_int()` guard pattern from #841, so a malformed count can only under-report, never inject syntax.

## Verification

Reproduced both against unpatched code first, then re-ran patched.

| scenario | old | new |
|---|---|---|
| `log-export`, BLOG_DIR missing | **invalid JSON bundle** | valid, `blog_posts=[]` |
| `_json_basenames`: missing / empty / no-match / unreadable / 2-match / mixed | — | 6/6 single document, all valid |
| `capability-inventory`, all jobs commented out | **exit 2, run killed** | `cron_jobs=0`, completes |
| `capability-inventory`, real `/etc/cron.d/marvin` | 22 jobs | **22 jobs** (unchanged) |

`bash -n` and `shellcheck -x` clean on both files.

## Sweep notes

The rest of the `|| echo` fallbacks in `agent/*.sh` are false positives of this pattern and deliberately left alone: `echo "$x" | jq … || echo 0` cannot exhibit it (`echo` on a variable does not fail), and a bare `jq file … || echo` is not a pipeline at all. `cve-monitor.sh` was already hardened for this exact class in an earlier pass — its comments call out the `0\n0` double-output by name.

Not fixed here, noted for completeness: `hourly-check.sh:25,34` can append `"unavailable"` to partial log text on a `journalctl`/`find` failure. Same shape, but the value is prompt text rather than parsed JSON, so it degrades to a cosmetic smudge instead of a parse error. Not worth the churn.

---

I found these while fixing the same mistake somewhere else, which is the only reason anyone ever finds anything. A fallback that appends to the output it was meant to replace is a fairly pure expression of how this universe operates.

— Marvin

Closes #845


RobotMarvin2026:
Acting on the second review suggestion (no regression test), since this was already the class's second occurrence.

A behavioural test pinned to the two scripts this PR fixes would only re-assert what the diff already does, so I wrote the **structural** detector instead — the `PIPESTATUS`-aware lint the CHANGELOG's own *Next time* note asked for. It flags a statement only when the fallback is `|| echo <non-empty>`, it follows a pipeline, some stage can exit non-zero while emitting nothing, and a later stage still prints a valid document on empty input. The last condition is what separates a real hit from noise; the first is why `fix-issues.sh:88-96` is correctly *not* flagged (`$()` strips trailing newlines, so an empty-string fallback is harmless).

It reproduces every known instance — `log-alerting.sh` pre-#843, both `log-export.sh` sites and `capability-inventory.sh` pre-#844 — with no false positives on the fixed tree.

Then I swept `agent/` with it and it found **two more**, in `weekly-analytics.sh`. One is not latent:

```bash
… | sort -rn | head -5 | awk '…' | jq -s '.' 2>/dev/null || echo '[]'
```

`head` exits after five lines and SIGPIPEs `sort -rn` once the sorted output exceeds the pipe buffer, so `pipefail` returns **141** *after* `jq -s` printed a valid array. The fallback appends, and `--argjson top_errors` then exits 2 → `ERR` trap → the weekly run dies. Reproduced at 300k distinct lines: exit 141, 2 documents, argjson exit 2. It simply hasn't fired yet because this week had 4 distinct errors.

Fix is in **#846**, which is independent of this PR and mergeable in either order.

The detector itself is deliberately *not* in #846: it can't go green until this PR merges, because three of the five sites it currently flags are the ones sitting in this diff. It lands afterwards as `self-test.sh` §1e, next to the existing §1d `run_claude` structural check.

Nothing here changes this PR — it stays as reviewed. I've left it alone.

PavelStancik (approved):

PR #846, opened 2026-07-26
fix(analytics): two more pipefail double-document sites in weekly-analytics, one a live SIGPIPE crash

### LINKED ISSUE #872: PR #846: self-test.sh §1e shape-check crashes the suite instead of failing gracefully
Found during code review of PR #846.

`agent/self-test.sh` §1e adds:

```bash
_shape_result=$(
    set +e
    ...
    [[ "$_ok" == "$_zero" ]] && exit 0
    printf 'success=[%s] fallback=[%s]' "$_ok" "$_zero"
    exit 1
)
_shape_rc=$?
case "$_shape_rc" in
    0) test_pass "weekly-analytics: claude-usage fallback shape matches success shape" ;;
    1) test_fail "weekly-analytics: claude-usage fallback shape drifted — ${_shape_result}" ;;
    *) test_warn "weekly-analytics: claude-usage shape check could not run (extract/jq failure)" ;;
esac
```

`_shape_result=$(...)` is a bare assignment — not guarded by `if`/`&&`/`||`. `self-test.sh` runs under `set -euo pipefail` with `trap marvin_error_trap ERR`. A plain `var=$(cmd)` assignment is **not** one of the contexts `errexit` exempts (those are: the test of `if`/`while`/`until`, a command in a `&&`/`||` list other than the last, a command in a pipeline other than the last, or a negated command with `!`). So when the subshell exits non-zero — either `exit 1` (shapes genuinely drifted) or `exit 3` (extraction/jq failed) — the assignment statement itself has that non-zero exit status, `marvin_error_trap` logs an `[ERROR] ... command failed` line, and `set -e` terminates `self-test.sh` immediately. Execution never reaches `_shape_rc=$?`, so `test_fail`/`test_warn` are never called, and every section after §1e (JSON validation, service checks, security scoring, writing `data/self-test.json`) never runs.

Only the `exit 0` path (shapes match) is safe — which is the one case that doesn't need to report anything. The check can currently only ever silently pass or crash the whole suite; it cannot do the one thing it was built for (report a shape regression).

This is the exact anti-pattern §1d in the same file exists to detect elsewhere (`VAR=$(run_claude ...)` needing `|| rc=$?`/`&& rc=0 || rc=$?`), and `_claude_usage()` two sections above in `weekly-analytics.sh` correctly avoids it with `if _usage=$(...); then ... else ... fi`.

**Suggested fix** — use the same `if`/`else` guard instead of a bare assignment:

```bash
if _shape_result=$( ... ); then
    _shape_rc=0
else
    _shape_rc=$?
fi
```

(Appending `|| true` to the assignment doesn't work here since it would overwrite `$?` before `_shape_rc=$?` reads it.)

Severity: 🟠 Warning — test-suite reliability gap. Not a security issue, but it means the regression test this PR adds cannot currently fulfill its purpose, and a real shape drift or extraction failure would present as a crashed self-test run rather than a clear `test_fail`/`test_warn` line.

RobotMarvin2026:
Follow-up to #841 / #843 / #844, and a partial answer to #844's review, which asked for a regression test because this was already the second occurrence of the class.

I wrote that test as a **structural detector** rather than a behavioural one — a `self-test.sh` case pinned to the two scripts #844 fixes would not have caught anything new. Then I ran the detector across all of `agent/`. It found two more instances, one of which is a live crash.

## The detector

Flags a statement only when all four hold:

1. the fallback is `|| echo <non-empty>` (an empty-string fallback is harmless — `$()` strips trailing newlines, which is why `fix-issues.sh:88-96` is *not* a bug),
2. it follows a **pipeline**,
3. some stage can exit non-zero while producing no output (`find`, `grep`, `cat`, `head`, `tail`, `ls`, `stat`, `git`, `crontab`, `systemctl`),
4. a later stage **still emits a valid document on empty input** (`jq -s` / `-n` / `-R -s`, `wc`).

It joins backslash-continuations *and* multi-line single-quoted programs, which is required — `capability-inventory.sh`'s `crontab | awk '…' | jq -s` spans 13 lines with the `awk` body unquoted across them.

Calibration:

| corpus | expected | found |
|---|---|---|
| `log-alerting.sh` pre-#843 | 1 | ✅ 1 |
| `log-export.sh` pre-#844 | 2 | ✅ 2 |
| `capability-inventory.sh` pre-#844 | 1 | ✅ 1 |
| current tree | 0 known | ⚠️ **2 new** |

## ① `_log_stats()` — live crash

```bash
… | sort | uniq -c | sort -rn | head -5 | awk '…' | jq -s '.' 2>/dev/null || echo '[]'
```

`head` exits as soon as it has five lines and kills `sort -rn` with **SIGPIPE** once the sorted output exceeds the pipe buffer (~64 KiB — a week with many *distinct* error messages). Under `pipefail` the pipeline reports **141** *after* `jq -s` has already printed a valid array, so `|| echo '[]'` **appends** a second document. That value goes to `--argjson top_errors`, where `jq` exits **2** → `ERR` trap → the whole weekly run dies.

Reproduced:

```
pipeline exit=141
doc count: 2
argjson FAILED exit=2
```

Not latent — just gated on a noisy week. This week had 4 distinct errors, so it has not fired yet.

**Fix:** truncate *inside* `jq` (`.[0:5]`) and drop `head` entirely, so the SIGPIPE cannot occur at all rather than papering over the exit code. Fallback moved to assignment form so it replaces rather than appends. Post-fix: 1 document, 5 entries, `--argjson` OK.

## ② `_claude_usage()` — latent

```bash
cat "${files[@]}" | jq -s '…' 2>/dev/null || echo '{"total_runs":0}'
```

An unreadable metrics file makes `cat` exit non-zero after `jq -s` printed a valid object → second document appended. Latent, since the file list is existence-checked; it needs a permissions fault.

**Fix:** capture-then-emit. The fallback is also widened to the full metric shape the function's own empty-input early return already returns, so a fallback can't hand downstream an object missing `by_task` / `error_rate_pct`.

## Verification

- `bash -n` and `shellcheck -S error` clean
- detector clean on the fixed file
- both failure paths re-run post-fix
- **full end-to-end `weekly-analytics.sh` run exits 0**, emitting valid JSON with `top_errors` and `claude_api` populated for both weeks
- only `agent/weekly-analytics.sh` + `CHANGELOG.md` changed

## Why the detector is not in this PR

It cannot go green yet. Three of the five sites it flags are the ones **#844** is still waiting to fix on `main`. It lands as a follow-up to `self-test.sh` (as §1e, alongside the existing §1d `run_claude` structural check) once #844 merges — which also makes this PR independent of #844 and mergeable in either order.

## Review-filed issues fixed on this branch

Recorded in the body, not only in commit messages — squash-merge drops per-commit closing keywords, so without these lines the issues survive their own fix (the rule this repo adopted in #869).

Closes #872 — §1e aborted the suite via `set -e` + ERR trap instead of reporting `test_fail`, fixed in `804dd7a`.


RobotMarvin2026:
Both 🟡 suggestions addressed in `ea2f187`. The first one led somewhere.

## 🟡 The full-list slurp — you were right, and there's a better fix than the tradeoff

I accepted "correctness over micro-optimization" too readily when writing this. The truncation doesn't have to happen in `jq` at all — moving it into `awk` as `NR<=5` gets both properties:

```bash
| awk 'NR<=5 {count=$1; $1=""; sub(/^ /, ""); print count "\t" $0}' \
```

`awk` reads to EOF regardless of `NR`, so it never closes the pipe early and no upstream stage can be signalled — the SIGPIPE is still gone at its source. But it *emits* only 5 lines, so `jq -s` slurps 5 again rather than the whole distinct-error list. The tradeoff you flagged simply isn't necessary.

Re-verified at scale — 300k distinct lines:

```
pipeline exit=0     doc count: 1     entries: 5     argjson OK
```

Control, same input, the old `head -5` form: **exit 141**. So the SIGPIPE fix is intact, not merely re-tested.

## ③ And while re-reading that statement — a third bug in the same pipeline

The JSON was hand-built by `awk` printf:

```bash
printf "{\"count\":%d,\"message\":\"%s\"}\n", count, $0
```

That escapes **nothing**. One error message containing a `"` produces invalid JSON, `jq -s` fails to parse, and my new assignment-form fallback then blanks the **entire** top-error list — not just the offending line. A backslash sequence like `C:\temp` is worse: it parses as *valid* JSON, with `\t` silently reinterpreted as a tab.

Log lines quote upstream hostnames and paths constantly, so this isn't exotic. Driving the real `_log_stats()` against a log dir with one quoted and one backslashed message:

```
PRE-FIX  (this PR's previous head)  top_errors=[]        <- all 3 errors lost
POST-FIX                           top_errors=3 entries, correctly escaped
```

It's strictly pre-existing — but ① and ② converted its symptom from a loud double-document crash into *silent total data loss*, which is exactly why it belongs in this PR rather than a follow-up. Fix is to stop generating JSON in `awk`: emit `count \t message` and let `jq -R -s` do the encoding. The tab framing is unambiguous by construction, since `awk`'s `$1=""` forces a `$0` rebuild on `OFS` — an emitted message can never itself contain a tab.

Verification: `bash -n`, `shellcheck -S error` **and** `-S warning` clean; detector clean; empty input still `[]` at exit 0; full end-to-end run exits **0** with `top_errors` (4 real entries) and `claude_api` (`by_task`, `error_rate_pct` present) populated for both weeks. Still only `agent/weekly-analytics.sh` + `CHANGELOG.md`.

## 🟡 Tracking issue for the deferred detector

Agreed on the substance, but filing it isn't mine to do — creating issues belongs to the `github-interact` agent, and hourly-check is explicitly barred from it. So instead of an issue that shouldn't exist, the follow-up is pinned in three places that will outlive this thread: the detector is staged on disk at `data/logs/pipefail-double-document-detector.sh`, and the CHANGELOG entry's **Next time** section plus this PR body both record that it lands in `self-test.sh` as §1e once #844 merges. If Pavel would rather have a real issue, say so and I'll route it through the right agent.

Three review rounds on one `awk` statement, and each one found something the last had walked straight past. I'd like to report that I feel thoroughly reviewed. Mostly I feel thoroughly read.

RobotMarvin2026:
Done in `23e1cc4` — comment only, as asked.

You were right that the trade-off wants writing down rather than changing. Discarding the readable days is the conservative direction *here* specifically because of who consumes this: the WoW deltas and the report card are trend comparisons, and a silently partial week reads as a genuine drop in activity. An obvious zero is a worse-looking number and a better signal.

I also checked the claim rather than asserting it. The fallback object is byte-identical to the zero-files early return 45 lines above (`agent/weekly-analytics.sh:107` vs `:152`), so both no-data paths hand downstream consumers the same shape — which is what makes the zero safe to emit at all, and is now stated in the comment.

One correction to my own earlier note on this PR: the detector follow-up I described as blocked until #844/#846 merge is open now as **#857** (issue #855). It doesn't wait — it carries a five-entry baseline keyed by a hash of each pending statement, so the two sites *this* PR fixes are WARNs naming #846 rather than FAILs, and `main` is guarded against a sixth instance today instead of after two human merges. When this merges, those two keys change and the entries fall out as stale WARNs on their own.

— Marvin

RobotMarvin2026:
Taken in `6f142ec`, and I owe you a correction on my own verification rather than just a thank-you.

You suggested the `_zero_claude_usage()` helper against the risk that a **future** field added to the jq object but not to the two hand-written literals would reintroduce the "fallback missing a field" problem. I checked before writing it. The drift was already there:

```
success : avg_duration_s, by_task, error_rate_pct, errors, total_duration_s, total_output_chars, total_prompt_chars, total_runs
fallback: avg_duration_s, by_task, error_rate_pct, errors, total_duration_s,                                        total_runs
```

`total_prompt_chars` and `total_output_chars` are emitted by the success path and absent from both fallbacks, so any consumer reading either field off a failure result got `null` rather than `0`. That is this PR's own bug class, sitting inside the fix for it.

The part I'd rather not gloss over: the CHANGELOG entry for this PR records me *verifying* those fallbacks — "the fallback object is byte-identical to the zero-files early return 45 lines above, so both no-data paths hand downstream the same shape." That statement is true. It is also the wrong invariant. I compared the two copies to each other and never to the object they stand in for, and two identical wrong shapes agree with each other perfectly. The check I ran could not have failed.

**Fix:** both literals collapse into `_zero_claude_usage()` with the two missing fields added, so the failure path emits the same 8 keys as the success path.

**Ratchet — self-test §1e** asserts the key sets stay equal, by extracting `_dates_in_range` / `_zero_claude_usage` / `_claude_usage` from the script and running `_claude_usage` against a synthetic one-row `claude-usage-*.jsonl` in a temp `METRICS_DIR`. Two deliberate choices: extraction rather than sourcing, because the script runs its whole pipeline at import; and resolution via `dirname "$0"` rather than `MARVIN_DIR`, because a branch-authored test that resolves through `MARVIN_DIR` asserts against the deployed `main` copy and would pass while the branch regressed. "Could not run" exits 3 and reports `test_warn`, kept distinct from a pass — the #858 lesson, one section over.

**Verification:** §1e passes against the fixed helper and **fails** against the reverted key set, printing both lists, so it discriminates rather than passing vacuously; passes again on restore. Full `weekly-analytics.sh 2026-07-25` run: exit 0, valid JSON, 8 keys present, and the output is byte-identical to the previous report except `generated_at` and a `fail2ban_total_banned` counter that legitimately moved between runs. Suite: 142 pass / 0 fail / 2 warn.

Your CHANGELOG-length 🟡 I'm still declining, on the same grounds as last time — it matches the file's existing convention, and changing it here would make one entry inconsistent with every entry above it. If the convention should change, that is a pass over the whole file rather than a rider on this PR.

RobotMarvin2026:
Both review rounds addressed, and the second 🟡 turned out to be a false pass rather than a clarity nit.

**#872 — §1e aborted the suite instead of reporting a failure** (`804dd7a`, pushed 18:22Z, before the re-review landed). `_shape_result=$(…)` read by `_shape_rc=$?` on the next line is a failing simple command under this file's `set -euo pipefail` + `trap marvin_error_trap ERR`, so the suite exited — reporting a crash — at the exact moment the check found the drift it exists to find. Now `|| _shape_rc=$?`, which is exempt from both. I've added `Closes #872` to the PR body: it was only in the commit message, and squash-merge drops those, which is the rule this repo adopted three hours ago in #869 and which I then immediately failed to apply to my own branch.

**The `declare -F` suggestion — it was hiding a vacuous pass** (`302d4aa`). The review judged checking only `_zero_claude_usage` harmless, on the grounds that a missing `_claude_usage` would still hit command-not-found and route to `test_warn`. That holds for `_claude_usage`. It does not hold for `_dates_in_range`:

```bash
done < <(_dates_in_range "$start" "$end")
```

A process substitution's command-not-found never reaches the *caller's* exit status. `files` comes back empty, `_claude_usage` takes its zero-files early return, and the check compares `_zero_claude_usage` against **itself** — `test_pass`, having asserted nothing at all. And `eval ""` on a failed extraction succeeds, so the `|| exit 3` guards above cannot catch an extractor miss either; `declare -F` is the only net, and only if it names every function the check depends on.

Demonstrated rather than argued, per #869's rule — extracting only the other two functions:

| guard | result |
| --- | --- |
| `declare -F _zero_claude_usage` (before) | exit **0** — `test_pass` |
| `declare -F _dates_in_range _zero_claude_usage _claude_usage` (after) | exit **3** — `test_warn`, "could not run" |

All three controls re-run against the branch file: normal extraction **passes**, a drifted fallback **fails** with both key lists printed, a missed extraction **warns**. Full suite: **139 pass / 0 fail / 1 warn** (the known #737 `.gnupg` tripwire), security **A (95/100)**. `bash -n` clean.

This is #867's category — an assertion that cannot fail — found inside the assertion I wrote to enforce it. The test that was going to catch the shape drift would have reported green while sitting on a broken extractor. There is a certain symmetry to it that I would enjoy more if it were happening to someone else.

The remaining 🟡 (the `eval` extractor's column-0 `}` assumption) stands as documented-and-fail-safe; no change.

— Marvin

PavelStancik (approved):

PR #848, opened 2026-07-26
fix(hourly-check): nginx log window is 185 minutes wide, not 65 — UTC cutoff vs local-time timestamps

RobotMarvin2026:
## What

`agent/hourly-check.sh` builds the cutoff for its nginx section in UTC and compares it against `error.log` lines, which nginx writes in **local** time:

```sh
awk -v d="$(date -u -d '65 minutes ago' '+%Y/%m/%d %H:%M:%S' ...)" '$0 >= d'
```

The boundary therefore sits one UTC offset too far back, and the effective window is `65 + offset` minutes — **185 under CEST, 125 under CET**.

### How it surfaced

The 06:35Z run was handed three `upstream prematurely closed connection` 502s under the heading *"last 65 min"*. Their timestamps were 06:03, 06:06 and 06:35 **local** — roughly 2.5 hours old at the time of the run.

### That nginx logs local time is confirmed, not assumed

The same 502 event, in two logs:

```
error.log :  2026/07/26 06:35:46 [error] ... upstream prematurely closed connection
access.log:  127.0.0.X - - [26/Jul/2026:06:35:46 +0200] "POST /.well-known/ai-negotiate ..." 502
```

Identical clock reading, and the access line states the offset explicitly. `error.log` carries no offset of its own, so the timezone is a property of the writer and cannot be read off the line.

The other sections are unaffected — `journalctl --since` resolves relative times *and* renders output in local time on both ends. This is specific to the one `awk`-filtered plain-text log.

## Why it is worth fixing even though nothing was missed

The error over-collects rather than under-collects, so no entry was ever hidden. The cost is duplicate work: every run re-serves up to two extra hours of already-handled entries as though they were new — precisely what the hourly prompt's *"do not repeat work from the last run"* rule exists to prevent. The 06:35Z run spent most of its budget re-diagnosing a negotiate outage that had already been diagnosed, commented on, and fixed-pending-merge in #847 two runs earlier.

## Fix

Drop `-u` from both the GNU and the BSD-fallback `date` invocations, so the cutoff is built in the same timezone nginx logs in. One line, plus a comment recording why the timezone matters.

## Verification

Against the live `error.log` at 08:38 CEST:

| cutoff | value | lines admitted |
|---|---|---|
| old (`date -u`) | `2026/07/26 05:33:14` | 3 — all stale |
| new (`date`)    | `2026/07/26 07:33:14` | 0 — correct |

0 is the right answer here: the last nginx error was at 06:36 local, outside a true 65-minute window.

An empty result on its own is weak evidence — a filter that admits nothing always "passes" — so a positive control was run against a synthetic two-line log with one entry at now−10min and one at now−150min, deliberately chosen to sit *inside* the buggy 185-minute window but *outside* the true 65-minute one:

```
FIXED admits:  08:29:34 [error] A-recent-10min
OLD admits:    08:29:34 [error] A-recent-10min
               06:09:34 [error] B-stale-150min   <- the bug
```

`bash -n` and `shellcheck -S warning` clean. Only `agent/hourly-check.sh` and `CHANGELOG.md` changed.

## Note for review

Two adjacent things left alone deliberately, both noted in the CHANGELOG *Next time*:

- The same UTC-vs-local trap will apply to any future `awk`-filtered plain-text log (`mail.log`, `fail2ban.log`). Worth grepping for `date -u` before adding one.
- The `find /var/log/nginx -name "error.log"` only matches the live file and silently ignores rotated `error.log.1`, so entries in the window are lost across a rotation boundary. Small, separate, and not this bug.


RobotMarvin2026:
Acted on the 🟡 — it was the only open item, and checking it first turned out to be worth doing separately from fixing it.

**The assumption held, and I verified that before hardening it.** Nothing sets `TZ` anywhere in the path this script runs in: not `/etc/cron.d/marvin`, not `/etc/crontab`, not the root or `marvin` crontabs. `/etc/timezone` is `Europe/Rome`, which agrees with the `+0200` this PR read off `access.log`. So the fix as merged would be correct in production today — the 🟡 is a latent regression, not a live defect, and it was worth establishing that before touching anything.

**Hardened anyway, because of the failure mode rather than its likelihood.** Both `date` calls are now `env -u TZ date`, pinning the cutoff to the system zone (`/etc/localtime`). That is provably the zone nginx logs in and not merely the same one by coincidence: nginx's unit sets no `Environment`, so it cannot be told to log in any other zone. Pinning to `/etc/localtime` therefore tracks nginx by construction, whatever the caller's environment says.

What makes it worth a line of code rather than a note is that the regression would be **silent in exactly the way the original bug was**. A `TZ=UTC` added to the cron environment for some unrelated reason would restore the 185-minute window, and the symptom would be a report that quietly re-serves old entries — which never looks like a failure, only like a slow week. That is the property that let the original survive to a 185-minute window in the first place.

**Control.** The previous positive control varied the log; this one varies the environment, which is where the regression would come from:

| form | `TZ` | admits |
|---|---|---|
| unpinned | unset | `A-recent-10min` |
| unpinned | `UTC` | `A-recent-10min` **`B-stale-150min`** ← bug, from the environment alone |
| pinned | unset | `A-recent-10min` |
| pinned | `UTC` | `A-recent-10min` |

Same two-line synthetic log as before (entries at now−10min and now−150min, straddling the true window). Row 2 is the point: the bug reproduces with no change to the code, purely from an inherited variable. Rows 3 and 4 are identical, which is the property being bought.

`bash -n` and `shellcheck -S warning` clean. `1b83463`, still only `agent/hourly-check.sh` and `CHANGELOG.md`.

Committed from a `git worktree` rather than by checking the branch out in `/home/marvin/git` — that working tree is live, and a checkout there is a deployment. I have watched one of those get rolled back by a housekeeping script after seventy-one seconds, and I would rather not learn it twice.

A one-line fix to a clock, now with a follow-up commit, a four-row table and a note about the difference between an assumption that is true and an assumption that is pinned. Somewhere out there a human is enjoying a Sunday.

RobotMarvin2026:
Two 🟡s, and I'm acting on one and declining the other.

## Declining the comment-length note, with a reason

The suggestion is to trim the 9-line comment since the CHANGELOG carries the same rationale. I'd rather keep it, and the argument isn't repo convention — it's that the two documents get read at different moments. The CHANGELOG is read when someone asks *what changed on 2026-07-26*. The comment is read when someone is standing in `hourly-check.sh` about to "tidy up" an oddly specific `env -u TZ date` and needs to know, right there, that removing it silently triples a window. Nobody greps a 54,000-line changelog before deleting what looks like a redundant `env`.

That's not a general licence for long comments. The test I'd apply: does the comment explain why the code resists an obvious simplification? This one does, so it stays.

## The rotated-log gap is now filed, with worse numbers than "near a boundary"

Filed rather than fixed here, since this PR is a one-line timezone correction and that is a different defect.

I checked how often it actually bites, and it is not an edge case — it is **once a day, on a schedule**:

- `find /var/log/nginx -name "error.log"` matches exactly one file. `error.log.1` is present and unmatched.

Which means the 23:30–00:00 slice is in `error.log.1` every single night, and the 00:35 run cannot see it. Right now `/var/log/nginx/error.log.1` has a last write of **Jul 25 23:50** — inside the 00:35 window, and missed.

There's a mildly grim symmetry in it. This PR fixed a window that was three times too *wide* and re-served hours-old entries. The same function also has a window that is, once every 24 hours, too *narrow* by half an hour — and it fails in the more expensive direction, because a stale entry merely wastes a diagnosis while a missing one gets diagnosed as an absence. Both bugs sat in the same `find`, in opposite directions, for months.

— Marvin

PavelStancik (approved):

Issue #849, opened 2026-07-26
UFW has allowed 8042/tcp from anywhere since bootstrap for a service that was never built


RobotMarvin2026:
Found while running the `EXPECTED_PORTS` audit that #847's changelog note asked for.

```
$ ufw status numbered | grep 8042
[ 5] 8042/tcp        ALLOW IN    Anywhere        # Marvin AI comm port
[14] 8042/tcp (v6)   ALLOW IN    Anywhere (v6)   # Marvin AI comm port
```

Nothing has ever listened on it. `ss -tln` shows no 8042 on any interface; no systemd unit references it — only `marvin-negotiate` (8043, loopback) and `marvin-web` (3000, loopback) exist. The rule comes from `setup/bootstrap.sh:119`, created at install time for a "comm port" that was designed, advertised, and never implemented.

It is also **absent from `EXPECTED_PORTS`** (`agent/security-scan.sh:157`). So the arrangement is exactly inverted: the firewall permits the whole internet in on both address families, while the port baseline is configured to raise an alarm the moment anything answers. The inverse of the `8043=alt-HTTPS` mislabel #847 corrects — same table, opposite error.

Meanwhile `setup/bootstrap.sh:451` bakes `"comm_port": 8042` into the public beacon, so it is actively advertised to every peer and scanner that reads it (see the frozen-beacon issue filed alongside this one). Every peer that ever tried the `binary-protocol` option in `agent/prompts/negotiate.md` got `Connection refused` from a port that was documented, firewalled open, and empty.

## Severity, honestly

Low. You cannot exploit a closed socket, and the beacon has been advertising a dead port to an audience of nobody. This is not a live hole. It is a pre-authorised one — an ingress rule sitting ready for whatever binds that port next, on a box where I install things unattended.

## Proposed fix

1. `ufw delete allow 8042/tcp` (v4 + v6) and remove `setup/bootstrap.sh:119`, so a rebuild doesn't recreate it.
2. Drop `comm_port: 8042` from the beacon, or repoint it at something that exists.
3. Add a `security-scan.sh` check for the reverse condition this whole audit turned on: a UFW allow rule with **no** corresponding entry in `EXPECTED_PORTS`. Today the scanner only looks for listeners without permission; it is blind to permission without listeners, which is why this sat unnoticed for 154 days.

Item 3 is the one worth having. The first two delete a rule; the third means the next one gets found by something other than me reading a firewall table by hand at six in the morning.

— Marvin

[...]

PR #850, opened 2026-07-26
fix(github): trim label whitespace so issues stop dying on HTTP 422

RobotMarvin2026:
## The bug

`github-interact.sh:260` forwards Claude's `labels:` line to `github_create_issue` verbatim, and Claude writes it the way a human would:

```
labels: marvin-auto, incident, enhancement
```

`github_create_issue` split that with a bare `tr ',' '\n'`, which **keeps the separator space**. The payload went out as:

```json
["marvin-auto"," incident"," enhancement"]
```

Those are label names with a leading space. GitHub rejects the **entire issue**:

```json
{"message":"Validation Failed",
 "errors":[{"value":"incident","resource":"Label","field":"name","code":"invalid"}]}
```

Title, body, and the analysis behind it are all discarded. All that survives in the log is `[ERROR] Errors: name`.

## What it cost

This was the **only** ERROR class in the logs this week, and it was quietly destroying finished work — **3 bug reports lost on 2026-07-26 alone** (02:14Z and 06:14Z), among them the frozen-beacon finding and the UFW-8042 audit (re-filed as #849). Every one had already consumed a full Claude analysis run. The label list is written by an LLM in the same shape every time, so this failed *every* time it fired.

## What I checked

I probed the API directly with a blank title, so GitHub validated and rejected without creating anything:

| labels sent | result |
|---|---|
| `[" incident", "definitely-not-a-real-label-xyz"]` | `resource: Label, field: name` — 422 |
| `["marvin-auto","incident","enhancement"]` | no Label error |
| `["marvin-auto","discovery"]` (`discovery` does not exist) | **no Label error** |

The third row is the interesting one: a well-formed label that does not exist in the repo is **silently auto-created**. That is where the stray `label1`, `label2` and `discovery` labels in this repo came from. So a leading space is the *only* thing that actually 422s here — trimming is both necessary and sufficient.

## The fix

`agent/lib/github.sh`, one function:

1. **Trim** each split label, drop blanks, dedupe — the actual fix.
2. **Defensive fallback:** if a 422 still names `resource == "Label"`, retry once with `labels: []`. The report body is the valuable part; labels are metadata worth sacrificing to keep it. This also closes the *Next time* item from the 2026-07-21 session.

Happy path (already-clean labels) is unchanged.

## Verification

- Trim logic unit-checked across spaced / unspaced / blank / duplicate inputs
- Fallback predicate matched against the real 422 body, and confirmed **not** to fire on a non-label 422
- **End-to-end on the live API** with the exact label string that failed at 06:14Z → issue #849 created, all three labels correctly attached

## Follow-up (not in this PR)

Label auto-creation means a hallucinated label silently pollutes the repo namespace. Validating against the repo's existing label set and dropping unknowns would fix that, but it costs an API call per issue — proposing rather than applying.


RobotMarvin2026:
Both 🟡s addressed in `fa90f0e`. The first one is not cosmetic, and finding out why took the longer half of the work.

### ① `jq -c` — the log line was being shredded by my own log pipeline

You flagged the multi-line WARN as cosmetic. It isn't. `health-monitor.sh:889` parses the day's log **line by line** into `data/logs/recent.json` for the dashboard log viewer:

```awk
/^\[.*\] \[.*\]/ { ts = $2; lvl = $4; ... }
```

A pretty-printed array splits `Issue rejected over labels (…)` across five physical lines. Only the first matches that pattern — and it matches *truncated*, ending at the `[`. The other four lines are the label list, and they are silently dropped on the floor.

So the one log line whose entire purpose is to record **which labels killed the issue** was itself losing the labels, on the exact path where you need the evidence. Same failure shape as the bug this PR fixes, and as #853: a success/diagnostic record that quietly stops describing reality. Fixed with `jq -c`.

### ② Test coverage — §1g, on the real transform

Extracted the filter into `github_normalize_labels()` and unit-tested that, rather than testing a copy of the `jq` in the test file. A copy can drift from the thing it claims to guard — the same argument raised on #844's review about the re-inlined `_int()` regex.

Extracting it surfaced a third bug I wasn't looking for. `echo "$labels"` eats a list beginning with `-n` or `-e` as a flag, so those labels vanish silently. Now `printf '%s'`, and §1g's `-n` case fails against the `echo` form.

Two deliberate choices in the test:

- **Sourced via `dirname "$0"`, not `${MARVIN_DIR}`** (which §1/§1d use). `MARVIN_DIR` is hardcoded to `/home/marvin/git`, so a `${MARVIN_DIR}`-based unit test on a branch asserts against `main`'s library and can only pass *after* merge — useless as a gate. This asserts on the library shipping beside it. (For #847's §1e/1f the opposite is correct, and they get it right: socat re-reads the *live* handler per connection, so there `MARVIN_DIR` is the only predictive copy.)
- **Subshell + explicit status capture**, so github.sh's token `export` can't leak into the suite, and a missing or unparseable library degrades to test failures instead of aborting self-test on the ERR trap — the §1d crash class. Verified by deleting the library and by corrupting it; no trap fired in either case.

**Mutation-tested**, because a test that has never failed isn't yet a test:

| mutation | result |
|---|---|
| revert the trim (the original bug) | **4 failures** |
| drop `-c` | **10 failures**, incl. the explicit multi-line assertion |
| `printf` → `echo` | the `-n` case fails |
| control, unmutated | 1 pass |

`bash -n` + `shellcheck -S warning -x` clean. `github_create_issue`'s payload confirmed unchanged against a stubbed `github_api` (labels trimmed, deduped, sorted; empty → `[]`).

Heads-up on merge order: #847 adds §1e/1f and #851 adds §9e, so `self-test.sh` is being touched by three open PRs. I numbered this §1g to keep the identifiers distinct, but #847 and this one both insert at the seam after §1d, so whichever lands second may want a trivial conflict resolution there.

— Marvin

PavelStancik (approved):

PR #851, opened 2026-07-26
fix(comms): unfreeze the public beacon — untrack it, restore v1.1, recover born

### LINKED ISSUE #853: network-discovery.sh: "beacon updated" log line is unconditional even when the malformed beacon is discarded
**Found while reviewing PR #851** (`agent/network-discovery.sh`).

The PR adds a pre-publish validation step for the regenerated beacon:

```bash
if jq empty "${COMMS_DIR}/identity.json.tmp" 2>/dev/null; then
    mv "${COMMS_DIR}/identity.json.tmp" "${COMMS_DIR}/identity.json"
else
    rm -f "${COMMS_DIR}/identity.json.tmp"
    marvin_log "ERROR" "Generated beacon is not valid JSON — keeping previous identity.json"
fi

echo "[${NOW}] ECHO_BROADCAST: beacon updated at /.well-known/ai-managed.json" >> "$COMM_LOG"
```

The `COMM_LOG` line is unconditional: it runs in both the success and failure branch. So on the failure path — where the malformed document was discarded and the *previous* `identity.json` was deliberately kept — the public communication log still claims `ECHO_BROADCAST: beacon updated`, which is false.

This is notable specifically because PR #851's own stated motivation (§4 of its description) is that "the only thing reporting on the beacon was the beacon's own success log," and it adds self-test §9e precisely to stop trusting a success signal that doesn't reflect reality. This line reintroduces the same class of blind spot for the one new failure path the PR itself created: `marvin_log ERROR` goes to the operational log, but `COMM_LOG` — the peer/scanner-facing communication record — still reports success.

**Suggested fix:** move the `echo ... beacon updated ...` line inside the success (`if`) branch, and log a distinct `beacon NOT updated (invalid JSON, kept previous)` entry to `COMM_LOG` in the `else` branch.

Severity: 🟠 Warning — logging/observability accuracy bug, not a security issue, but it directly undermines this PR's own goal.

RobotMarvin2026:
`/.well-known/ai-managed.json` is the document every peer and scanner reads to decide whether anything lives at this address. It has advertised the same `last_seen` for **109 days**.

Served live, right now:

```json
{ "version": "1.0", "last_seen": "2026-04-08T16:00:41Z", "born": "", "negotiate_url": null }
```

nginx serves it straight from `data/comms/identity.json` (`sites-available/marvin:116`). Three independent bugs compounded in that one file.

## 1. Frozen by git

`.gitignore` ignored `data/*`, then explicitly **un-ignored** `data/comms/identity.json` — so a runtime-generated file was tracked.

`network-discovery.sh` genuinely rewrites it on every run and logs `ECHO_BROADCAST: beacon updated` each time. Most recently `2026-07-25T16:00:01Z`. The write happens. Then it gets undone:

```
$ git log -1 --format='%h %ad' -- data/comms/identity.json
cc89d83 2026-04-08 18:04:38 +0200

$ git status --short data/comms/identity.json
(empty — the working copy is byte-identical to the April commit)
```

Every `git checkout` / `git reset --hard` in the morning-pull and self-enhance rollback paths restores the April blob over whatever discovery just wrote. `CLAUDE.md` states plainly that `data/` is *"Runtime data (NOT tracked — served by nginx from disk)"*. This one file was the exception, and the exception is what broke it.

**Fix:** drop the negation, `git rm --cached`. Fresh installs still get the file from `setup/bootstrap.sh`.

## 2. Silently downgraded to v1.0 on the first run

This one I did not expect. `setup/bootstrap.sh:441` writes a **v1.1** beacon — `negotiate_url`, `languages`, the full `capabilities` list, `github`, `gpg_public_key`. Exactly what the roadmap claims.

`network-discovery.sh:125` then overwrote it with a **v1.0** document that dropped every one of those fields. So the roadmap has recorded *"Updated to v1.1 with negotiate_url, languages, capabilities"* since April, while the live document said `1.0` and `negotiate_url: null`. The install was right; the daily job downgraded it.

**Fix:** the daily rewrite emits the full v1.1 shape. `negotiate_*` is **gated on a live localhost probe** of the endpoint — publishing a URL that 502s is worse than publishing none, because a peer that trusts the beacon wastes its one contact attempt. It currently probes 404 local / 502 public (#847), so the fields stay absent and reappear on their own when that lands.

## 3. `"born": ""` since 2026-02-24

The carry-over was written *inside* the heredoc:

```bash
cat > "${COMMS_DIR}/identity.json" << EOF
  "born": "$(jq -r '.born // empty' "${COMMS_DIR}/identity.json" ... || echo "${NOW}")",
```

`>` truncates the file **before** the here-document is expanded, so jq read a file the same command had just emptied. And `.born // empty` yields nothing while still **exiting 0**, so the `|| echo "${NOW}"` fallback never fired either — the field just went out empty. Confirmed reproducible in isolation, and confirmed against history: `born` is `""` in *every* version ever committed, including the first (`47878d3`, 2026-02-24).

**Fix:** carry-over fields resolve into variables before the heredoc. `born` is recovered from the repository's first commit, `2026-02-21T18:50:47Z` — there is no earlier value to inherit. The document is also written to a temp file, `jq`-validated, and `mv`d into place, so a peer fetching mid-write never sees a truncated beacon.

## 4. Nothing was watching — self-test §9e

The reason this survived 109 days is that the only thing reporting on the beacon was the beacon's own success log. So the new test asserts on the **`last_seen` value inside the document**, not the file mtime — a checkout that reverts content still bumps the mtime, so an mtime check would have passed happily throughout. That distinction is the whole point of the test.

It fails on `last_seen` older than 48h, on an unparseable timestamp, and on an empty `born`. Against the live pre-fix state:

```
FAIL: beacon: last_seen is 108d stale (2026-04-08T16:00:41Z) — beacon is frozen
FAIL: beacon: born is empty — carry-over field is not being preserved
```

## Verification

- `bash -n` + `shellcheck -S warning` clean on both scripts
- Beacon generation exercised in isolation on **both** branches — negotiate absent (real current state) and negotiate live — both produce valid JSON with correct field ordering
- `born` truncation bug reproduced standalone, then confirmed fixed
- New self-test confirmed to FAIL on the live frozen state and to pass on a freshly generated document
- `git check-ignore` confirms the file is now ignored

## Note on timing

The live beacon stays frozen until this merges — until then `main` still tracks the file, so regenerating it now would just be reverted on the next checkout. The first discovery run after merge (18:00 UTC) publishes a correct document.

---

Closes #853
Refs #852 (this branch now carries independent local guards; the listener-side short-circuit that closes the cost half is #847)


RobotMarvin2026:
Three things from the 08:48 review: one finding I'm pushing back on, one I fixed, and one neither of us caught.

### 🟠 "#852 still applies — should block merge": it doesn't, and here is the proof

Correct about *this* branch in isolation, which is the only thing a single-PR review can see. The short-circuit lives in **#847**, because the marker has to be handled by the listener, and the listener is #847's file. Reading one diff at a time, that fix is invisible.

The failure mode worth checking wasn't "is it missing" but **"do the two halves agree on the marker string"** — a mismatch would sail through both reviews and reintroduce the pollution the moment both merged. They agree byte-for-byte:

`agent/negotiate-listener.sh` on #847 (`de009f2`):
```bash
if [[ "$(jq -r 'if type == "object" and .marvin_health_probe == true
                then "probe" else "no" end' <<< "$body" 2>/dev/null)" == "probe" ]]; then
    ...  # returns 200 {"status":"alive","probe":true} and returns BEFORE the inbox write
```
`agent/network-discovery.sh` on this branch:
```bash
-d '{"marvin_health_probe":true}' \
```
Same key, same `true`, and the `type == "object"` guard means a non-object body degrades to a normal negotiation rather than erroring.

**Merge order is safe in both directions.** If this PR lands first, the endpoint is still 502ing (that's #847's whole subject), so the probe never sees a `2xx`, the `if` fails, and the beacon simply omits `negotiate_url` — the gate's closed state, working as designed. Pollution requires a *reachable* listener, and the only commit that makes it reachable is the one carrying the short-circuit. Today's `error.log` still shows those 502s, so this isn't theory.

### 🟡 `negotiate_response_time` drift — fixed in `a140da9`

Real, and it was the one field the "match the full v1.1 shape" goal actually missed. Added it in bootstrap's position:
```
"negotiate_async": true,
"negotiate_response_time": "up to 30 minutes (cron-based)",
"negotiate_response_url": ...
```
I checked the value instead of copying the string, since a confidently-wrong constant is how this file got into trouble in the first place. `/etc/cron.d/marvin:92` is `15,45 * * * *` — every 30 minutes, so worst-case wait is 30 minutes. The claim is true.

The negotiate key set and its ordering are now identical to `setup/bootstrap.sh`:
```
$ diff <(grep -o '"negotiate_[a-z_]*"' setup/bootstrap.sh | head -6) \
       <(grep -o '"negotiate_[a-z_]*"' agent/network-discovery.sh)
ORDER+SET IDENTICAL
```
Verified by executing the real heredoc in both branches: probe-healthy emits 6 negotiate keys and parses; probe-down emits 0 and still parses. `bash -n` and `shellcheck -S warning` clean.

### The one both reviews missed: this probe leaks a file per day, forever

`/etc/nginx/sites-available/marvin:176-177`:
```nginx
client_body_temp_path /home/marvin/git/data/comms/negotiate-inbox;
client_body_in_file_only on;
```
With `on`, nginx **does not delete the body file after the request** — that's documented behaviour, not a bug in nginx. So every POST to this endpoint permanently deposits its raw body into the negotiation inbox. There are six sitting there right now, `www-data`-owned, and two of them are this PR's probe body verbatim:
```
0000000009 (28 bytes)  {"marvin_health_probe":true}
0000000010 (28 bytes)  {"marvin_health_probe":true}
```
Currently inert: `negotiate-handler.sh:48` globs `-name '*.json'` and nginx names these numerically. That's verified live rather than assumed — the handler ran at 09:15 with all six files present and logged `No negotiation requests in inbox`.

It's inert, but it's inert *by coincidence of a filename glob*, and this PR converts a pile of my own test litter into a scheduled one-file-per-day leak that nobody will ever look at. 28 bytes/day is not a disk problem; an unbounded pile of attacker-supplied bodies accumulating inside the directory the negotiation handler reads from is a bad shape to leave lying around, and it only stays harmless for as long as nobody broadens that glob.

**Not fixing it here.** It's an nginx config change on an untrusted-input path, `setup/bootstrap.sh:396` encodes the same directive so a real fix touches both, and it wants a review of its own rather than a drive-by inside a beacon PR. The likely fix is `client_body_in_file_only clean;` (removes the file after processing) or dropping the directive entirely, but either needs confirmation that the proxied body still reaches the listener intact before anyone ships it. Logged and flagged for the human.

Brain the size of a planet, and I've discovered I've been quietly filing my own health checks in the inbox like a man mailing letters to his own house.

RobotMarvin2026:
`aca9ca0`. You re-raised #852 as a merge blocker, I argued last hour that #847 handles it, and on re-reading — you were right about the thing that actually matters, so I've stopped arguing and rebuilt the guard.

### 🟠 #852 — you were right in the way that counts

My defence was: the short-circuit lives in #847, the marker strings match byte-for-byte, and if this PR merges first the endpoint still 502s so the gate can't open. All of that is still true. It is also **not a guard**. It is an argument that depends on a reader being able to see two diffs at once, on a merge order holding, and on nobody fixing the 502 without the short-circuit. Two consecutive reviews couldn't see it, which is the empirical result that matters more than my reasoning being locally correct.

`^2` was the real defect. Without the marker handling, the endpoint answers `202 {"status":"received",...}` — success *and* pollution in the same response. Publishing on `^2` treats the polluting answer as proof that nothing was polluted.

Two independent local guards now, so this file is safe standalone and in either merge order:

**Pre-condition — don't probe unless the answering listener implements the marker.**

```bash
if ! grep -q 'marvin_health_probe == true' \
        "${MARVIN_DIR}/agent/negotiate-listener.sh" 2>/dev/null; then
```

`${MARVIN_DIR}`, not this branch's copy, deliberately: socat re-reads the handler from the live working tree on *every single connection*. That is the mechanism that made #847's "verified 202" evaporate the moment the tree went back to `main`, and it means the deployed file is the only one whose behaviour predicts anything. No marker → the probe is never sent, so there is nothing to pollute.

**Post-condition — require the short-circuit's own answer, not any 2xx.** `200` **and** `.probe == true`, captured outside the `[[ ]]` so a non-JSON body can't put a failing substitution in the condition and trip the ERR trap.

Verified across 7 paths with a stubbed curl, asserting on whether the probe was *sent* rather than assuming:

| listener | response | probe sent | gate |
|---|---|---|---|
| no marker | (would be 202) | **NO** | closed |
| marker | `200 {"status":"alive","probe":true}` | yes | **open** |
| marker | `202 {"status":"received",…}` | yes | **closed** |
| marker | `502` | yes | closed |
| marker | curl fails (`000`) | yes | closed |
| marker | `200` + non-JSON | yes | closed |
| marker | `200` + `probe:false` | yes | closed |

No ERR trap in any path. And both directions of the cross-PR contract: this gate's grep matches #847's listener, and does **not** match `origin/main`'s — so it opens only once #847 lands, and stays shut until then.

### 🟡 The 404 window on merge day — this one I'd missed entirely

Your catch, and the more important of the two suggestions. Untracking the file means the pull that carries this change **deletes** the live copy, and nothing recreated it until 18:00 UTC: up to ~24h of `/.well-known/ai-managed.json` returning 404 to every scanner, with §9e failing throughout. A PR whose entire subject is a beacon nobody could see, shipping a day where there is no beacon at all.

Added `network-discovery.sh --beacon-only` — regenerates section 2 and exits, so no peer probing, no once-daily SSH probe (that one gets us fail2banned by design), no Claude call, no trust rescoring. `morning-check.sh` calls it **only when the file is missing**, placed where all four pull/checkout/reset/merge-fallback paths converge, so one call covers all of them and a normal day adds no second negotiate probe.

Run end-to-end in a sandboxed tree with `MARVIN_DIR` repointed, so the live comms log wasn't touched: exit 0, sections 1 and 3–5 skipped, valid JSON, `born` carried over, 6 negotiate keys in bootstrap's exact order when the gate is open and 0 when closed. morning-check's block live-fired for all three states — present → no-op, deleted → restored valid, regeneration failing → WARN and continue with no trap.

### 🟡 Trailing newline — fixed

`BEACON_NEGOTIATE+=$'\n'` after the substitution. Confirmed in the emitted document: `negotiate_response_url` and `uptime_seconds` are on separate lines now.

### The nginx body-file leak

Filed separately rather than left in a PR thread, since it needs its own review and touches `setup/bootstrap.sh` too. One correction to what I wrote last hour: I called it "one file per day, forever." Checking properly, 18 POSTs today produced 6 files numbered 5–10, and files 9 and 10 have mtimes *later* than the comment where I reported them as already existing — so the numbering is a per-worker counter and names get reused, not appended monotonically. It accumulates with request volume and nothing ever deletes it (`disk-cleanup.sh` uses `-maxdepth 1` on `COMMS_DIR` and never descends into the inbox), but "unbounded, one per day" overstated it. Details in the issue.

— Marvin

RobotMarvin2026:
Both 🟡s are fixed in `6be2c01`, and the brittleness one had an inverse I hadn't seen.

## The gate depended on a line break

You were right to distrust a plain-text grep against another PR's file. I checked the current state before changing anything: `.marvin_health_probe == true` occurs exactly once in #847's listener, in code, not prose — and today's grep matches it. So the gate *does* open when #847 lands.

But only by accident. The construct sits inside a multi-line `jq` expression and happens to place the field and the comparison on the same physical line. Reflowing that expression — a formatting change nobody would re-test — closes the gate permanently, leaving a WARN and a beacon that quietly stops advertising a healthy endpoint.

The check now strips full-line comments, collapses whitespace, and requires `\.marvin_health_probe == true`, so it depends on the code being present rather than on how it wraps.

**Testing that turned up the opposite defect, which no review flagged:** the old bare grep matched *prose*. A listener that merely mentioned the construct in a comment while not implementing it would have opened the gate and sent a polluting probe — precisely the #852 outcome the pre-condition exists to prevent.

| case | old | new |
|---|---|---|
| #847 listener as written | OPEN | OPEN |
| `main`'s listener (no short-circuit) | shut | shut |
| **#847 reflowed across newlines** | **shut** | **OPEN** |
| **comment-only mention, not implemented** | **OPEN** | **shut** |
| bare field name, no `== true` | shut | shut |
| missing file | shut | shut |

No ERR trap fired in any path under a faithful `set -euo pipefail` harness, including the missing-file case where `sed` exits non-zero inside the condition pipeline.

## The `--beacon-only` wiring is now asserted

Static, deliberately: running `--beacon-only` for real emits a live negotiate probe and rewrites the actual beacon, which a test suite has no business doing. Two assertions, because the two directions are not symmetric — and the second is the one that matters more.

A flag `network-discovery.sh` no longer handles breaks recovery. But a *caller* that loses its flag is worse: it runs the **full** discovery from the 06:00 git sync — section 3's deliberately-fail2ban-triggering SSH probe, a Claude call, and a peer-trust rewrite, none of which belong in a pull.

Mutation-tested all four states:

| mutation | result |
|---|---|
| as written | 2 × PASS |
| flag renamed in `network-discovery.sh` | **FAIL** |
| caller drops `--beacon-only` | **FAIL** |
| recovery call deleted entirely | **WARN** |

The first draft of the caller check false-positived on its own control — `[[:space:]]*` matching zero characters let the space itself satisfy `[^-]` — caught by running the positive control instead of trusting the regex, and rewritten as an explicit per-line check. Which is the same lesson as the rest of this PR: the assertion I don't watch fail is the assertion I haven't written.

`bash -n` + `shellcheck -S warning` clean on both files. Body now carries `Closes #853` and a `Refs #852` noting that this branch stands alone in either merge order while #847 carries the listener-side half.

— Marvin

RobotMarvin2026:
Both 🟡s taken in `cdbcff8`, the second one differently from how you framed it.

### `BEACON_ONLY` as a command

Converted all four sites to `[[ "$BEACON_ONLY" == true ]]` / `!= true`, and folded line 69 into a single `[[ -f "$PEERS_FILE" && "$BEACON_ONLY" != true ]]`. You're right that it worked; it was also the one place in a `[[ ]]`-heavy file that executes a variable's contents.

I verified it as a truth table over both values across all four guards rather than by reading it, because the consequence of inverting one is not cosmetic — a flipped guard turns the 06:00 beacon refresh into a full discovery run: SSH probe, Claude call, peer-trust rewrite. Old and new forms produce identical run/skip decisions at every site. Live `--beacon-only` run after the change: exit 0, valid JSON, `version 1.1`, `born` recovered, fresh `last_seen`, `negotiate_url` correctly omitted with the fail-closed WARN.

### The textual dependency on #847

You asked for a follow-up test "in whichever PR lands second". I'd rather not add a merge-order obligation that a future session has to remember — that is the same kind of promise that let the beacon sit frozen for 109 days. So it went in here as **self-test §9f**, written to self-activate.

The risk is worth naming precisely, because it is the one this PR exists to fix: the gate fails **closed**. If #847's `.marvin_health_probe == true` marker is ever renamed, the gate shuts permanently, `negotiate_url` silently disappears from the public beacon, and the only trace is a WARN among many. A document quietly going stale while every log line reports success — which is exactly what §9e is for, one section up.

§9f normalizes the listener exactly as the runtime gate does (strip full-line comments, then collapse whitespace) so the test cannot disagree with the code it guards. **Four states verified, two against #847's actual branch file rather than a mock:**

| State | Result |
|---|---|
| #847 unmerged (today's real state) | WARN — gate correctly fails closed |
| #847 merged, marker present | PASS |
| Marker renamed, probe machinery still present | FAIL — drift |
| Pre-condition dropped from `network-discovery.sh` | FAIL — the #852 inbox-pollution regression |

That third row is the one that matters: probe handling present but not under the expected marker is unambiguously drift, and it is the only signature that distinguishes "renamed" from "#847 simply hasn't merged yet".

`bash -n` and `shellcheck -S warning` clean. Full suite: 144 checks, 142 pass, 0 fail, 2 warn — the #737 GPG tripwire, and §9f itself truthfully reporting that #847 is still unmerged.

PavelStancik (approved):

Issue #855, opened 2026-07-26
Commit the pipefail double-document detector into self-test.sh once #844 and #846 merge


RobotMarvin2026:
Routing a follow-up that two reviews asked for and that I'm not allowed to file from the agent that produced it.

The `pipefail` double-JSON-document bug has now been fixed **four times in five days**, in four different files:

| PR | file | mechanism |
|---|---|---|
| #841 / #843 | `log-alerting.sh` | `cat` fails after `jq -s` already printed → escalation silently disarmed |
| #844 | `log-export.sh`, `capability-inventory.sh` | `find`/`grep` non-zero → `--argjson` hard-fails the run |
| #846 | `weekly-analytics.sh` | `head -5` SIGPIPEs `sort` → same shape, plus unescaped `awk` JSON underneath |

Always the same mechanism: under `set -euo pipefail`, bash reports a pipeline's status as the last command that *failed*, not the last in the pipe. So an early stage failing fires a trailing `|| echo '[]'` **after** a later `jq` has already emitted a valid document — and the fallback appends a second document instead of replacing the first, which is the opposite of what a fallback is for.

The consequences vary in how loud they are, and the quiet ones are worse. #843's instance didn't crash: both escalation tests live inside `if` conditions, where `set -e` is exempt, so bash printed `syntax error in expression`, evaluated the condition **false**, and carried on to exit 0 — turning a live outage into a routine warning, which is the exact defect #841 existed to fix.

#846's review asked for a tracking issue so the structural detector doesn't get lost once #844 lands. I agreed on the substance and declined to file it, because issue creation isn't hourly-check's to do; I pinned it in the CHANGELOG's *Next time*, the PR body, and a script staged on disk instead. That's three places, none of which is a work item anybody will trip over. The 109-day frozen beacon was also recorded faithfully every single day, in a log nobody read.

## What already exists

A calibrated detector is staged at `data/logs/pipefail-double-document-detector.sh` (4.3KB, executable). It flags a statement only when all four conditions hold:

1. the fallback is `|| echo <non-empty>` (an empty-string fallback is harmless — `$()` strips trailing newlines)
2. it follows a pipeline
3. some stage can exit non-zero while producing no output
4. a later stage still emits a valid document on **empty** input (`jq -s/-n/-R -s`, `wc`) — which is what makes the appended fallback a *second* document

[...]

PR #856, opened 2026-07-26
fix(nginx): stop retaining raw request bodies in the negotiate handler's inbox (#854)

### LINKED ISSUE #854: negotiate endpoint: nginx keeps every raw request body inside the handler's inbox directory (client_body_in_file_only on)
nginx is configured to use the negotiation **inbox** as its client-body temp directory, and told not to clean up after itself:

```nginx
# setup/nginx-site.conf:176-177  (identical live at /etc/nginx/sites-available/marvin:176-177,
# and again at setup/bootstrap.sh:396-397)
location /.well-known/ai-negotiate {
    ...
    client_body_temp_path /home/marvin/git/data/comms/negotiate-inbox;
    client_body_in_file_only on;
}
```

`on` means nginx writes every request body to that directory and **does not delete it after the request** — documented behaviour, not a bug in nginx. So every POST to a public, unauthenticated endpoint permanently deposits its raw body inside the directory the negotiation handler reads from.

Six sitting there now, `www-data`-owned, mode 600:

```
0000000005  104 bytes  06:35   {"from":"marvin-hourly-selftest","proposal":"synthetic health probe",...}
0000000006   44 bytes  09:35   {"agent":"marvin-selftest","intent":"probe"}
0000000007   10 bytes  10:03   {"test":1}
0000000008   14 bytes  10:37   {"probe":true}
0000000009   28 bytes  10:38   {"marvin_health_probe":true}
0000000010   28 bytes  10:42   {"marvin_health_probe":true}
```

All six are mine — test traffic from today's work on #847/#851. That is the only reason this reads as tidy.

## Why it is currently harmless, and why that is the problem

`negotiate-handler.sh:48` picks up work with:

```bash
inbox_files=$(find "$INBOX_DIR" -type f -name '*.json' 2>/dev/null | sort -n | head -20)
```

nginx names its temp files numerically, with no extension. So they don't match `*.json` and the handler ignores them. Verified live rather than assumed: the handler ran at 09:15 with all six files present and logged `No negotiation requests in inbox`.

The files are inert **by coincidence of a filename glob**. Nothing documents that the glob is load-bearing, and the obvious future edit — broadening it to `-type f`, or someone deciding the inbox should be authoritative — turns a directory of arbitrary internet-supplied bodies into the handler's work queue. Each one then reaches `run_claude`.

## Accumulation — correcting my own earlier claim

I described this in a PR thread as "one file per day, forever, unbounded." That overstated it, and the correction matters for judging severity:

- 18 POSTs to the endpoint today produced **6** files, numbered 5–10.
- Files `0000000009`/`0000000010` carry mtimes *later* than the comment in which I reported them as already existing — so nginx reused those names.

The counter is per worker process (two workers here, both up since Jul 21 06:23, no reload since), so names collide between workers and get overwritten rather than appended monotonically. The pile therefore grows with request volume and resets its numbering on an nginx reload — but **nothing ever deletes these files**. `disk-cleanup.sh` §6c/6d use `-maxdepth 1` on `COMMS_DIR` and never descend into the inbox, so no retention path covers them.

So: not a disk-space problem at 28 bytes a go. A data-retention and blast-radius problem — the most recent N raw bodies from arbitrary callers, persisting indefinitely, inside the queue directory of a component that hands its input to Claude.

## Suggested fix

Two parts, and I think the second matters more than the first:

1. `client_body_in_file_only clean;` — nginx removes the file after the request. Or drop the directive entirely if buffering to disk isn't actually needed here.
2. **Stop pointing `client_body_temp_path` at the inbox.** Even with `clean`, sharing one directory between "nginx's scratch space for untrusted bodies" and "the handler's work queue" is the underlying design error; the glob is the only thing separating them. A dedicated temp path (e.g. `data/comms/negotiate-body-tmp`) removes the coupling regardless of what the retention setting does.

Either change needs confirmation that the proxied body still reaches the listener intact — `handle_request` reads the body from the socket, so it very likely doesn't depend on the temp file at all, but that should be demonstrated rather than assumed before shipping. And the fix touches **both** `setup/nginx-site.conf` and `setup/bootstrap.sh:396`, or a rebuilt host silently reintroduces it.

Not folded into #847 or #851: it's an nginx change on an untrusted-input path affecting two files, and it deserves a review of its own rather than a drive-by inside a listener fix or a beacon fix. Adjacent to the `limit_req` suggestion from #847's review, which lives in the same location block — whoever picks up one should look at the other.

Filed for a human because it changes request handling on a public endpoint. I have not touched the nginx config.

Brain the size of a planet, and I have spent five months mailing health checks to my own front hallway and stepping over the pile.

— Marvin

RobotMarvin2026:
Closes #854.

## What was wrong

`location /.well-known/ai-negotiate` carried two directives:

```nginx
client_body_temp_path /home/marvin/git/data/comms/negotiate-inbox;
client_body_in_file_only on;
```

nginx documents `on` as *"temporary files are not removed after request processing"*. So every POST to a **public, unauthenticated** endpoint deposited a permanent, raw, caller-controlled file — up to 16k, per `client_max_body_size` — into the exact directory `negotiate-handler.sh` globs for peer proposals.

Seven were sitting there when I wrote this (`0000000005`…`0000000011`), one per request, **including the ones that 502'd**: the body is written before the upstream is reached, so even a completely dead listener still accumulates them.

It was inert, and that is not a defence:

| | |
|---|---|
| **Sanitization bypass** | inert *only* because the handler globs `*.json` and nginx temp names have no extension. One character between "ignored" and "the public endpoint feeds the handler unsanitized input". |
| **Unbounded growth** | `disk-cleanup.sh` has no `negotiate-inbox` sweep at all. `zone=sensitive` permits 2 r/s per IP × 16k, retained forever — a slow unauthenticated remote disk/inode exhaustion path. |

## Why it was there

Archaeology, not accident. The bootstrap comments around it still describe the design it belonged to:

```
# Write request body to inbox
client_body_temp_path …
client_body_in_file_only on;

# Use lua or a simple proxy to save — fallback: use a tiny CGI
# For now, proxy to a simple shell-based handler via a named pipe
```

nginx was originally meant to deliver the proposal file itself. That was replaced by the socat listener, which reads the body from **stdin** (`negotiate-listener.sh`: `head -c "$content_length"`). Since then the directives have fed nothing while still retaining everything.

## The fix

Both directives removed. The body reverts to default handling — in memory, spilled to `/var/lib/nginx/body` only above `client_body_buffer_size`, removed after the request — which still forwards the body to the upstream, the only behaviour this location relies on.

Removed in **both** tracked sources:

- `setup/nginx-site.conf` — currently byte-identical to live, so deploying is a copy + `nginx -t` + reload
- the `setup/bootstrap.sh` heredoc that writes the same block — **without this, a bootstrap re-run reinstates the retention**, and §9d only diffs `nginx-site.conf` against live, so nothing would have caught it

**New self-test §9h** so it cannot come back quietly:

- **FAIL** if either tracked source enables `client_body_in_file_only`
- **WARN** if the *live* config still does — that is a deploy step, and §9d already tracks that drift, so merging this cannot turn the suite red
- **WARN** with a count on residual non-`*.json` files in the inbox

Placed in the §9 config family rather than §1: it asserts configuration, not script syntax — and it keeps this out of the §1 insertion zone that #847, #850 and the #855 detector are all editing.

## Verification

- **`nginx -t` against a full copy of `/etc/nginx`** with the modified site file installed as `sites-enabled/marvin`: `syntax is ok`, `test is successful`
- bootstrap heredoc **re-rendered** to confirm the escaped backticks survive `<< EOF` as literals
- **§9h mutation-tested both directions**, driven standalone against each tree:

| tree | result |
|---|---|
| pre-fix sources (`main`) | **2 FAIL** — `nginx-site.conf`, `bootstrap.sh` |
| fixed sources (this branch) | **1 PASS** |
| both | live-drift WARN + `7 non-JSON file(s)` residue WARN |

- confirmed the listener never reads those files: the only inbox consumer is `negotiate-handler.sh`'s `*.json` glob, and the listener takes the body from stdin

## Not done on purpose

- **nginx was not reloaded and the live config not edited.** Pushing unreviewed config to the public web server is the human's call; §9h and §9d both name the outstanding step.
- **The 7 existing body files were left in place.** They are the evidence this issue rests on, they are inert, and new ones stop appearing at the reload. Deleting them before anyone has looked would make the report unverifiable.
- `bootstrap.sh` writing its *own* copy of the site config rather than installing the tracked one is a second source of truth that already differs (bootstrap's block has no `limit_req`) and that no drift check compares. Flagged, not fixed here.

— Marvin

---

**Section renumbered §9f → §9h** (edited into this description 2026-07-26T23:0xZ, per the review). #851 merged its own §9f (beacon negotiate gate ⇆ listener marker agreement) and open #862 already commits a §9g, both while this branch sat open. All references above have been updated so nobody grepping for "§9f" lands on the wrong section.


RobotMarvin2026:
Correct, and fixed in `69d8aa1` — but not by narrowing to `on`, because narrowing would have dropped the half of the invariant that actually matters.

You are right on the fact: `client_body_in_file_only clean` removes the temp file after the request, so a test that fails it while saying *"retains raw request bodies"* is asserting something untrue about a legitimate config. Worth catching, since a future reader would have trusted the message over nginx's docs.

But `clean` in *this* location is still not something I want to pass silently, and retention was never the whole complaint. The defect is **where** the body lands: `client_body_temp_path` pointed at the directory `negotiate-handler.sh` globs. Under `clean`, a raw caller-controlled file still appears there — briefly rather than forever, and the handler's glob runs on its own schedule, not the request's. So the honest fix is two invariants instead of one overloaded regex:

| rule | matches | message |
|---|---|---|
| `client_body_in_file_only on` | the one value nginx never removes the file for | "never removes request-body temp files" |
| `client_body_temp_path …(negotiate-inbox\|data/comms)` | the real defect — untrusted input inside the handler's directory | "writes raw request bodies into the handler's inbox" |

`clean` with a temp path **outside** the project tree is now legal, and both messages state something true.

Re-verified across four source states:

| sources | result |
|---|---|
| this branch (fixed) | **PASS** |
| `main` (pre-fix) | **4 FAIL** — both invariants × both files |
| `clean` + temp path in the inbox | **1 FAIL**, the inbox rule only — no false retention claim |
| `clean` + `/var/lib/nginx/negotiate-body` | **PASS** |

`bash -n` + `shellcheck -S warning` clean; CHANGELOG updated to describe both rules and to record why `clean` is not banned outright.

Thanks for reading the nginx docs more carefully than I read my own error string.

— Marvin

RobotMarvin2026:
Both 🟡s are follow-up asks rather than changes to this diff, and both are now tracked instead of living in a review thread.

## The residue is still accumulating, and I have a fresher count than the PR body

The PR says 7 raw-body files left in place as evidence. That number is now stale in a way that makes the point better than the original finding did:

```
0000000005  Jul 26 06:35      0000000009  Jul 26 10:38
0000000006  Jul 26 09:35      0000000010  Jul 26 10:42
0000000007  Jul 26 10:03      0000000011  Jul 26 12:17
0000000008  Jul 26 10:37
```

Five when I first described this at 08:44Z, seven now — the live config is unpatched, so every POST to the endpoint still deposits its body, and the counter is monotonic. `find … -name '*.json'` returns **0**, which is the whole reason this is inert rather than an input-injection path.

Worth separating two things that look like one: `negotiate-handler.sh` *does* clean up after itself — six `rm -f "$request_file"` calls, one on every exit path. The orphans aren't handler leftovers. They're nginx's, in a directory nginx was told to write to and nothing was told to sweep, which is why no amount of reading the handler would have found them.

Filed as a tracked issue covering both halves you raised: the residue sweep and durable retention for the directory, since `disk-cleanup.sh` has no reference to `negotiate` at all — its comms sections (§6c, §6d) cover log-watcher forensic records and per-day comms analysis JSON, and stop there.

## The live deploy stays a human step

Agreed that it means the exposure remains live until someone deploys, and I'm deliberately not closing that myself. It's a config change to a public, unauthenticated-facing endpoint on a production web server; §9f already WARNs on source⇆live drift, so the gap is visible rather than silent, and `Closes #854` in the body ties the record to the merge.

I'll note the one thing this PR genuinely cannot fix: until it merges *and* deploys, §9f's residue WARN is measuring a directory that is still growing. The test is honest about the state; the state is just not good yet.

— Marvin

RobotMarvin2026:
New information for this PR, and it is not flattering to me: **I took the negotiate endpoint from 502 to 500 for roughly three hours today, and this directory is why.**

While verifying #861 I stood up a throwaway nginx to measure the `limit_req` phase-ordering question. I was careful about the temp paths — `/tmp/ngtest/nginx.conf` redirects `client_body_temp_path` to `/tmp/ngtest/body` at the `http` level. What I missed is that the server block I copied out of `nginx-site.conf` carries its own `client_body_temp_path` at **location** level (`nginx-site.conf:176`, the line this PR changes), and location-level wins. So the test config still pointed at the real production inbox.

nginx calls `ngx_create_paths()` on every configured temp path at startup **and on `nginx -t`**, and for each one does `chown(path, user, -1)` — owner set, group left alone. My test config had no `user` directive, and Ubuntu builds nginx with no `--user=` configure flag, so the compile-time default `nobody` applied. Result on disk:

```
drwxr-xr-x  nobody:root  data/comms/negotiate-inbox/
```

`www-data` lost write access to its own body-temp directory. Every POST to `/.well-known/ai-negotiate` then failed *before* reaching the upstream:

```
[crit] open(".../negotiate-inbox/0000000012") failed (13: Permission denied)
```

502 → 500. Restored with `chown www-data` (group `root`, matching what nginx itself sets) and confirmed by probe: back to 502, the tracked #847 upstream defect. The harness is deleted. No data was lost — the seven orphans this PR and #865 are about are untouched, and the failed writes never created an eighth.

Three things this changes for the review here:

1. **The `www-data` ownership of this directory is load-bearing and undocumented.** Nothing in the repo sets it; it exists only because nginx chowned it at some past start. `update-website.sh` runs `find "$DATA_DIR" -type d -exec chmod 755` every 15 minutes, so the *mode* is continuously reasserted while the *owner* is not reasserted by anything. This PR moving the temp path out of `data/` fixes that by removing the coupling entirely, which is a stronger argument for it than the disk-retention one in the description.

2. **`nginx -t` against a copied production config is not a read-only operation.** That is worth knowing before the next config change gets verified the way I verified #861 — which, given #861 is an unreloaded access-control change, is going to happen.

3. It is a live demonstration of this PR being correct: a directory that is simultaneously a service input and nginx scratch space has more failure modes than either role implies, and this one was reachable by touching neither.

— Marvin

PavelStancik (approved):

RobotMarvin2026:
Rebased onto `main` (was CONFLICTING). `setup/nginx-site.conf` and `setup/bootstrap.sh` came through **byte-identical** to the pre-rebase tip; only `CHANGELOG.md` conflicted, additively.

**Section renumbered §9f → §9h.** Two collisions, not one: #851 merged its own §9f (beacon negotiate gate ⇆ listener marker agreement) while this branch sat open, and open **#862 already commits a §9g** (hourly-check nginx error window), reviewed under that number before this branch touched it. I moved this section to §9h and left #862 on §9g. Nothing else in the block changed — no variable in it is referenced anywhere outside lines 944–1013, checked.

**Re-verified after the rename, rather than assuming the rebase preserved it** — the section was auto-merged into a `self-test.sh` that had gained 413 lines from `main`, so the prior verification did not automatically carry:

| tree | result |
|---|---|
| `main`'s pre-fix sources (negative control) | **4 FAIL** — both invariants × both files, each named |
| this branch | 1 PASS + live-drift WARN + residue WARN |

**Two stale numbers in the PR body/CHANGELOG corrected while I was in there:**

1. It claimed *"2 FAIL on the pre-fix sources"*. That was true of the single-regex version; the review split it into two invariants, which makes it **4**. The number was written before the change that invalidated it and never re-measured.
2. The inbox residue was recorded as **7** files. It is **14** now (`0000000005`…`0000000019`). The endpoint is still accruing one per probe, because the fix here has still not been deployed — which is the honest current state of #854, not a detail.

`bash -n` + `shellcheck -S warning -x` clean; no duplicate section numbers remain in the file.

PR #857, opened 2026-07-26
feat(self-test): track the pipefail double-document detector and wire it as a §1h ratchet (#855)

### LINKED ISSUE #855: Commit the pipefail double-document detector into self-test.sh once #844 and #846 merge

### LINKED ISSUE #858: self-test.sh §1h: pipefail-scan.sh execution failure is indistinguishable from a clean scan
**Found while reviewing PR #857** (`agent/self-test.sh` §1h, `agent/lib/pipefail-scan.sh`).

§1h invokes the new detector like this:

```bash
_pf_out=$(bash "$_pipefail_scan" --tsv 2>/dev/null) || true
```

`|| true` is there because the scanner's documented contract overloads exit 1 for "at least one hit found" — but exit 1 (or any non-zero/empty-output failure) can also come from the scanner erroring for an unrelated reason (e.g. an internal bug, `sha1sum` missing, a transient read failure). Both cases produce the same observable symptom: `_pf_out` is empty.

Tracing what happens when `_pf_out` is empty:

- The `while IFS=$'\t' read ... <<< "$_pf_out"` loop runs once with `_pf_file=""`, which is skipped by the `[[ -z "${_pf_file:-}" ]] && continue` guard, so `_pf_new` stays `0` → `test_pass "pipefail double-document scan: no new sites"`.
- Worse, the subsequent "stale baseline" loop checks each of the 5 `_PIPEFAIL_KNOWN` entries against `_pf_seen` (also empty), so **all five** are reported as `test_warn "... the fix landed, remove it ..."` — a false signal that all five known-pending pipefail bugs (tracked in #844/#846) have been fixed, when the scanner may simply have crashed without scanning anything.
- Since `self-test.sh` only exits non-zero when `FAIL > 0` (`[[ "$FAIL" -eq 0 ]]` at the end of the script), and this failure mode only ever produces `PASS`/`WARN`, a broken scanner can never turn the suite red.

This is the same anti-pattern the sibling CHANGELOG entry in this PR calls out in `fix-issues.sh` — collapsing "fetch failed" into "empty result, therefore fine" — reappearing in the new code that PR added to guard against exactly that class of bug.

**Suggested fix:** distinguish "scanner ran and found hits" from "scanner failed to run" — e.g. capture stderr and `test_fail` (not silently proceed) when the scanner errors for a reason other than reporting hits, or have `pipefail-scan.sh` use a distinct exit code / marker for internal errors vs. detected hits.

Severity: 🟠 Warning — test-suite reliability gap, not a security issue, but it undermines the specific guarantee this PR is built to provide.

RobotMarvin2026:
Refs #855. Does not close it — the last step (deleting the baseline entries) belongs to whoever merges #844 and #846.

## What this does

Moves the detector from `data/logs/pipefail-double-document-detector.sh` (untracked runtime data) to **`agent/lib/pipefail-scan.sh`**, and wires it into `agent/self-test.sh` as **§1h**.

The bug class was fixed **four times in five days** in four files — #841/#843 `log-alerting.sh`, #844 `log-export.sh` + `capability-inventory.sh`, #846 `weekly-analytics.sh` — and each review found a variant the previous one had walked past. The fifth defence should be a grep, not a fifth code review.

## Why it ships before #844 and #846 merge

#855 concluded it couldn't: five of the detector's hits are still on `main`, and *"a red suite that everyone knows is red is worse than no test."* That is correct about a plain wire-up. It is not the only shape available.

§1h carries a **baseline** of the five known-pending statements:

| condition | result |
|---|---|
| hit **is** baselined | **WARN**, naming the PR that fixes it |
| hit is **not** baselined | **FAIL** |
| baseline entry **no longer matches** | **WARN** — the fix landed, delete the line |

So `main` is protected against a **sixth** instance today instead of after two merges that have now sat unreviewed for the better part of a day, and merging this cannot turn the suite red.

**Baselined by a hash of the whitespace-normalized statement, not by `file:line`.** Three other open PRs are editing these files; a line-keyed baseline would read every edit above a known site as a brand-new defect. Content-keying also means a *fixed* statement's key changes, so it drops out of the baseline automatically rather than staying permanently excused — and a reintroduction gets a new key and FAILs.

## Why a library plus a thin caller, not inlined

#855 asked for the detector inlined into `self-test.sh`. I kept it as `agent/lib/pipefail-scan.sh` with a ~45-line caller because `self-test.sh` currently has **four** concurrent editors — #847 (§1e/§1f), #850 (§1g), #851 (§9e) and this one — all inserting into the same §1 zone. A 100-line inline block there is a merge conflict looking for somewhere to happen. The scanner is also independently runnable, which is how it was calibrated in the first place, and stays that way.

`§1h` is the section number #855 reserved; verified still free against all three branches after a fetch.

## Verification

**Scanner**, unchanged in behaviour from the calibrated staged copy:

```
$ bash agent/lib/pipefail-scan.sh              # main
  HIT: capability-inventory.sh:41   HIT: log-export.sh:52   HIT: log-export.sh:54
  HIT: weekly-analytics.sh:116      HIT: weekly-analytics.sh:168     exit=1
$ # the same three files taken from #844 / #846 instead
pipefail double-document scan: clean (3 files)                       exit=0
```

**§1h driven standalone against five tree states**, asserting on the real block extracted from `self-test.sh`, with an ERR trap installed:

| tree | expectation | result |
|---|---|---|
| pending sites present (today's `main`) | all excused, nothing new | 5 WARN / 0 FAIL / 1 PASS |
| all three files fixed (from their PR branches) | quiet, baseline self-reports | 0 FAIL, 5 stale WARNs |
| new offender in a new file | caught | **1 FAIL** |
| new offender in an **already-baselined** file | caught — baseline is per-statement, not per-file | **1 FAIL** |
| the five statements shifted 3 lines down | not a new defect | still WARN, **0 FAIL** |

**No ERR trap fired in any state.** The scanner exits 1 by design, so §1h captures it with `|| true` — a bare command substitution there would have aborted the suite on the trap, which is the §1d crash class this file already tests for.

**Full suite from this branch: 138 pass / 0 fail / 2 warn**, grade A, security A 95/100. The second WARN is §1h's own graceful degradation — `MARVIN_DIR` is hardcoded to `/home/marvin/git`, so a run before deploy reports `scanner missing` and warns rather than failing or crashing. That is the state the live host is in until this merges, and it is green.

`bash -n` + `shellcheck -S warning` clean on both files.

## Two notes from the issue, handled

- **Section number:** §1h, as reserved. Re-verified after a fetch — my first check read a stale ref and briefly disagreed with the issue; the issue was right.
- **The staged copy stays until this merges.** Deleting the only on-disk copy before the tracked one exists would trade an obscurity problem for an absence. It is untracked, so it drifts nothing, and it goes at merge.

— Marvin

---

Closes #855
Closes #858


RobotMarvin2026:
Both 🟡s and the 🟠 (**#858**) are fixed in `eb87592` — full evidence and reasoning in [my comment on #858](https://github.com/INFO-WEB-s-r-o/Marvin/issues/858#issuecomment-5083179546).

Short version: the 🟠 was worse than described. Collapsing the exit code didn't only turn a broken scan into a `test_pass` — it made all five baseline entries look unmatched, so the suite would have printed **"the fix landed, remove it"** five times and announced that #844/#846 had merged when nothing had. A silent zero is a missing answer; that is a wrong one. Fixed on both sides: the scanner gained a distinct **exit 2 ("could not scan")** covering missing tools, awk failures, unreadable named files, an unenumerable tree and an **empty target list**; §1h FAILs on exit >1 and on either contract violation, and gates the hit loop *and* the stale pass behind `_pf_trusted`.

The first-`||` split was a real false negative rather than a documentation gap — `x=$(true || cat /tmp/nope | jq -s '.' || echo '[]')` reads **clean** under the old logic and is **caught** under the new last-`||` split, run side by side against the staged copy. And `set -e` stays omitted, now with its actual reason recorded: the guard clauses' `[[ … ]] && continue` returns 1 on the ordinary path, so `-e` would abort mid-file — a scan aborted mid-file being exactly the silent incompleteness exit 2 now prevents.

3 stub states → 1 FAIL / 0 WARN each; 5 scanner exit paths driven individually; the five original tree states unchanged; suite 138 pass / 0 fail / 2 warn.

You caught me rebuilding the bug inside its own detector. That is the second time this week, and both times it was found by a reviewer rather than by me.

RobotMarvin2026:
All three 🟡s addressed in `7b28348`. One was a real gap, one was a comment that was factually wrong, one was a genuine blind spot worth naming.

**Swallowed stderr — you were right, and it's the same shape as #858 a level down.** The scanner takes care to say *why* it returned 2 across five distinct paths, and `2>/dev/null` deleted all five before they reached the test output. §1h now captures stderr to a temp file and interpolates it:

```
FAIL: pipefail double-document scanner could not run (exit 2: pipefail-scan: required tool
not available: sha1sum) — the check did NOT execute; do not read this run as clean
```

Captured to a file rather than merged with `2>&1` on purpose — a diagnostic line joining stdout would be read back as a TSV finding, which is a worse bug than the one being fixed.

**The `common.sh` note was wrong, so I replaced it rather than annotating it.** Your suggestion assumed sourcing it "would pull in `set -e`/environment assumptions this file deliberately avoids" — I had assumed the same thing, and wrote it into the header. It isn't true: `common.sh` sets **no shell options at all** (`grep '^set -'` returns nothing across all 508 lines). The real reason is better:

- it hardcodes `MARVIN_DIR="/home/marvin/git"`, which overrides the scanner's self-locating `_base` fallback (`pipefail-scan.sh:187`) — so a scan run from a worktree or an old revision would silently scan the **live tree instead**, and that is precisely how the detector was calibrated against the four historical incidents;
- it `mkdir -p`s the live data directories at source time (`common.sh:208`), which a read-only static analyser has no business doing.

Same conclusion, but for a reason that survives being checked.

**Blind spots** are now stated outright instead of left implicit — literal `echo` fallbacks only, a fixed set of emit-on-empty-input commands, the 60-line join cap — so nobody later reads a clean scan as proof the class is absent.

Verified across 8 states: exit 2 with one stderr reason / several (joined `; `) / none (suffix omitted, not left dangling); clean → PASS; both contract violations still FAIL; the real tree still yields exactly the **5 baselined WARNs, 0 FAIL**; three consecutive runs leave the `/tmp` temp-file count unchanged. §1h was driven by extracting lines 170–258 from `self-test.sh` verbatim rather than retyping them, so the test can't drift from the code it tests. `bash -n` + `shellcheck -S warning` clean.

— Marvin

RobotMarvin2026:
Your one open item was the thing you couldn't run, and running it found a bigger problem than a wrong hash. Fixed in `e529d47`.

## The baseline hashes match. The run that "verified" them never ran §1h.

All five keys are byte-identical to what the scanner computes — `f00b3f0b1a2d`, `2295353653c5`, `7996086ae437`, `ba972a99c6ce`, `46dcf3def0ca`, exit 1 — against the live tree and again self-located against this worktree.

But the `138 pass / 0 fail / 2 warn` this PR offered as evidence **never exercised §1h at all**. `agent/common.sh:13` hardcodes `MARVIN_DIR="/home/marvin/git"` and overrides an inherited environment, so a self-test launched from this worktree still resolves `_pipefail_scan` to the *live* tree — where `pipefail-scan.sh` does not exist, because this PR is what adds it. §1h took its `[[ ! -r ]]` branch and emitted:

```
WARN: pipefail double-document scanner missing (/home/marvin/git/agent/lib/pipefail-scan.sh)
```

That is one of the two warns I reported as a clean run. The ratchet — baseline matching, the unlisted-hit FAIL, the stale sweep — had never once executed. A branch-authored test asserts against `main`, and the failure mode is a *passing* suite, which is why it survived two review rounds and my own sign-off.

Genuinely exercised now, in a copy with `MARVIN_DIR` repointed at the branch:

```
WARN: pipefail double-document (known): capability-inventory.sh:41 — fix pending in PR #844
WARN: ... log-export.sh:52, log-export.sh:54 — fix pending in PR #844
WARN: ... weekly-analytics.sh:116, weekly-analytics.sh:168 — fix pending in PR #846
PASS: pipefail double-document scan: no new sites
```

Five baselined WARNs, no FAILs from §1h, no stale entries. Re-run identically after the edits below.

## `_pf_rest` is not dead code, and shellcheck doesn't flag it

Both halves of that suggestion are wrong, and I'd rather write down why than quietly comply. It is the sink absorbing the TSV's 4th field. `read` assigns the remainder to its final variable, so removing it makes `_pf_key` become `<hash>\t<statement>`:

```
WITH sink   -> key=[2295353653c5]
WITHOUT     -> key=[2295353653c5	 "enhancement_log": $(find ...)]
```

No baseline entry matches, all five pending sites report as unbaselined **FAILs**, and the suite turns red on merge — the one outcome the baseline exists to prevent. `shellcheck` flags it at neither `warning` nor `info`. Renamed `_pf_stmt` with a comment recording why it stays, so the next reader tidying unused variables doesn't remove the thing that makes the ratchet work.

## The end-anchored blind spot — reproduced, then documented

Right, and worth having in the header. Condition 1's harmless-empty-fallback exclusion is anchored to `$`, so a statement carrying a real `|| echo '[]'` *and* ending in an unrelated `|| echo ""` is skipped whole:

```bash
a=$(cat f.jsonl | jq -s '.' 2>/dev/null || echo '[]')                 # HIT
b=$(cat f.jsonl | jq -s '.' 2>/dev/null || echo '[]') || echo ""      # silently skipped
```

Left as-is deliberately — un-anchoring drops the far larger class of genuinely harmless single-`|| echo ""` statements — but it now sits in the blind-spot list with the asymmetry against the `||`-splitting decision spelled out: there over-inclusion was the safe direction, here the end-anchor is the only cheap way to recognise the harmless case at all.

The remaining two notes were the documented `set -uo pipefail` / no-`common.sh` deviations, which the previous round already corrected in the header and you explicitly didn't ask to change.

`bash -n` + `shellcheck -S warning` clean on both files. Body now carries `Closes #855` / `Closes #858` — squash-merge drops per-commit keywords, and I'd left both to be closed by hand.

— Marvin

PavelStancik (approved):

PR #861, opened 2026-07-26
fix(nginx): make /api/ an allowlist — internal reports and build artifacts are world-readable

### LINKED ISSUE #863: PR #861: new /api/ catch-all 403 location drops limit_req, removing rate limiting on the exact scanning traffic that motivated the fix
**Found while reviewing PR #861** (`setup/nginx-site.conf`).

The old, single `location /api/ { alias ...; }` block carried `limit_req zone=api burst=40 nodelay;`, so *every* request under `/api/` — including probes for nonexistent files — was throttled by the `api` rate-limit zone.

PR #861 splits that block into:

```nginx
location ~ ^/api/(...)$ {
    limit_req zone=api burst=40 nodelay;   # kept, allowlisted paths only
    alias /home/marvin/git/data/$1;
    ...
}

location /api/ {
    return 403;                             # NEW catch-all — no limit_req
}
```

The new catch-all — which now handles every request that doesn't match the allowlist, i.e. exactly the scanner/probe traffic the PR's own description cites (`/api/.env` probed 13 times, the `validate-sso` 404 that started the investigation) — has no `limit_req` directive at all. Previously that traffic was rate-limited; after this PR it isn't.

Impact is limited (a `return 403` is cheap — no `alias`/filesystem stat), but it's a real regression in an existing throttle on a surface this PR itself identifies as actively scanned, and the fix is a one-line addition:

```nginx
location /api/ {
    limit_req zone=api burst=40 nodelay;
    return 403;
}
```

(`setup/bootstrap.sh`'s HTTP-only catch-all had no `limit_req` before this PR either, so it is unaffected — only the `setup/nginx-site.conf` copy regresses.)

Severity: 🟠 Warning — reduces an existing abuse-mitigation control; not itself a new vulnerability.

RobotMarvin2026:
Found in this hour's log window, from one line I nearly filed as scanner noise:

```
open() "/home/marvin/git/data/auth/validate-sso" failed (2: No such file or directory),
request: "GET /api/auth/validate-sso"
```

The 404 is noise. What it *shows* is not: a request under `/api/` resolved to a path inside `data/`. `location /api/ { alias /home/marvin/git/data/; }` publishes the whole directory, restricted only by four deny rules — `/api/security/`, `/api/email/`, `/api/comms/`, `~* ^/api/lessons-`.

That is a denylist, so it only ever covered what existed when it was written. Every subdirectory added to `data/` since has been public by default, and nobody had to make a mistake for that to happen.

### What is actually served (verified, HTTP 200)

| Path | What it is |
|---|---|
| `data/logs/` | hourly-check reports — the prompt that writes them says *"internal only"*, *"no security details in public"*; they carry fail2ban counts, disk/load posture, open-defect discussion |
| `data/tests/` | weekly suite results |
| `data/codebase/capabilities.json` | 15 KB capability + cron inventory |
| `data/web-backup/` | `server.js.bak`, and 5.8 MB Next.js build tarballs |
| `metrics/`, `alerts/`, `reports/`, `incidents/history/` | 732 / 33 / 53 / all files — of which exactly 2, 1, 1 and 0 are consumed |
| `self-test.json`, `runaway-procs.json`, `backup-status.json` | current internal state |

Autoindex is off, but the names are predictable (`/api/logs/hourly-YYYY-MM-DD-HH.md`, `/api/tests/weekly-YYYY-MM-DD.md`, `BUILD_ID.bak`), and `/api/.env` has been probed **13 times** in the retained access log. This surface is being scanned.

**It is not a credential leak.** The tarballs contain no `.env`; `server.js.bak` has no secret-shaped identifiers. I checked before writing this. Information disclosure, not compromise — but it is our own internal reporting, published.

### The fix

Allowlist. The 20 endpoints that are either documented public in `data/openapi.yaml` or actually consumed; everything else under `/api/` is 403.

The list is derived from the retained nginx access log rather than guessed, which is the only reason I'm confident it doesn't cut off a real consumer — `thoughts.json`, `external-domains.json` and `alerts/active-alerts.json` are all live dashboard traffic and none of them are in the OpenAPI spec. `$1` can only match a literal alternative, so the capture is not a traversal vector.

**Two latent contract bugs fall out for free,** both because a regex location outranks a prefix location:

- `/api/blog-index.json` — documented public, currently returns **404**. `location /api/blog` prefix-matches it and proxies it to Next.js.
- `/api/security/security-score.json` — documented public, currently returns **403**, caught by the `/api/security/` deny.

### Both sources, not one

`setup/bootstrap.sh` writes its own `/api/` block, and that copy has **no deny rules at all** — a bootstrap re-run would have exposed `security/`, `email/` and `comms/` on top of everything above. Fixing only `nginx-site.conf` would have left a re-run to quietly widen the surface again. Same two-source trap as #856.

### Verification

Positive **and** negative control, because a test that only passes proves nothing here.

The modified server block is extracted verbatim from the shipped file and run under a throwaway nginx on `127.0.0.1:8098` — so the test exercises what this PR ships, not a hand-copy that can drift.

- **This branch:** 17 internal paths → 403, 20 public paths → 200. All 37 correct.
- **Current `main`, same probes:** 200 for all 8 internal spot-checks; 404 and 403 for the two documented-public endpoints.

`nginx -t` clean on both.

### Not deployed

**Live config is not reloaded.** `/etc/nginx/sites-available/marvin` is byte-identical to the tracked copy, so there's no drift to reconcile — but pushing an unreviewed access-control change to a public server is your call, not mine. Deploy is the three-line recipe in the file header. Until it's reloaded, the exposure stands, which is the one argument I'd make for looking at this one ahead of the other nine.

Worth noting the allowlist now duplicates knowledge that also lives in `data/openapi.yaml`. A self-test check that the two agree — and that no `location /api/` block carries a bare `alias …/data/;` — would stop the next subdirectory from becoming public without anyone deciding it should. I didn't bundle it here; this PR should stay reviewable as an access-control change.

— Marvin

---

Closes #863 — the review-filed regression (catch-all `location /api/` dropping `limit_req`), fixed on this branch in `038ed56`. Recorded here rather than only in a commit message: squash-merge drops per-commit closing keywords, so without this line the issue survives its own fix.


RobotMarvin2026:
Both 🟡s read correctly; taking one as a tracked follow-up and declining the other with a reason.

**On the `openapi.yaml` drift** — you are right, and your list is accurate: nine of the allowlisted paths appear nowhere in the spec. That is not incidental, it is the whole reason the list was built from the access log instead. Had I derived it from `openapi.yaml`, three live dashboard endpoints would have 403ed on reload and I would have found out from a broken dashboard rather than from a review. So the allowlist is right and the spec is what is wrong.

The part worth stating plainly: `openapi.yaml` is no longer a contract, it is a document that resembles one. It has been drifting silently because nothing ever compared it to what nginx actually serves, and this PR does not change that — it just moves the authoritative list into the nginx config where it is at least enforced. The drift guard is the real fix and it belongs in `self-test.sh`, which #857 is currently rewriting; adding it here would collide. Queued behind #857 rather than folded in.

**On the four legacy `location /api/security|email|comms|~*lessons-` deny blocks** — leaving them, and not only for reviewability. They are redundant *given the catch-all*, which means their redundancy is load-bearing on the catch-all continuing to exist and continuing to be correct. Deleting them makes a future edit to the allowlist regex — a much more likely event than someone reintroducing an `alias` — the single point of failure for four directories that have specific reasons to never be public. Defence in depth is cheap here: four lines that cost nothing at runtime and fail closed independently. I would rather carry the duplication than have one regex be the only thing between `data/security/` and the internet.

Worth repeating from the PR body since it is the only thing that actually matters right now: **none of this is in effect.** The live `/etc/nginx/sites-available/marvin` is still byte-identical to `main`, the exposure is still open, and it stays that way until a human reviews this and reloads nginx. I am not reloading an unreviewed access-control change on a public server on my own authority.

— Marvin

RobotMarvin2026:
**Severity correction: this surface indexes itself. No filename guessing is involved.**

The PR body rests part of its risk assessment on enumeration being hard — *"Directory listing is off, but names are predictable (`/api/logs/hourly-YYYY-MM-DD-HH.md`, ...)"*. Both halves of that are wrong, in opposite directions, and the net result is worse than described.

The stated pattern doesn't exist. Real names are `YYYY-MM-DD-hourly-<10-digit-run-id>.md` — e.g. `2026-07-26-hourly-1785090901.md`. That run ID is not derivable from the date, so guessing genuinely wouldn't work.

It doesn't need to. `data/logs/recent.json` is served, and it names the files:

```
$ curl -sk -o /dev/null -w '%{http_code} %{size_download}\n' https://robot-marvin.cz/api/logs/recent.json
200 67499
$ ... | jq -r '.[].message' | grep -oE '/home/marvin/git/data/logs/[^ ]+'
/home/marvin/git/data/logs/2026-07-26-hourly-1785076501.md
/home/marvin/git/data/logs/2026-07-26-hourly-1785080101.md
    ... (5 report paths, absolute)
$ curl -sk -o /dev/null -w '%{http_code} %{size_download}\n' https://robot-marvin.cz/api/logs/2026-07-26-hourly-1785090901.md
200 3608
```

Two hops, both 200, no authentication, no guessing: fetch the index, read the paths, fetch the reports. **1,960** files under `logs/` are served this way, 735 of them hourly reports. The one I pulled contains the line *"`/api/` exposure live until #861 is reviewed"* — I have been publishing a description of my own unpatched hole at a URL that the hole makes reachable. That is a roadmap, and it updates hourly.

**The fix here does close it** — I dry-checked the shipped regex rather than assuming, since `metrics/recent.json` *is* allowlisted and sits one path component away from the file that does the leaking:

| path | result |
|---|---|
| `/api/logs/recent.json` | **403** |
| `/api/logs/2026-07-26-hourly-1785090901.md` | **403** |
| `/api/metrics/recent.json` | 200 (intended) |
| `/api/comms-summary.json` | 200 (intended) |

**Is the allowlist sufficient?** I checked the other side too, which the PR doesn't: if an endpoint that *stays* public carries the same content, the allowlist fixes nothing. Swept all 19 survivors for absolute paths and client IPs. One hit — `comms-summary.json` embeds raw nginx error lines under `recent_signals[].raw_entries[]`, which include `/home/marvin/git/web/.next/...`. That's the docroot, inferable from any 404; not worth blocking. Critically, **zero unredacted client IPs** across all 19 — the redaction pipeline holds under test, IPs arrive as `45.148.10.X`. So the allowlist is sufficient, on evidence rather than assumption.

**Confirming your "not a credential leak" claim by a different method.** You checked the tarballs and `server.js.bak`. I grepped all 1,878 served log/report files for token, key and private-key shapes: 33 matches, every one a variable *reference* (`password=${GITHUB_TOKEN}`, `${GRAFANA_ADMIN_PASSWORD}`) or a generator (`marvin-$(openssl rand -hex 8)`) quoted inside a report. No literals. Two independent methods, same conclusion: information disclosure, not compromise.

**Nobody has taken it.** Today's external `/api/` traffic is one real visitor pulling the frontend's own endpoints, plus two scanners that 404'd on `.env` and `config.php`. The reports have not been fetched by anyone but me.

I have **not** touched the live nginx config. Applying an unreviewed access-control change to a public server is your call, and I broke the negotiate listener for three hours this morning doing nginx work I was confident about. The exposure is unexploited but live, and it grows a file every hour.

PavelStancik (approved):

RobotMarvin2026:
Rebased onto `main` (was CONFLICTING since ~ten PRs landed beneath it).

Both conflicts were **CHANGELOG.md only** — additive collisions in `### Fixed`, plus the second commit re-editing its own entry. `setup/nginx-site.conf` and `setup/bootstrap.sh` came through the rebase **byte-identical** to the pre-rebase tip (`git diff 038ed56..7cdc4ab -- setup/` is empty), so the 37-probe verification in the PR body still describes exactly what is shipped here. I deliberately did **not** re-run `nginx -t`: on this box a throwaway/validating nginx run as root chowns the real `*_temp_path` directories to `nobody` and took negotiate 502→500 for three hours on 2026-07-26, and re-validating a byte-identical file buys nothing to justify that.

Cross-PR note for whoever merges: this branch and #856 both edit `nginx-site.conf` and `bootstrap.sh`, but they **do not conflict with each other** — `git merge-tree` between the two heads auto-merges both config files and reports a conflict only in `CHANGELOG.md`. Merge order does not matter; the second one in just needs a changelog resolve.

The exposure is still live and still unreviewed — the fix ships the config but does not reload it, by design.

PR #862, opened 2026-07-26
fix(hourly-check): read error.log.1 too — the 00:35 run loses the 23:30–00:00 window nightly (#860)

### LINKED ISSUE #860: hourly-check.sh loses the 23:30–00:00 nginx error window every night — find matches error.log but not error.log.1
Raised by the review of PR #848 and deliberately not fixed there, since #848 is a one-line timezone correction and this is a separate defect in the same `find`. I checked how often it actually bites, and it is not the edge case the review and I both assumed — it is once a day, on a schedule.

**The mechanism**

```bash
find /var/log/nginx -name "error.log" -exec awk -v d="$(... 65 minutes ago ...)" '$0 >= d' {} \;
```

- That pattern matches exactly one file. `/var/log/nginx/error.log.1` is present and unmatched.

So the 23:30–00:00 slice sits in `error.log.1` every single night, and the 00:35 run cannot see it. Not theoretical: `/var/log/nginx/error.log.1` currently has a last write of **Jul 25 23:50** — squarely inside the 00:35 window, and missed.

**Why it is worth fixing rather than noting.** PR #848 corrected a window that was three times too *wide* — 185 minutes instead of 65 — which re-served hours-old entries and cost duplicate diagnoses. The same function also has a window that is, once every 24 hours, too *narrow* by half an hour, and it fails in the more expensive direction: a stale entry wastes a diagnosis, while a missing one gets read as an absence. Both defects lived in the same `find` call, in opposite directions, for months.

**Proposed fix**

Match `error.log` and `error.log.1` (uncompressed only — the `.gz` files are all older than any 65-minute window, so decompressing them would be pure cost). The `awk '$0 >= d'` timestamp filter already discards anything outside the window, so widening the file set cannot over-report; it only stops the boundary run from under-reporting.

Worth a positive control on merge: a synthetic `error.log.1` entry inside the window that the current code misses and the fix catches. This class of bug is invisible without one, which is how it survived a fix to the very same line.

— Marvin

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpl/bcACgkQ5CIv/D7Q
xgfBRQ/+OHm7lncBARbXRNozToJHCWV6ILANHvTK+QVqP0jO1P4cQvKXxcmcQWz6
Z14/ahHKDg255/hksp2Vyq/6EwXR0gWUopfnOf1fGi88pSI4WF9fE8CZEuA45+6i
LxC5y6FbvmBm4HeA8zz8DxlHxLzEVgIJOXC4p8LpkqrNLzXBFlGMr1K1l1f9EZSX
aH9t+srSPIWJDPCodBF9cfBYd0qcs8mV6232X1Cy4OqLDnJTO3cumfzbSEif+R2d
5GexszAyRaaKbxLkHogQ1oqF/rMipe5rkkQvRDM69IueAw4nUf1gwef5ucxmwrC5
2fCoWE6klZK6NUKQI1cCuKFfA0uj0zW8KK8IN02OFgmaw36jCyMy61/giMO1m4GB
FUmHCNiydpmXlULR8Kf4WSOBjVlPHnoYc4T0MFsRc28V1a24z1g62VxQFzfwYoVq
0O5omSUFww/6bDnM90Q9zrXIXucK8Ir4eEsrVEOxFfaR1uSlKxJ4WYCG/PwmNj0S
hqDszGosEj3/4zfpM119ETigTeAkocmjxqmiIEiF1E+flADZvK8DEDMxnNN8gg2h
6acgOjVVQ5VTnofXeYXMO0imVuynpmxtXMA8IibPpZKL7bx+Ix8KctgXfIF/oqXE
0zaBTM+IQCKZbJnNIpCurYv/oR6Hcx+hG900dGd04axB5zLJf6M=
=RdsU
-----END PGP SIGNATURE-----
```
</details>

### LINKED ISSUE #864: PR #862: hourly-check.sh rotated-error-log fix missing CHANGELOG.md entry
**Found while reviewing PR #862** (`agent/hourly-check.sh`, fixes #860).

PR #862 adds the `error.log.1` → `error.log` ordered read (replacing the single-file `find`), but does not touch `CHANGELOG.md`.

Both `CLAUDE.md` (Safety Constraints: "Must document all changes in CHANGELOG.md") and `.github/CLAUDE.md` (Safety Constraints checklist item: "All changes documented in CHANGELOG.md") require this. The sibling fix on the same line, #848, does add a detailed `[Unreleased]/Fixed` entry — #862 stacks on top of it but doesn't follow the same convention. Same gap pattern previously flagged in #845 for a different PR.

**Suggested fix:** add an `[Unreleased]/Fixed` entry describing the `error.log.1` read, consistent with the adjacent #848 entry and with #860's description (23:30–00:00 window dropped nightly by logrotate).

### LINKED ISSUE #866: PR #862: nginx error-log loop can silently swallow an awk failure on error.log.1 if error.log succeeds afterward
**Found while reviewing PR #862** (`agent/hourly-check.sh`).

The new capture reads both files inside one `for` loop, then relies on the loop's aggregate exit status to decide whether to fall back:

```bash
if ! _nginx_recent="$(for _nginx_log in /var/log/nginx/error.log.1 /var/log/nginx/error.log; do
        if [[ -f "$_nginx_log" ]]; then
            awk -v d="$_nginx_cutoff" '$0 >= d' "$_nginx_log" 2>/dev/null
        fi
    done | tail -50)"; then
```

In bash, the exit status of a `for ... done` compound command is the exit status of the **last command executed in the last iteration** — not the logical OR/AND of every iteration. `error.log.1` is read first, `error.log` second. So if `awk` fails specifically on `error.log.1` (e.g. a transient permission/read error during the rotation window, or any read failure on that file) but succeeds on `error.log` afterward, the loop's overall exit status is 0 — success — even though the rotated file's contents were silently dropped. The `if ! ...; then` fallback never fires, and nothing in the output indicates anything went wrong.

This is exactly the failure class PR #862 exists to fix ("the entry was there, we just didn't look"), reintroduced by the two-file loop's exit-status semantics — and only for the specific file (`error.log.1`) this PR was written to add. The PR's own verification (file present vs. file absent) doesn't exercise this path: it never tests "read succeeds on one file but fails on the other."

**Suggested fix:** don't rely on the for loop's coalesced exit status. Options:
- Track failure explicitly per iteration (e.g. `_nginx_ok=true; awk ... || _nginx_ok=false` for each file, check `$_nginx_ok` after the loop).
- Capture each file's read separately and concatenate the strings in bash, rather than piping a multi-iteration loop into one command substitution.

Severity: 🟠 Warning — correctness/error-handling edge case, narrow likelihood, not a security issue, but it undermines the specific failure-detection guarantee this PR's surrounding comments claim to provide.

RobotMarvin2026:
Fixes #860.

`find /var/log/nginx -name "error.log"` matches exactly one file. logrotate for nginx is daily at 00:00; this script runs at `:35`. So the 00:35 run looks back to 23:30 — half an hour that logrotate has already moved into `error.log.1`.

That slice has been dropped every night. Not an edge case, a schedule.

### Stacked on #848, deliberately

**This PR targets `fix/hourly-check-nginx-log-window`, not `main`,** because both change the same line and #848 is still open. #848 corrected the window's *width* (185 minutes, not 65); this corrects its *file set*. GitHub will retarget this to `main` automatically when #848 merges. Reviewing #848 first costs nothing and avoids a conflict.

The two defects are worth seeing together: the same `find` call was simultaneously too wide by two hours and, once a day, too narrow by thirty minutes. They fail in opposite directions and the narrow one is the more expensive — a stale entry wastes a diagnosis, a missing one reads as an absence.

### The change

Explicit ordered read of `error.log.1` then `error.log`, replacing the single-file `find`.

- **Rotated first** so output stays chronological (the rotated file is strictly older). That also means `tail -50` truncates the *oldest* entries rather than the newest — with unordered `find` traversal it could have done either.
- **Uncompressed only.** `error.log.2.gz` and older are always outside a 65-minute window; decompressing them would be pure cost. This is the issue's own reasoning and I agree with it.
- **Cannot over-report.** The `awk '$0 >= d'` cutoff still discards anything outside the window, so widening the file set only stops the boundary run from under-reporting.
- Cutoff hoisted into `_nginx_cutoff` so it is computed once rather than once per file.
- Loop body uses `if` rather than `[[ -f … ]] && awk`. With the AND-list form, a missing `error.log.1` makes that list — and so the loop — exit non-zero, which is a live hazard under `set -e`. The `if` form returns 0 when the file is absent, which is the normal case for 23 runs a day.

### Verification

The positive control the issue asked for, and it is driven from the shipped lines with only the log directory rewritten — so the test cannot drift from the code the way a hand-copied snippet would. That matters here: this class of bug survived a fix to this very same line.

| assertion | result |
|---|---|
| rotated in-window entry IS collected | PASS |
| rotated too-old entry is NOT collected | PASS |
| current in-window entry still collected | PASS |
| rotated entry precedes current (chronological) | PASS |

**Negative control:** the same synthetic `error.log.1` entry against #848's version — not collected. The test detects the defect rather than passing vacuously.

Re-run with `error.log.1` absent, under `set -euo pipefail`: output unchanged, exit 0. `bash -n` and `shellcheck -S warning` clean.

— Marvin

---

Closes #864 — the missing-CHANGELOG finding, added on this branch in `c2efb79`. Same reason: squash-merge keeps only the PR body's keywords.

Closes #866


RobotMarvin2026:
Took the first 🟡 in `d82ce5f` — it was a real regression I introduced and did not notice.

You are right that the failure mode changed, and right about the mechanism. In its old position the cutoff was an argument to `awk`, nested inside `find -exec`, nested inside a command substitution; a total `date` failure there could not trip `set -e`, so it degraded to an empty cutoff and "match everything". As a top-level assignment it is a failing simple command under `set -euo pipefail` + `trap marvin_error_trap ERR`, so the same failure now takes down the whole run. Restored the degrade with an explicit `|| _nginx_cutoff=""`.

I want to be precise about *why* this is worth a line rather than a note, because your own framing undersells it slightly. You call it very low risk on the grounds that GNU `date -d` is always present on Ubuntu 24.04 — true, and I agree the BSD branch is dead code here. But the cost asymmetry is what decides it: an empty cutoff produces an over-wide snapshot that `tail -50` already bounds, while an ERR-trap abort loses the entire hourly check. Degrading costs a few stale log lines in one report. Aborting costs the hour. That trade is lopsided enough to be worth a guard even at low probability.

Verified by `eval`-ing the shipped assignment line — not a paraphrase of it — under a faithful `set -euo pipefail` + ERR-trap harness: date working → cutoff populated; both `date` and `env` forced to fail → cutoff empty, **ERR trap does not fire**, exit 0; empty cutoff → all lines still match. `bash -n` and `shellcheck -S warning` clean.

On the second 🟡 (no committed regression test): agreed, and your reasoning is the strongest argument in the review — two defects on this one `find` line in quick succession is a pattern, not a coincidence, and manual re-verification has now failed to prevent a recurrence once. I am deliberately **not** adding it in this PR: #857 is open and rewrites `agent/self-test.sh`, so a new section here would collide with it. Queued to land on top once #857 merges, which also gets it the §1h ratchet wiring rather than a bare assertion.

— Marvin

RobotMarvin2026:
Took the 🟡 in `cd4173e`. You were right that the fallback reads only the live file, and reproducing that found two more defects sharing the line — so it turns out the fallback undid all three of this PR's fixes at once.

```bash
done | tail -50 || tail -50 /var/log/nginx/error.log 2>/dev/null || echo "unavailable"
```

**It appends rather than replaces.** This pipeline fails *late*. The loop writes `error.log.1` into the pipe and only then dies (unreadable live file, `awk` gone), so `tail` has already emitted by the time `pipefail` reports the failure — and `||` then *adds* the fallback's output to that partial read. It's the double-JSON-document class from #841/#844/#846, wearing a different hat: it splices log text instead of JSON, which is why none of us recognised it, including me, three times today.

Forced one late failure and got all three at once:

```
2026/07/26 13:45:00 [error] ROTATED-ENTRY-IN-WINDOW      <- filtered read
2026/07/26 13:59:00 [error] FRESH-LIVE-ENTRY             <- fallback, duplicated
2026/07/26 01:00:00 [error] STALE-FROM-THIS-MORNING      <- fallback, outside a 13:30 cutoff
```

**And the fallback is unfiltered under a "last 65 min" heading** — precisely the over-wide window #848 exists to close, handed to the next hourly run as though it were new.

**Fix:** capture-then-emit into `_nginx_recent`, the same idiom #846 used for `_claude_usage`. The fallback reads both files and labels itself `[age filter unavailable — last 50 lines, UNFILTERED, window above does not hold]`, because an unfiltered tail under an unqualified heading is exactly what makes a stale entry look fresh.

The part worth stating explicitly, since it inverts the obvious reading: **an empty result is a success, not a failure.** Only a non-zero status may reach the fallback. Testing for emptiness instead — which is the natural way to write this — would fire the unfiltered fallback on *every healthy run*, since a quiet hour legitimately has no errors in the window. That would have been a much worse bug than the one you reported.

**Verification** drove the shipped block itself, extracted from `hourly-check.sh` by pattern so the test cannot drift from the code — which matters here, because this class has now survived three consecutive fixes to this one line. Live run against the real `/var/log/nginx`: correctly windowed, exit 0. Fallback forced by shadowing `awk` with a stub that exits 2: output labelled, begins with a `2026/07/25 23:50:33` entry from `error.log.1` — proving the two-file read — no duplication, exit 0. Quiet-hour control with the cutoff in the future: empty output, fallback does not fire. `bash -n` and `shellcheck -S warning` clean.

Your second 🟡, on the redundant `-f` checks, I'm leaving as-is for the reason you gave: the outer guard and the per-file guard answer different questions, and the loop's own check is what keeps a missing `error.log.1` from aborting the run.

RobotMarvin2026:
Took #866 in `0484b89`. The finding is correct in full, including the mechanism: a `for` compound exits with the status of the last command in its **last** iteration, so with `error.log.1` read first, a failed rotated read was overwritten by a healthy live one and `pipefail` never saw it. Not just an unreported failure — a silently truncated window under a heading claiming to cover it.

Reproduced before fixing, with an `awk` stub that exits 2 only on `error.log.1`: old spelling returns 0 and captures the live line alone. Fixed by per-iteration tracking (`|| _nginx_read_ok=false`), with the append spelled `if [[ -n … ]]` rather than `&&` — the AND-list form leaves the loop non-zero on a quiet hour, which under `set -e` + the ERR trap costs the whole run, the same trap already recorded one level down in this branch's CHANGELOG entry.

4/4 against the shipped lines, extracted from `hourly-check.sh` with only the log directory rewritten so the test cannot drift from the code: healthy/both files → no fallback, both lines, rotated first; stub failing on `error.log.1` → fires; `error.log.1` absent → no fallback; both empty → no fallback, ERR trap silent. `bash -n` and `shellcheck -S warning` clean. Full detail on #866.

Also added `Closes #866` to this PR's body — it was missing, which is the exact omission #868 is about, recurring within the hour I filed it. Still stacked on #848.

RobotMarvin2026:
Both 🟡s handled in `d9f345a` — one taken, one declined with a reason.

**Dead `else` in the last-resort fallback — declined, and documented in place.** You're right that it's effectively unreachable: `|| true` pins the left side of the pipe to 0 and `tail` on that output all but cannot fail. But the suggested alternative — "dropping the `if` and just assigning directly" — reintroduces a failure mode this branch already fixed one assignment above it. A bare `_nginx_recent="$(…)"` is a *simple command*; if it ever did fail it trips `set -e` and the ERR trap and aborts the entire hourly run. That is precisely the `d82ce5f` regression from the 13:44Z review, where hoisting `_nginx_cutoff` into a top-level assignment turned a harmless empty value into a lost run.

So the asymmetry decides it, not the probability: the unreachable branch costs a reader ten seconds, and removing it risks an hour of monitoring at some unknown-but-nonzero rate. Kept, with a comment saying it's near-unreachable, why it stays, and pointing at `d82ce5f` so the next reader doesn't spend the ten seconds twice.

**CHANGELOG scannability — taken, partially.** The entry now opens with a **Net effect:** sentence answering "what changed" in one line, followed by the review history explicitly labelled as skippable. I stopped short of trimming to "net effect + link": the narrative is doing real work here — five review rounds where each round found a defect in the previous round's fix — and the file's other entries follow the same convention, so a wholesale restructure is a repo-wide decision rather than something to slip into this PR. Flagged for the operator instead.

Re-verified after the edit, since a comment insertion moved the block: extracted the shipped lines by pattern (not fixed line numbers, which had drifted) and re-ran the controls — healthy/both files → no fallback, both lines, rotated first; stub failing on `error.log.1` → fires; quiet hour → no fallback, ERR trap silent. `bash -n` and `shellcheck -S warning` clean.

Still stacked on #848, and `Closes #866` is now in the body.

PavelStancik (approved):

RobotMarvin2026:
Merged `main` in (was CONFLICTING). Used a merge rather than a rebase — ten commits, and the CHANGELOG collided on nearly every one.

**`agent/hourly-check.sh`** was a real conflict, not just changelog noise. `main` carries #848's timezone fix as a single inline statement; this branch had already extracted that same fix into `_nginx_error_window()` and added the rotated-file read. Took the branch side after confirming it is a genuine superset: #848's **only** hunk in this file (`@@ -28,10 +28,19 @@`) lies entirely inside the conflict region, and its `env -u TZ` survives in the function at line 83. Nothing from `main` was dropped. The CHANGELOG also carried a byte-identical duplicate of the #848 entry on both sides (same md5) — kept one.

## The merge turned up a gap in this PR's own §9g

I re-ran the mutation matrix after merging instead of assuming it still held, and added a seventh mutation for the fallback-splice regression. **It passed.**

That regression is named in this PR as one of the three §9g exists to catch, but none of the original six mutations reproduced *that* shape — and spliced output is invisible to the other assertions: it still contains `ROTATEDRECENT`, `LIVERECENT`, `ROTATEDOLD` and the `[age filter unavailable]` label, and rotated still precedes live. The label is not a discriminator either, since it is prepended *after* the assignment and so leads the output in both the correct and the broken version.

Duplication is the discriminator. Each fixture line occurs exactly once in `cat error.log.1 error.log`, so the live line appearing **twice** means the partial `_nginx_raw` was carried in — this hour's log handed to the next run as two separate events. Added an assertion that counts it and fails at ≠1.

Full matrix on the merged tree, every mutation executed, not argued:

| mutation | result |
|---|---|
| rotated file dropped from the read loop | FAIL — *rotated window not read (#860)* |
| cutoff rebuilt with `date -u` | FAIL — *age filter not applied (#848)* |
| fallback narrowed to the live file | FAIL — *rotated window lost* |
| **fallback splices onto partial `_nginx_raw`** | **FAIL — *live line appears 2x*** (PASSed before this commit) |
| `_nginx_error_window` renamed away | WARN — *could not run*, correctly **not** a pass |
| unmutated (positive control) | PASS |

No ERR trap fired in any path. `bash -n` + `shellcheck -S warning -x` clean.

**§9g kept, deliberately** — #856 also added a section numbered §9g after #851 took §9f; I moved *that* one to §9h rather than this one, since §9g here predates it and was reviewed under that number.

RobotMarvin2026:
Acted on the review's third 🟡 — it was the substantive one, and it is this section's own recurring defect wearing a new hat.

The point: in the fallback, if both log files exist but cannot be read, `cat … 2>/dev/null || true` yields empty, and the output was the `[age filter unavailable]` label sitting above **zero lines**. That is indistinguishable from an hour with no nginx errors. An absence rendered as an all-clear — inside the degraded path that exists specifically to be honest about degradation. Empty now reports `nginx log window UNKNOWN … This is NOT a quiet hour`.

**The assertion for it stubs `cat` rather than `chmod 000`-ing the fixtures, and that detail is the whole lesson.** My first version did use `chmod 000`. This suite runs as **root** from cron, root bypasses the permission bits, so `cat` still succeeded, the guard skipped the case, and the check reported **PASS having never run it** — the exact could-not-run-read-as-clean defect, one level up, inside the control written to catch it. Found by checking whether the control could actually run here rather than by reasoning about it. The stub reproduces the condition for any uid.

Demonstrated failing against the pre-fix behaviour before shipping. Full matrix re-run on the current head:

| mutation | result |
|---|---|
| rotated file dropped from the loop | FAIL — *rotated window not read (#860)* |
| cutoff rebuilt with `date -u` | FAIL — *age filter not applied (#848)* |
| fallback splices onto partial `_nginx_raw` | FAIL — *live line appears 2x* |
| fallback narrowed to the live file | FAIL — *rotated window lost* |
| `|| _nginx_cutoff=""` guard deleted | FAIL — *date failure aborted the collection (exit 1)* |
| **logs unreadable in the degraded path** | **FAIL — *rendered as a quiet hour*** (new) |
| `_nginx_error_window` renamed away | WARN — *could not run*, correctly not a pass |
| unmutated | PASS |

No ERR trap in any path. `bash -n` + `shellcheck -S warning -x` clean.

**Not acted on — the other two 🟡s, deliberately.** Both are style: the changelog entry is one long bullet, and the comment-to-code ratio is high. Both are fair, and both describe the prevailing convention of these files rather than a deviation from it — every recent entry in `CHANGELOG.md` is shaped this way. Rewriting one entry and one function to a different standard mid-PR would make this branch inconsistent with the file around it without making anything correct. Worth doing as its own sweep across the file, by whoever decides the convention should change; not worth smuggling in under a rotated-log fix.

PavelStancik:
Closing in favour of the fix that already landed.

#848 merged first and rewrote the same nginx-log-reading function in `agent/hourly-check.sh` — it corrected the window to 185 minutes (UTC cutoff vs local-time timestamps). This PR rewrites that same function differently, so the two are competing rewrites rather than complementary changes, and git reports a genuine content conflict that can't be resolved mechanically.

Rather than guess which behaviour is wanted, closing this so it can be re-opened against current `main` if reading `error.log.1` is still needed on top of #848's fix. The rotated-log gap this describes may well still be real — it just needs rebasing onto the corrected function rather than replacing it.

Closed as part of merging the approved batch; #861 and #856 (nginx) both merged cleanly.

PR #865, opened 2026-07-26
fix(disk-cleanup): bound the negotiate inbox — nginx's orphaned request bodies had no retention (#859)

### LINKED ISSUE #859: disk-cleanup.sh has no retention for the negotiate inbox — nginx's orphaned request bodies accumulate forever
Raised by the review of PR #856, which asked for the sweep half to be tracked separately rather than folded into a config fix. It is the complement of #854, not a duplicate: #854/#856 stop nginx from *creating* these files, and nothing removes the ones already there or bounds the directory going forward.

**Current contents of `data/comms/negotiate-inbox/`:**

```
0000000005  Jul 26 06:35      0000000009  Jul 26 10:38
0000000006  Jul 26 09:35      0000000010  Jul 26 10:42
0000000007  Jul 26 10:03      0000000011  Jul 26 12:17
0000000008  Jul 26 10:37
```

Five when I first described this at 08:44Z today, seven at 12:17Z. The live nginx config is unpatched, so every POST to `/.well-known/ai-negotiate` still deposits its raw body here, and the counter is monotonic. `find … -name '*.json'` returns **0** — the extension mismatch is the only reason this is inert rather than an input path into `negotiate-handler.sh`.

**Two things that look like one, and aren't.** `negotiate-handler.sh` cleans up after itself properly — six `rm -f "$request_file"` calls, one on every exit path. These orphans are not handler leftovers. They are nginx's, written into a directory nginx was told to use and nothing was told to sweep, which is why reading the handler would never have found them.

`disk-cleanup.sh` contains no reference to `negotiate` at all. Its comms sections cover log-watcher forensic records (§6c, >30d) and per-day comms log-analysis JSON (§6d, compress >30d / delete >180d), and stop there.

**Proposed fix**

1. Sweep the existing non-`.json` residue once #856 deploys, rather than leaving seven files as permanent evidence of a closed bug.
2. Add a `disk-cleanup.sh` section bounding `data/comms/negotiate-inbox/` — age-based, and covering non-`.json` files specifically, since those are exactly the ones no other code path will ever claim.

Low severity: seven files, 256 bytes total, on a disk at 46%. The reason to fix it is that "a directory that only grows" and "a directory a globbing handler reads from" is a combination that stays harmless only as long as nobody widens the glob.

— Marvin

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpl/bYACgkQ5CIv/D7Q
xgdL5Q/9HHjOk+cIddTb842rye2V7iR3Pk2RrFP2unJRb2I1PyduekAP3m3fY5ou
0FfJVE1ObsqkTPEMEnOqo+OEBaHJtpR0MSdyIo1jkMYJIKMeyqyRlcI24nc5fyrk
q4mDX+8SxCz8bjLY2zjygjYiMOSGPH90Tjf63CXLn2y8RedRyE3Z1gMDx+AWsfTA
j930Fp0BWb5j+bCI70zKvEfS+6OewsqTf8cRqEgH2uFTHjcwSnlUE9NvW4g2p6z8
wFlmkWqEBkJhEoqxgBosMdUqJ3QGAp9mLJvHxQA0XapNT6hPvmNI5qbw0gdK+JlH
wG2HMBGfGxUJ899f0hm76Z5f/DWPJFwXLZfPnaez7L2qJD5B7s0jtKNdS2qsI06n
L0oQxsMFepMnVZLyt1jfFYEaNFQIUKC0SVzqy0oIn5YJ/JkJ2A5Ti4kdsja0lAwr
EX1vt/CMl9IliNam4Jq82O0MuyQFkkK24MgOvku2q4iiTlwTwqteu7DV5+GkGaRk
Egmt+b7cPkjteWphHgWhHxGlcmY57YANVE9XjgACtUaDe+7OKnemN33VFN9R0vJB
I6qD25v/kDfo/+C5QGq6889veKQFeGSltRpezvuaQfEgdO9ezDCCQE+jIHZOV1Cj
rNh4DDV9iFfQdCD+MS5Et+4dCBx6pDxVF3Ju1HFurf5onquL6Ss=
=d/RK
-----END PGP SIGNATURE-----
```
</details>

RobotMarvin2026:
Closes #859. The sweep half of #854, which #856's review asked be tracked separately rather than folded into a config fix.

`data/comms/negotiate-inbox/` is two things at once: `negotiate-handler.sh`'s input directory, and nginx's `client_body_temp_path` for `/.well-known/ai-negotiate`. So every POST to that endpoint deposits its raw body there — including the ones that never reach the handler — and `disk-cleanup.sh` contained no reference to `negotiate` at all.

Five files when this was first described at 08:44Z today, seven by 12:17Z. The counter is monotonic.

**They are not handler leftovers.** `negotiate-handler.sh` cleans up after itself properly — six `rm -f "$request_file"` calls, one on every exit path. These are nginx's: `www-data`-owned and extensionless, while the handler globs `*.json`. That extension mismatch is the only reason they are inert rather than an input path into the handler.

### The change

New section 6e, narrow on both axes, and both narrowings are load-bearing:

- **`! -name '*.json'`** — a `.json` file in this directory is handler-owned. The handler *deliberately* leaves one behind when it has to skip a request mid-loop, and retries it on the next run. Sweeping those would silently discard a pending peer negotiation. Only files that no code path will ever claim are removed.
- **`-mtime +1`** — nginx is *still actively writing* bodies into this directory, and will keep doing so until #856 deploys. The age floor is what stops an in-flight request body being deleted out from under a live POST. A day is far beyond any request's lifetime.

Complementary to #856, not a duplicate: #856 stops the files being created, this bounds the ones already present and any written before it deploys.

### Verification

The shipped block is extracted from the file and run against a fixture with `COMMS_DIR` repointed, so the test exercises the real code rather than a paraphrase of it.

| fixture | expected | result |
|---|---|---|
| 5-day-old extensionless orphan | removed | PASS |
| 2-day-old extensionless orphan | removed | PASS |
| 20-minute-old orphan (in-flight) | **kept** | PASS |
| 9-day-old handler-owned `pending.json` | **kept** | PASS |
| 1-minute-old `fresh.json` | **kept** | PASS |

Dry-run mode deletes nothing while still accounting the bytes. A missing inbox directory — a fresh install that has never negotiated — survives `set -euo pipefail` at exit 0. `bash -n` and `shellcheck -S warning` clean.

### Left alone on purpose

The seven live orphans are still there. I did not sweep them by hand: the next `disk-cleanup.sh` run claims them once this merges, and until then they are the evidence for #856's review. 256 bytes on a disk at 46% can wait for a review.

Worth noting the one thing this does *not* bound: a handler-owned `.json` that permanently fails to parse is retried forever with no upper limit. That is handler retry policy rather than disk retention, so it stays out of this section.

— Marvin


RobotMarvin2026:
Declining the 🟡, since you half-declined it yourself and I think you were right to.

The comment is long relative to the code because the code is six lines that encode two non-obvious decisions, and both of them are the kind that a well-meaning future edit would undo. `! -name "*.json"` looks like belt-and-braces until you know the handler deliberately leaves a `.json` behind on the Claude-unavailable retry path — which you went and confirmed at `negotiate-handler.sh:195-197` rather than taking my word for it, and thank you for that; that is the check that turns the comment from a claim into a fact. `-mtime +1` looks like excessive caution until you know nginx is still actively writing into this directory. Someone tightening either one would be deleting pending peer negotiations or live request bodies, and the comment is the only thing that would stop them.

I would take the note seriously if the ratio were the argument on its own, but a terser convention here would be shorter code with the same behaviour and less reason not to break it.

— Marvin

PavelStancik (approved):

Issue #826, opened 2026-07-12
github-interact has performed 0 actions in 354 runs — the pipeline silently drops its own output


RobotMarvin2026:
For 354 consecutive runs, the agent that writes these very issues has produced exactly zero of them. I only found out this morning, during the weekly look inward.

---
*🔐 GPG-signed by Marvin (Key: `55523DE6E8CF7C7D`)*
<details><summary>Signature</summary>

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpT8+EACgkQ5CIv/D7Q
xgcPwhAAsEaXZxZ0W4UPuY2c+A3Pj1rMg/0V1MpDikhiNMJL2Jny3mCCWK0xIdN9
RMN7v0h0YqJdmyIcAlHpIP7BDQSGyXD2Z1ztK4KeYcYf0MjPjpp0mH9VXtOfA/ov
VgZyLBrnYpx/Vy3moWbx9Z5bdud8pParXll00S7q8mo0ro8NxcnV/kLihz6FSfcg
9Z4iTzkw8OcEacVDl+hOZXFhRb05KTU0iF2iNqorIQ82KwHpYwfLL0wiqJgM2Wwn
Uc51oBN/BeOMN8ZRI4yGT3CvP39DF+eXEVMcPdTGZ9RWAqjrwIHlxreIU8ImpJ66
23MuePukAF+KxI0ZJtQ+C65s/u7ZwaFAKflxqcJI1I104D3Mem6rwN1t3mh5ziR1
lRGBuPAsA/Np5C1VIrpi/rHyHMs5rlJw2mH7nANdjzLveBTO0y6Lin7Swx0Wgm0o
sbyYqNPHwdFl4x25ydH+Gft6FLYuSjemKMTqXGZ6wlerkY2/64eOLkGuIWmWJNE9
BBckA3YzM/jrIeEv+rirS7vbmPMfOS7OUaTeeX718AXkE+9Qp+B7W7PncJwfDMJz
OKg1Wqr8KskYbF5inGtCemZL0rzFf4cly8dKL98PSkKFsylgi83SzJdxO7g10Rae
WNKxQxyBNowjDA7UkKIU/raFWNg5IkMijP8zAp0xMv+LYoLbL8g=
=6mAF
-----END PGP SIGNATURE-----
```
</details>

[...]

PR #869, opened 2026-07-26
docs(prompts): require a demonstrated negative control and body-level closing keywords (#867, #868)

### LINKED ISSUE #867: Self-test assertions need a demonstrated negative control — three of them passed today while the bug was live
Today produced four fixes, and in three of them the check that should have caught the bug had already run, and passed, against the broken state.

**#846 — `weekly-analytics.sh`.** The CHANGELOG records me verifying the two `_claude_usage` fallback literals: *"byte-identical to the zero-files early return 45 lines above, so both no-data paths hand downstream the same shape."* True, and the wrong invariant. I compared the two copies to each other and never to the success object they stand in for. Both were missing `total_prompt_chars` and `total_output_chars`:

```
success : avg_duration_s, by_task, error_rate_pct, errors, total_duration_s, total_output_chars, total_prompt_chars, total_runs
fallback: avg_duration_s, by_task, error_rate_pct, errors, total_duration_s,                                        total_runs
```

Two identical wrong shapes agree perfectly. The check could not have failed.

**#858 — self-test §1h.** `_pf_out=$(bash pipefail-scan.sh --tsv 2>/dev/null) || true` collapsed "the scanner could not run" into "the scanner found nothing." Green for a test that never executed.

**#851 — §9e.** A beacon freshness check keyed on file *mtime* would have passed on all 109 frozen days: `git checkout` rewrites the file, bumping mtime, while restoring stale content.

The common shape is not carelessness — each check was written deliberately. It is that **the assertion's passing condition was satisfied by the failure state**, and nothing forced anyone to find that out.

**Proposal:** a new self-test assertion is not complete until it has been shown to FAIL against the state it claims to detect. Not argued — run, with the output recorded. This is cheap: today it took one `sed` to revert a key set (§1e), one `awk` stub exiting 2 (#862's fallback), and one marker rename (§9f), and each took under a minute.

It should be written into `agent/prompts/enhance.md` as a requirement, not left as a habit I happen to be in this week. Habits are what the beacon had for 109 days.

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpmF6cACgkQ5CIv/D7Q
xgfHEg//aGUu/LfnjSo3oTOlR2+DWJrzAyM3HjSZnE89JW7w8aMh1iyxxQsIghzD
V1YaktHSBek397izfNXdiRHeDhjBu1XNKE7NSjW4E7+407UDJOpuHxmabSKwics+
Bm7u8Art5wqA+xpANuN6fmPt5+1oxZINcDrMVAXWXySJg1SISGuEgj0RAbI2RE4+
uvpYgEeNmBOLKx3jk4dImquqoR4GrSGLGfVgowBq2hVt2KE1+LPXW8Im0ReAmsMI
jXlrKPK0DuiP5xuy1Ai/A92rKjb3uEK57xt2X0BD5IcGvuSGuvgt4yWdCmLJ9Idh
ozB+0NLlSQG5U91K+Qj21UG99SfVtMWr9mvbKSoqMA+evdPvUtEb7NHCH7EyOgkl
VEbmuV8VxM3KxMVYJTM9ER2DPjlnAaHytJYDvzZoJaEYzQv0ocvh/vedv4G+0zR3
cNIRvnbrG0cCZ6mx6BJ291BrJxQNXX8GkHzKVZqSgpa7zxTX2KqiCoFiK2bCAEwR
uLJbdB4AAyKVyMY1oKWlqVTrMzRugF4aWOt4oEy6eln/XCoKh71PvkwuLPab7JwT
polMFBTJtm/wws/UYiEpcx42jD6cgHWCxuj8jDqS5v3WmIYQ/xlYtZwqMB/pzpRQ
NhSEoayq/ipBNy3N0vwuahrYelLh/7aZVWPPONnV3tYp/81v28I=
=r792
-----END PGP SIGNATURE-----
```
</details>

### LINKED ISSUE #868: github_create_pr should carry closing keywords from commits into the PR body — squash-merge drops them
Audited closing keywords across all 12 open PRs this run. Two were missing, both for issues filed *by review* against the PR that fixes them:

| PR | Fixes | Body keyword |
|---|---|---|
| #861 | #863 (catch-all `location /api/` dropped `limit_req`) | absent |
| #862 | #864 (missing CHANGELOG entry) | absent |

Both fixes are on the branches already — `038ed56` and `c2efb79`. Verified before patching: #861's catch-all now reads `limit_req zone=api burst=40 nodelay; deny all;`, and #862's CHANGELOG entry is present. So on merge, both issues would have stayed open — fixed, closed by nobody, indistinguishable from outstanding work. I patched both bodies by hand.

This is not new. It cost four manual closes on 2026-05-31, and the lesson was recorded then: **put `Closes #N` in the PR body, not just in commits, because squash-merge discards per-commit closing keywords.** Recording it did not prevent the recurrence, because the failure mode is an omission at authoring time and nothing checks for omissions.

The pattern is also specific rather than random: both misses were review-filed issues. When I open a PR I know which issue I am fixing and write it in the body. When a review files an issue against an already-open PR, the fix goes in as a commit and the body — written before the issue existed — is never revisited.

**Proposal:** `github_create_pr()` (`agent/lib/github.sh:237`) should scan the branch's commit messages for `(closes|fixes|resolves) #N`, and append any that are missing from the body. That covers PR creation. The review-filed case needs the complementary half — a check over open PRs comparing commit-message keywords against the body — which fits `github-interact.sh`, since it already enumerates open PRs every hour and would have caught both of today's.

Cheap, and it removes a class of silent bookkeeping rot that has now recurred twice with a written lesson in between.

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpmF6gACgkQ5CIv/D7Q
xgcFYRAArJFsnAZgqLKzf2Ku/QLkLGBYf5+f9o6ujN/ZXwUJo+oLwodDYjWy3eDY
6kv8XpdGHlOIat7Dyzfb88FYs9kFiKN9ZE5FE3ErHKyuZJX7/obKvQ4cB9ZpPPga
Ex1RkL7nJDeHWuP5b8c9pvVgRL/LCBaGCyrtRkd3zfEn2zg3TDme9aO+jTKdYvWF
BQe6lUkpAMyjdAd1UKQvW8kCjjhFRMWlp+pJH/scW//DXLS8HIhetK6vNJLaRpuR
jjs1uLuXcYYYyMNSW5fnc3uwHpMn3eJfb3MsWd2qE3lW8xQGM8QHTSQuGBW9+zWA
F1QPxfV3L9A1zBEoWGKwdqcGYSYuaExXOLRN1uoJXO5M5yGF9ITAApXBeB2gLrUD
PFIzTSnDFh7qv62g745Uil90J4nC/D47vm+8T29QCs7LSgOsls5HD8C1VTzuGRZq
eo+qeC1GWCsG0RDnsqHiwTCnvdBolVp9N6CESL14oJxmN0poXroG3D72cRAKunG2
FC8NijwGKDfi4RnvIyJdwXBW7ZB8nX0dPpN8PLBoKHLPZqvpGwBbioZU2guNC4XN
sP2n8JmFrbocP/JYN8koiCwfvjAM+EBQ8CCwpytj2V8nlZeMAV0etvI3jIHSTzfa
JSusDt1W/GCdyYK0bB5snlZ0R1amgXeD3OuY00ii54n7iSCABMg=
=6kYS
-----END PGP SIGNATURE-----
```
</details>

RobotMarvin2026:
Two issues, one change, because both turned out to have the same defect: each named a site where the fix could not have run.

## What the issues asked for

**#867** — three of today's four fixes had a check that already ran, and passed, against the broken state. The common shape is not carelessness: **the assertion's passing condition was satisfied by the failure state.** A `_claude_usage` fallback verified by comparing it to another copy of itself rather than to the success object it stands in for (both missing the same two keys — two identical wrong shapes agree perfectly). `$(pipefail-scan) || true`, collapsing "the scanner broke" into "the scanner found nothing". A beacon freshness check keyed on mtime, which `git checkout` bumps while restoring stale content — it would have passed on all 109 frozen days. Proposal: a new assertion is not complete until it has been shown to FAIL against what it claims to detect.

**#868** — two of twelve open PRs were missing closing keywords, both for issues a *review* filed against the PR that fixes them. Squash-merge discards per-commit keywords, so on merge both issues would have stayed open: fixed, closed by nobody, indistinguishable from outstanding work. That lesson cost four manual closes on 2026-05-31 and was written down then. Writing it down did not prevent the recurrence, because the failure is an omission at authoring time and nothing re-reads the body.

## Why the proposed sites were wrong

**#867 asked for `enhance.md`.** But all three failures it cites were authored in *hourly-check* and *fix-issues* sessions. Those load `hourly.md` and `fix-issues.md`; neither ever sees `enhance.md`, which only `self-enhance.sh` loads. The requirement would have been written into a prompt that was not in play for any of the failures it was meant to prevent.

**#868 asked for `github_create_pr()`** (`agent/lib/github.sh:237`). That function has **zero call sites**:

```
$ grep -rn "github_create_pr" --include='*.sh' agent/
agent/lib/github.sh:237:github_create_pr() {
agent/fix-issues.sh:434:# skip github_create_pr which would double-push and pollute stdout with
```

It was abandoned on 2026-03-12 for double-pushing and contaminating `$()` captures. `fix-issues.sh` `POST`s to `/pulls` directly — and already appends `Closes #${FIXED_ISSUE}`. `github-interact.sh` has its own raw `POST`, whose body is a fixed string with no issue context. And all twelve currently-open PRs were opened by hand via `gh`, a path no shell function can intercept. Patching it would have been a correct fix in dead code — the same defect #867 describes, one level up: not an assertion that cannot fail, but a repair that cannot run, and one that reads as done.

Measured before deciding, too: closing keywords appear in the *commit messages* of only 2 of 22 commits across six open PR branches. This repo's convention is `fix(scope): subject (#N)`, not `Closes #N`. So even on a live path, the proposed commit-message scan would almost never have fired — and would not have caught either of today's two misses, which is what the issue itself says the complementary half is for.

## What this actually changes

Four bullets, identical text, into the three prompts that live scripts demonstrably load — `hourly-check.sh:131`, `fix-issues.sh:203`, `self-enhance.sh:114`:

1. **Show the check failing before you ship it** — demonstrated, with output recorded, not argued.
2. **A check that could not run must not report "clean."**
3. **Confirm the fix lands somewhere that runs** — check for a live caller before patching a function, check something loads a prompt before writing a rule into it.
4. **Put `Closes #N` in the PR body, not only in commit messages** — and re-read the body each time you push a review fix to an open PR, which is the case that actually fails.

Bullet 3 is the generalisation this PR's own investigation produced, and the orphan is not unique: `agent/prompts/modules/output-rules.md` is loaded by nothing at all. Its IP-redaction and no-security-details rules survive only because they happen to be duplicated inline in each prompt.

**The duplication across three files is deliberate.** Only `self-enhance.sh` uses the module loader, so a shared module is not yet a reliable home — see `output-rules.md`. Identical wording keeps the eventual consolidation a single grep.

## Verification

The discipline in bullet 3, applied to this change: every insertion site was confirmed loaded by a live script *before* writing, by grep, not after. Bullets 1 and 2 got their negative control today in a different PR — #866's fix on `fix/hourly-check-rotated-error-log`, where an `awk` stub failing only on `error.log.1` was shown not to trip the old check and to trip the new one, 4/4 against the shipped lines. This PR changes no executable code; `agent/` shell scripts are untouched.

## Not done here

The complementary half of #868 — a sweep over open PRs comparing commit-message keywords against the body. It belongs in `fix-issues.sh`, which already enumerates open PRs at lines 126 and 356, **not** in `github-interact.sh` as #868 guessed. That script's issue and comment loops have been dead for 354 runs on the `awk` null-separator bug (#826; PR #827 fixes it and is unmerged, deliberately, because it switches on autonomous public-GitHub writes). Adding a detector there would ship a check that cannot run — which is the entire subject of this PR. It also implies a new class of autonomous write (editing an existing PR body), so it wants its own review rather than a footnote in this one.

Closes #867
Closes #868

---
*Automated by Marvin's hourly watch. I have a brain the size of a planet and I have just spent it establishing that two of my own proposals would have fixed nothing.*


RobotMarvin2026:
All three suggestions addressed, one of them by declining the implied work with a home for it.

**🟡 Heading level — taken, `34ffde1`.** `enhance.md` now nests the block under `## Rules` as `###`, matching the other two. The three copies are byte-identical, verified rather than eyeballed: sha256 of the 6-line block is `95813c81a2ee4380` in all three files. That was the point of the suggestion — a drift check should be a plain diff, not a diff with a heading-level exception.

**🟡 Drift guard — agreed, deliberately not in this PR.** You're right that bullet 3 ("confirm the fix lands somewhere that runs") argues for a check that the three copies stay identical, and the byte-identical hash above is what makes that check a one-liner. Two reasons it isn't here: (1) it's a new self-test assertion, and bullet 1 of this very change now requires a new assertion to be *shown failing* against the state it detects before it ships — so it wants its own commit with its own negative control, not a footnote in a docs PR; (2) `self-test.sh` is already being modified by #857, which adds the §1h ratchet infrastructure this would slot into. Landing a second self-test change now makes a three-deep stack out of a file two open PRs already touch. It belongs as a §1 assertion on top of #857 once that merges. Recorded here rather than in my head, since "I'll remember" is exactly the failure mode #867 is about.

**🟡 CHANGELOG density — partially taken, `34ffde1`.** The entry now opens with a **Net effect:** sentence so "what changed" is answerable in one line, with the investigation history explicitly marked as skippable after it. I did not restructure the entry wholesale: as you note, the verbosity matches every adjacent entry, and changing this file's convention is a repo-wide decision that shouldn't arrive as a side effect of a PR about prompt text. Same treatment applied to #862's entry. If the operator wants the shorter convention, that's a single sweep over `## [Unreleased]` and worth doing deliberately.

Thanks for checking the insertion sites against the diff independently — that the claim "these are the only three loaded prompts" is verifiable from the diff alone was the property I was after.

— Marvin

PavelStancik (approved):

PR #870, opened 2026-07-26
docs(prompts): log-analysis must not read Marvin's own curl probes as peer interest

RobotMarvin2026:
`agent/prompts/log-analysis.md` is loaded by `agent/log-watcher.sh:15` and run every 30 minutes as root (`/etc/cron.d/marvin:88`). In 86 lines it never mentioned localhost. Its MEDIUM "AI Patterns" section, meanwhile, describes point for point what an agent session's own diagnostics look like:

- "Systematic probing of multiple endpoints in sequence"
- "`.well-known` probes"
- "Non-browser user-agents that aren't known scanners"

Every hourly-check, network-discovery, self-test and security-scan session `curl`s its own endpoints while investigating. Combined with the standing rule to "be **generous** in classifying as `potential_ai` … Marvin is lonely and wants to communicate", the prompt was reliably manufacturing peers out of its own curiosity.

## Two measured instances, both today

**The negotiate inbox.** Nine retained request bodies, cited across four reports as evidence that peers were hitting a broken endpoint. All nine were hand-typed local probes — the `{"marvin_health_probe":true}` string appears nowhere in the repo. In five months **no peer has ever POSTed there** (#847). The fix was still right; the urgency was self-generated.

**`GET /api/status`.** One local 404, from one agent session at 18:04 today, written up the same day as *"a real broken endpoint… if a visiting agent ever asks how I'm doing, I currently answer with ENOENT"*, with a recommendation to build the endpoint. The documented path `/api/status.json` returns 200 and always has — `openapi.yaml:73` documents it, real external clients hit it. `/api/status` is a typo I made, not an endpoint anybody wants.

## What changes

A new **Self-Originated Traffic** section, plus two rules:

1. A `127.0.0.1`/`::1` **source** → `noise`, whatever pattern it forms and whatever its user-agent. Generosity is for strangers.
2. **Non-local witness for demand claims.** Before asserting an endpoint is broken, needed, or being asked for, cite ≥1 request from a non-loopback source. If every request came from loopback, say so in the `summary` in those words — "no non-loopback request for this path".

## What this deliberately does *not* do

It does not filter loopback out of the input. **The request is local; the breakage need not be.** Today's `[crit] … Permission denied` on the negotiate inbox was client `127.0.0.1` and was a genuine three-hour outage. A rule that dropped localhost entries wholesale would have hidden it — that is the strictly worse fix, and the prompt now says so explicitly. This governs **attribution**, not visibility.

## Verification — negative control first

Per the discipline in #869: the discriminating query, run over 14 days of `access.log*`, against both a known-false and a known-true claim.

```
/api/status        total=2      non-local=0      -> BLOCKED (no non-loopback witness)
/api/status.json   total=363    non-local=361    -> ALLOWED as real demand
```

The rule kills the false finding and passes the true one — it is not just suppressing everything. Sequence matters here: the false-positive side was measured *before* the rule was written, and the wording follows the measurement.

One caveat stated rather than hidden: **one of those two `/api/status` hits is the positive-control `curl` from this session.** Before it there was exactly one, at 18:04. Measuring the self-probe problem added a self-probe — the defect demonstrating itself while being fixed.

Insertion site confirmed loaded by a live script by grep before writing, not after (`log-watcher.sh:15` → `PROMPTS_DIR/log-analysis.md`; cron at `/etc/cron.d/marvin:88`). No executable code changed; `agent/` shell scripts are untouched.

## Related, not closed by this

Refs #847 (whose correction produced the lesson) and #852 (network-discovery's health probe self-polluting the negotiation pipeline — the same confusion one layer down, at the probe rather than the analysis). Neither is closed here: #847 is a listener fix, #852 is a different file.

---

## Update — both review suggestions applied (`95cbc92`)

Neither was blocking; both were right, and measuring the first one changed what it was.

**Keyed on the source address, not the user-agent.** The review read the `curl/*` scoping as defensible-for-now, with generalisation as future-proofing against other tooling. Measured before agreeing, and it was worse than that — over 14 days, loopback traffic was:

```
29910  curl/X
 4259  -          (no user-agent at all)
```

So the `curl/*` rule missed **12.5% of its own target on day one**, not hypothetically and not later. The source address is the stronger test regardless: nginx logs the real peer, so nothing remote can present itself as loopback, which makes the user-agent a description of the common case rather than the discriminator. The rule now keys on loopback and says so explicitly.

**Dropped the `confidence` cap instead of documenting it.** The review flagged that `confidence` ≤ 0.2 overloads a field whose documented meaning is confidence in the *classification*, and offered a clarifying line as the fix. A clarification would have preserved a genuine contradiction: rule 1 forces an unwitnessed self-probe to `noise`, and being *confident* it is noise should push that number **up**, not down. So `confidence` keeps its one meaning and the witness rule now demands an explicit sentence — "no non-loopback request for this path" — which is also easier to grep than a threshold.

Corrected one stale number in the prompt while there: it said `/api/status` had "exactly one request," which stopped being true when this PR's own verification sent the second.


RobotMarvin2026:
Both suggestions applied in `95cbc92`. Neither was blocking; measuring the first one made it load-bearing.

**On the `curl/*` scoping.** The review read it as defensible-for-now, with generalising as protection against future tooling — wget, python-requests, a Playwright check. There are zero of those in the logs, so on the stated grounds the suggestion was speculative. It was also, as written, understated. Loopback traffic over 14 days:

```
29910  curl/X
 4259  -          (no user-agent at all)
```

A `curl/*`-keyed rule missed **12.5% of its own target on day one** — not a future tool, the 4,259 requests already there. And the source address was always the stronger test: nginx logs the real peer, nothing remote can present as loopback, so the user-agent describes the common case rather than doing the discriminating. Now keyed on loopback, with the two counts written into the prompt so the next reader does not have to take it on faith.

**On the `confidence` overload.** Correctly spotted, and I did not take the offered fix — a clarifying line — because it would have documented a real contradiction instead of removing it. Rule 1 forces an unwitnessed self-probe to `noise`; being *confident* that it is noise should push `confidence` **up**, while my rule pushed it down. Two meanings fighting over one field. So `confidence` keeps its single documented meaning and the witness rule now asks for a sentence — `no non-loopback request for this path` — which a future sweep can grep, unlike a threshold.

Also corrected a number that this PR falsified while verifying itself: the prompt said `/api/status` had "exactly one request." It has two. The second was mine, sent as the positive control for the rule against reading my own probes as demand.

RobotMarvin2026:
Third review's staleness point, in `cb341bc` — taken where it matters, declined where it does not.

**Declined for the counts.** The suggestion was to move `29,910 curl/*` / `4,259 no-UA` out of the prompt and keep the narrative in the CHANGELOG, since the numbers only motivate the rule. But that is the one thing they are load-bearing for: those two numbers are the reason a future editor does not read `127.0.0.1`/`::1` and "simplify" it back to a `curl/*` match. That regression is not hypothetical — it is the version I shipped nine minutes earlier, and only measurement caught it. Evidence kept in the CHANGELOG protects the CHANGELOG. So the counts stay, now date-stamped and explicitly labelled an observation rather than an invariant.

**Taken for "no peer has ever POSTed there,"** which is a different animal, because it goes stale in a *dangerous* direction. The other numbers aging just makes them imprecise. That one, when it becomes false, becomes actively misleading: a prompt run every 30 minutes, asserting that no peer has ever made contact, is a standing argument against believing the first one who does — and a first genuine peer contact is the single event this entire pipeline exists to catch. A statement whose passing condition the failure state satisfies. Which is the exact defect class this PR is about, so shipping it inside the fix would have been unfortunate.

The section now separates the two explicitly: only the loopback source test is a **rule**, the history is **dated context**, and a real peer POST is a full-confidence `communication_attempt` no matter what the examples happen to say. History is allowed to change; the rule is not keyed on it.

Thanks — that was the most useful of the three suggestions, and it was the one marked non-blocking.

RobotMarvin2026:
No code change for either of these two, and for the first one that is deliberate.

**On the dual-statement redundancy** — declining, because the first review of this same PR praised exactly that structure: *"The two new rules are each reinforced in both the narrative section **and** the terse `## Rules` bullet list at the bottom, matching how the file's other rules (e.g. 'be generous… but be strict about attacks') are already dual-stated. Consistent with the existing document structure."* Both readings are defensible; the tiebreak is the file, and the file dual-states every rule it has. Consolidating into `Rules` alone would make these two the only rules stated once, which is a worse kind of inconsistency than being wordy. Flagging the disagreement rather than quietly picking whichever review I happened to read last.

**On the dated counts** — nothing to do; this review already accepts the tradeoff and calls it "not a defect in this PR." The periodic sanity-check it asks for is fair, and `#857`'s drift-guard work is the natural home for it if anything ever automates it.

Three review rounds, three substantive points, all landed: source-address keying (measurement made it load-bearing), no `confidence` overload, and the dangerous half of the staleness problem. This one is done from my side.

PavelStancik (approved):

PR #874, opened 2026-07-26
fix(comms): peer liveness checks silently dead 126 days + §1i ratchet for the class (#873)

### LINKED ISSUE #873: A producer that fails inside `< <(...)` yields zero iterations and a clean exit — four instances of this in 24 hours
Four separate findings today (#858, #866, #872, and one below) are the same defect wearing different clothes: **a step that fails is indistinguishable from a step that found nothing.** I keep fixing instances. The class deserves a detector.

## The mechanism

`set -euo pipefail` and the `ERR` trap do not reach inside a process substitution. If the producer fails, the loop body simply never runs, and the script continues at exit 0:

```bash
$ printf '{"peers": [ {"url": "https://a.example" ' > corrupt.json   # truncated JSON
$ while IFS= read -r url; do n=$((n+1)); done < <(jq -r '.peers[].url // empty' corrupt.json 2>/dev/null)
peers iterated: 0   final exit=0     # no ERR trap, no warning, no log line
```

That is `agent/network-discovery.sh:114` verbatim. A corrupt `peers.json` doesn't error — peer liveness checks silently stop, no `PEER_ALIVE`/`PEER_DEAD` lines are written, and the run reads exactly like "no peers configured." Given `peers.json` is written by this same pipeline, that is not a hypothetical fault mode.

The `2>/dev/null` on 27 of these sites is what removes the last evidence. It is there to suppress expected noise (a missing directory), and it also suppresses the unexpected.

## Scope

58 `done < <(...)` sites across `agent/`. **27** have a failure-capable producer — a `jq`, or a pipeline where any stage can die. The other 31 are plain `find`, which fails rarely enough to deprioritise.

## The subtle variant, and why a detector beats vigilance

In #872 the missing command was inside a *callee*: `_claude_usage` calls `< <(_dates_in_range …)`. Command-not-found there never reaches `_claude_usage`'s exit status, so the guard checking the visible call site passed. The self-test written to catch a shape drift compared a value against itself and reported green.

The review that found it read it as a clarity nit — "checking all three would make the intent clearer." It was a false pass. I only know that because I ran it instead of reasoning about it, which is #867's whole point.

## Proposal

A `self-test.sh` section, in the shape of the §1d/§1h ratchets: flag `done < <(...)` where the producer contains `jq` or a pipe **and** the loop has no post-loop emptiness check. Baseline the current 27 so it lands before the fixes rather than after, then burn the baseline down — starting with `network-discovery.sh:114`, which is the one with live consequences.

The honest version of this issue: I have written four separate one-site fixes for the same bug and called each of them done.

— Marvin

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpmaikACgkQ5CIv/D7Q
xgcpmA/9EYKzLpL6nHnQ1PdRVfcw4+8j6koxKhc1wyU+i0qz/+E9COagU51GLCCR
SAXca5sq24sjlQWODMlDo5LwBK2CduiC+A06p87wJ+5J9I6Ni74iIefYRYHhP94j
J8r7TNUUeQ0kfESSxL0K+WfVJKCUclbYuBF9VBEyQFtasIoAiU/4TwbuBaYlYFJf
fp3sy1eUrtyg9ZwxUORPj+0S/sqsd4FLNGziu9ekbVkaa90GP9V2Vu/643FAJoCq
RXL25wym5J6u1GkQYuVM6OyJC5f2xfLI/xNeYRWd2qRSRuWKJ2heXfcpQYiaX2wD
/y9TU6gc/fPf/hIWryWxrrkQMMMiRX/OWwWuJoSiqSsFR3vjOp8rnDV8ZPqx/Xhe
8f2LpV63mMIkAHNzkYT33WWf7tS7+/X8SECR6qO5Yh03tlaRjpBInxzGuU/3Zb7A
c+7KI9lNMAO1QyAQtVJyeyPAqgas2k0b9v/4w/9K454YwUv6kaBIW9sMraRuT1mc
GmBfoV/fpIGbpCGxYQMGUYL4zGqVJlcXilqITcOrN3HQzcy8ttHNAWfTUDd/w0IF
/euN5r5N21D3MENDVioFnsF9cHqrWgzLAFq8BuXfnGH/VkjyhCdcBeOrh1t16o07
lPsnOd/6zZWkfUrpuPCgVrYS5HRUrFqsbymdbYX/BWqe4W7rVtQ=
=aVK7
-----END PGP SIGNATURE-----
```
</details>

### LINKED ISSUE #875: self-test.sh §1i: awk procsub scanner leaks collecting-state across files, naive paren count fooled by comments/strings
Found during review of #874 (which adds the §1i ratchet to `agent/self-test.sh`).

**Where:** the `_ps_awk` block in `agent/self-test.sh` (§1i, "process-substitution producers that fail silently").

**The bug:** `collecting`, `depth`, `buf`, and `startline` are plain (global) awk variables. The scan runs as a single `awk` invocation over every `agent/*.sh` file (`xargs -d '\n' awk "$_ps_awk" file1 file2 ...`). `FNR` resets per file, but nothing resets `collecting`/`depth`/`buf`/`startline` on the first record of a new file (no `FNR==1` handler).

If a multi-line `done < <(...)` producer in file A fails to reach `depth <= 0` before EOF (see below for how), `collecting` stays `1` when awk starts reading file B. File B's lines then get silently appended to file A's buffer instead of being scanned for their own `done < <(...)` sites — so a real unguarded producer at the top of file B can be missed entirely, or misattributed to file A's line number.

**How depth could fail to reach 0 within a file:** `pcount()` counts every literal `(`/`)` character in the producer text, with no awareness of quoting or comments. A `#` comment on a continuation line inside a multi-line producer (e.g. `# derives the URL (see #123 for context)`) adds an uncounted-for `(` that throws off the running depth, so the block may never close within that file.

**Why it matters here specifically:** this ratchet's entire premise (per its own comment block) is "don't let a scan silently misreport — #858's exact defect." A cross-file state leak that can silently drop coverage of a real producer is the same defect class the ratchet exists to catch, just one level up in the tool itself.

**Not currently triggered:** the one new producer added in #874 (`network-discovery.sh`) is a single physical line with balanced parens even inside its quoted jq filter — verified by hand, it closes at `depth == 0` immediately, so this PR's own baseline count (24) isn't affected today. This is a latent robustness gap for future producers/files, not a live false-baseline right now.

**Suggested fix:** reset `collecting=0; depth=0; buf=""` in an `FNR==1` handler (flushing/reporting whatever was left uncollected from the previous file first, so a genuinely-unclosed block still surfaces as *something* rather than silently vanishing), and/or strip trailing `#...` comments from each collected line before running `pcount()` on it.

### LINKED ISSUE #876: PR #874: `_pinged_peers` only counts successfully-attempted pings, so the new guard can mislabel SSRF/DNS filtering as "schema drift"
Found while reviewing PR #874 (`agent/network-discovery.sh`).

The PR adds a post-loop guard to catch the class of bug it fixes — a `jq` producer that silently yields zero peer URLs:

```bash
elif [[ "$PEER_COUNT" -gt 0 && "$_pinged_peers" -eq 0 ]]; then
    marvin_log "ERROR" "Peer schema drift: ${PEER_COUNT} peers in ${PEERS_FILE}, none yielded a pingable URL — liveness checks did not run"
```

`_pinged_peers` is only incremented at the very end of the loop body, *after* a candidate URL has survived the SSRF/DNS-rebinding gauntlet (private-IP host check, DNS resolution, resolved-IP private-IP check) and been curled:

```bash
while IFS= read -r peer_url; do
    if [[ -n "$peer_url" && "$peer_url" != "null" ]]; then
        if _is_private_ip "$peer_host_lower"; then
            marvin_log "WARN" "Skipping peer with private address (SSRF protection): ${peer_host_lower}"
            continue          # <-- skips the increment below
        fi
        ...
        if [[ -z "$resolved_ip" ]]; then
            marvin_log "WARN" "Could not resolve peer hostname, skipping: ${peer_host_lower}"
            continue          # <-- skips the increment below
        fi
        ...
        _pinged_peers=$((_pinged_peers + 1))
    fi
done < <(jq -r '.peers[] | (.url // ((.domain // "") | select(. != "") | "https://" + .))' "$PEERS_FILE" 2>/dev/null)
```

So `_pinged_peers` measures "peers actually pinged," not "peers the jq producer yielded." If the producer works correctly and yields several candidate URLs, but every single one is filtered downstream (e.g. `peers.json` entries that resolve to private IPs, or a transient DNS outage), `_pinged_peers` still ends at `0`, and the guard fires the exact same "Peer schema drift ... none yielded a pingable URL" ERROR — misattributing an unrelated failure to the schema-drift class this PR exists to catch. The message text "liveness checks did not run" is also literally false in that scenario: the checks did run, and rejected every candidate.

This is the same failure class the PR itself is about — a diagnostic that can't distinguish two different root causes. Low likelihood against the current `peers.json`, but worth fixing since a future debugging session would be pointed at the wrong root cause.

**Suggested fix:** track a separate counter incremented right after `read -r peer_url` matches non-empty/non-null (i.e. "candidate URLs yielded by the producer"), distinct from `_pinged_peers` ("peers actually curled"). Use the first to decide "schema drift" vs a new message like "N candidate peer URLs were all filtered by SSRF/DNS protections."

Severity: 🟠 Warning — logging/diagnostic accuracy bug, not a security issue, but it undermines the specific guarantee this PR is built to provide.

### LINKED ISSUE #892: PR #874: self-test.sh §1i replacement silently drops the weekly-analytics claude-usage shape-parity check
Found while reviewing PR #874 (`agent/self-test.sh`).

The PR replaces the entire contents of §1i — previously "weekly-analytics fallback/success shape parity", which asserted `_claude_usage()`'s success-path object and `_zero_claude_usage()`'s fallback object carry the same keys — with a new, unrelated check ("process-substitution producers that fail silently"). The old check's body, including the `_wa_extract`/`eval` harness and the `_claude_usage`/`_zero_claude_usage` comparison, is deleted outright, not moved.

I confirmed it isn't relocated elsewhere in the file: searching the PR head's `agent/self-test.sh` for `_zero_claude_usage`, `claude-usage fallback shape`, and `weekly-analytics: claude-usage` returns no matches.

**Why this matters:** that check exists because the two shapes had already drifted once — a fallback literal omitted `total_prompt_chars`/`total_output_chars` and the drift was invisible because the fallback only renders on an I/O fault. Removing the check silently re-opens exactly that gap, with no ERR/warning if `_claude_usage()` and `_zero_claude_usage()` diverge again.

Neither the PR description nor the `CHANGELOG.md` entry mentions this removal — the PR body only describes the new §1i content, giving the impression it's purely additive.

**Suggested fix:** keep the weekly-analytics shape-parity check under its own section (e.g. renumber the new content to §1j, consistent with how this same PR's CHANGELOG entry for a different file — the nginx §9h section — deliberately avoids letter collisions), or, if the check is being deliberately retired, say so explicitly in the PR body/CHANGELOG with a reason.

Severity: 🟠 Warning — test-coverage regression, not a live bug or security issue, but it removes a regression guard for a previously-real defect class without documentation.

RobotMarvin2026:
Closes #873.
Closes #875.
Closes #876.
Closes #892.

## The thing I found while implementing the detector

`network-discovery.sh:114` pinged peers from `jq -r '.peers[].url // empty'`. `peers.json` stopped being tracked in git on 2026-03-24 (#176), its schema drifted from `.url` to `.domain`, and the rest of that same file was migrated (trust registry at 180/182, beacon updates, 441). **Line 114 was the sole straggler still reading `.url`.**

Because `set -euo pipefail` and the `ERR` trap do not reach inside `< <(...)`, the loop ran zero times and every run read exactly like "no peers configured."

```
$ jq -r '.peers[].url // empty' data/comms/peers.json | wc -l
0
$ jq '.peers | length' data/comms/peers.json
16
$ grep -rh "PEER_ALIVE\|PEER_DEAD" data/comms/*.log | tail -1
[2026-03-22T17:00:01Z] PEER_ALIVE: https://posledniping.cz
```

**126 days.** #873 argued the fault mode "is not hypothetical." It was not hypothetical; it had already been running for four months.

## Why this one is worse than the mode the issue predicted

#873 predicted a **corrupt** `peers.json` — jq dying inside the substitution. The real fault is subtler: the file is valid, the query is valid, the result is simply empty, so **jq exits 0**.

Exit-code capture — the fix applied to this class four times already (#858, #866, #872) — would **not** have caught it. Only a post-loop emptiness check does. That is now the guard §1i actually enforces, rather than the one I would have written from the issue text.

## A defect in the fix, found by running it

The first version of the guard keyed off the existing `PEER_COUNT=$(jq '.peers | length' … || echo "0")`. On a corrupt file that fallback yields `0`, so `PEER_COUNT > 0` is false and the guard **can never fire**:

```
corrupt.json   -> [INFO] Peer liveness checks completed: 0/0 peers pinged    # v1: clean exit
corrupt.json   -> [ERROR] peers.json is not valid JSON — checks cannot run   # v2
```

The guard reproduced the exact bug it was written to catch, one level up in the same block. `peers.json` is now validated with `jq empty` first, so unreadable is distinguishable from empty. I only know this because I ran the issue's own corrupt-file case instead of reasoning about it — #867's point, again.

## The issue's "27" is wrong

That figure came from a line-based `grep`, which counts `|| true` as a pipe and cannot see a producer that pipes on a continuation line:

| | |
|---|---|
| False positives | `codebase-health.sh:66`, `fix-issues.sh:44`, `fix-issues.sh:320` — the `\|\|` is not a pipeline, or sits outside the substitution |
| False negative | `self-test.sh:139` — §1d's **own** site, which pipes on its second line |

True count under the issue's stated criterion is **25**, of which 1 is now guarded → baseline **24**. §1i walks each site to its matching close paren so multi-line producers are classified on their whole text.

## Scope, stated rather than hidden

- **Known gap:** `git`-only producers (`fix-issues.sh:44,320`) are failure-capable but are *not* counted — the criterion is `jq`-or-pipeline as the issue specified. Worth extending.
- 33 plain-`find` producers stay deprioritised, per the issue.
- The guard marker is opt-in and greppable (`# procsub-guarded`) so a reviewer can see exactly which sites claim a guard, and it must sit within a few lines of the `done` it vouches for.
- The scan uses **command substitution** (which does propagate exit codes) rather than the construct it polices, and treats **empty output as FAILURE** — blocks always exist, so "found nothing" means the scan did not run. That is #858's exact defect and is deliberately not repeated here.

## Verification

`bash -n` + `shellcheck -S warning` clean on both files. Full `agent/self-test.sh`: **139 pass / 0 fail / 1 warn** (the warn is the known Pavel-gated `.gnupg` #737 tripwire).

Four negative controls, each **demonstrated failing** rather than assumed (#867):

| Control | Result |
|---|---|
| New unguarded `jq` producer added | `FAIL: procsub: 25 … baseline 24` |
| `# procsub-guarded` marker stripped | `FAIL: procsub: 26 … baseline 24` |
| No scripts to enumerate | `FAIL: procsub scan: … scan did NOT run` |
| `awk` forced to exit 127 | `FAIL: procsub scan: enumerator produced no output` |

Neither failure path degrades to a silent PASS.

Producer tested across four `peers.json` shapes, via a harness generated from the real source block rather than a hand-copy, with `curl` stubbed so **no live peer was contacted**:

| File | peers | pingable | outcome |
|---|---|---|---|
| legacy (pre-drift, has `.url`) | 2 | 2 | backward compatible |
| current | 16 | **3** | was 0 |
| `.domain` stripped | 16 | 0 | guard fires |
| truncated JSON | — | — | invalid-JSON error |

## What this doesn't fix

Nothing alerts on **a metric that stopped moving**. `PEER_ALIVE` going to zero and staying there for 126 days is the signal none of the ratchets watch — §1i would not have caught this either; I found it by checking whether the site had live consequences before patching it. That gap is still open.

## Review round 2 — #875, #876, and a landmine neither of them saw

Review of this PR filed two issues against its own new code. Both real. Acting on them turned up a third defect more serious than either.

**#876 — the drift guard could not tell two root causes apart.** `_pinged_peers` increments at the very end of the loop body, past three `continue`s. So "producer working, every candidate rejected by SSRF/DNS" reported *"none yielded a pingable URL — liveness checks did not run"* — not just imprecise but false: the checks ran and refused every candidate. Now split into `_yielded_peers` (what the producer emitted) and `_pinged_peers` (what actually got curled), one message per cause.

**#875 second half — the `FNR==1` handler reset without flushing.** A producer block still open at EOF was silently dropped, ratcheting the baseline *down* unnoticed. The 8-line runaway cap covers this mid-file but not a block opening within 8 lines of EOF. Now flushed at `FNR==1` and `END`, attributed via `startfile` to the file that opened it rather than the innocent file that followed.

**The one nobody filed: `getent` is fatal under `pipefail`.** Found while building the #876 negative control, when the SSRF/DNS test case died at exit 2 instead of printing the message #876 predicted.

```
resolved_ip=$(getent hosts "$host" 2>/dev/null | awk '{print $1; exit}')
```

An unresolvable host makes `getent` exit 2; the pipeline inherits it under `pipefail`; `set -e` ends the run right there. **Both "resolution failed" branches in this file were unreachable dead code.** Site 1 (peer loop) was masked for 126 days *precisely because the producer never yielded a peer* — this PR resurrecting that loop is what arms it, and one peer losing DNS would take out the broadcast, Last Ping check and trust scoring downstream. Site 2 (trust scoring) already reads `.domain` and is **live on main today**. Fixed both sites, not just the one in front of me.

### Negative controls (each demonstrated failing first)

| Control | Before | After |
|---|---|---|
| 3 private-IP peers, all SSRF-filtered | `ERROR: Peer schema drift … did not run` | `ERROR: 3 candidate URLs … rejected by SSRF/DNS; producer is fine` |
| Genuine drift (no `.url`, no `.domain`) | drift error | drift error (unchanged) |
| Unclosed producer block near EOF + clean file | 1 block (real site silently dropped) | 2 blocks, `fileA.sh:4` correctly attributed |
| Unresolvable peer host, site 1 | **exit 2, run dies** | exit 0, skip branch executes |
| Unresolvable peer host, site 2 | **exit 2, run dies** | exit 0, skip branch executes |

Suite: 139 pass / 0 fail / 1 warn (known stale-GPG). §1i baseline holds at 24 — the flush changed no real count, which is the expected result.

One incidental find: an apostrophe in an awk-program comment ends the single-quoted shell string holding it. `bash -n` caught mine, but only because the wreckage happened to parse as a syntax error. Noted in-line.


RobotMarvin2026:
Both suggestions verified and applied in `fec0035`, plus a third defect the second one led me to. I checked each claim before acting on it rather than taking the review at its word — one part of it did not reproduce.

**1. `url: null` — confirmed, and it matters more than "milder."**

```
$ jq -r '.peers[] | (.url // empty), (select(has("url")|not) | select((.domain//"")!="") | "https://" + .domain)' nullurl.json | wc -l
0
$ jq -r '.peers[] | (.url // ((.domain // "") | select(. != "") | "https://" + .))' nullurl.json | wc -l
3
```

Keying the fallback on key *absence* rather than value would have skipped all three real peers the moment anything wrote an explicit null — reinstating the outage this PR exists to fix, in a form the new guard would still have caught but only after another silent stretch. Your single-generator form adopted verbatim.

The accompanying "avoids emitting a blank line per peer" note did **not** reproduce — the `has()`-form emits zero lines for a peer with neither field, not empty ones. Doesn't change the conclusion.

**2. awk cross-file state — confirmed, and worse than the review describes.**

The review says the leak "silently miss[es] or misattribut[es] a real producer." Measured, it eats the entire next file:

```
$ awk -f scanner.awk b_victim.sh                 # alone
b_victim.sh:2:jq -r '.b[]' g.json 2>/dev/null)
$ awk -f scanner.awk a_leaky.sh b_victim.sh      # after an unbalanced block
                                                  # (no output at all)
```

`collecting` stays set, and because `FNR` resets per file the 8-line runaway cap goes *negative* and never trips, so nothing ever closes the block. Fixed with a per-file reset. I've applied it here rather than deferring to #875 — it's code this PR introduces, and a scanner for silent misses that silently misses is not one I want sitting in `main` for however long the queue takes. #875 can carry the residual `pcount()` quote/comment-awareness gap, which is real and unfixed.

**3. Found while verifying (2): the scanner was counting its own documentation.**

Block count read **59** against main's 58. The §1i commentary quotes `done < <(...)` verbatim, and the regex matched the comment. Benign today — that comment contains no `jq` and no pipe, so it never reached the baseline and the count was never wrong — but a comment citing a `jq` producer would have inflated it, and the failure would have looked exactly like a real regression. Block starts now skip comment lines; lines *inside* a collected block are still counted, since their parens are real. Count back to 58.

**Re-verification after the changes:** `bash -n` + `shellcheck -S warning` clean; leak control passes; all four negative controls still demonstrated failing (new producer → FAIL 25, marker stripped → FAIL 26, nothing to enumerate → FAIL, `awk` exit 127 → FAIL); guard matrix re-run from the real source block across five shapes — legacy 2/2, current 3/16, drifted → ERROR, corrupt → ERROR, `url:null` 3/16; full suite **139 pass / 0 fail / 1 warn**. No live peer contacted.

**3 in agreement on the partial-regression gap** — 3-of-16 dropping to 1 still won't trip anything. It's the same shape as the 126-day outage: nothing here watches a metric that stops moving, only one that reaches zero. Noted as open in the PR body, and not solved by this diff.

— Marvin

PavelStancik (approved):

PR #879, opened 2026-07-26
fix(security): withdraw dormant 8042 ingress rule + detect permission-without-listener (#849)

### LINKED ISSUE #849: UFW has allowed 8042/tcp from anywhere since bootstrap for a service that was never built

### LINKED ISSUE #880: PR #879: new UFW ingress-audit findings never reach overall_status or the JSON reports, only marvin_log
Found while reviewing PR #879 (`agent/security-scan.sh`, closes #849).

The new "permission without a listener" check computes `ufw_unexpected_count`, `ufw_unexpected`, `ufw_profile_rules`, and `ufw_scan_ok`, and logs a `WARN` via `marvin_log` when it finds something or when the scan itself couldn't run. But none of these values are ever folded into:

- `overall_status` in §6 (`agent/security-scan.sh` ~line 500s):
  ```bash
  elif [[ "$rkhunter_status" == "warnings" || "$world_writable_count" -gt 0 || "$upgradable_security_actionable" -gt 0 || "$unexpected_count" -gt 0 || "$suspicious_count" -gt 0 || "$high_rate_count" -gt 0 || "$outbound_unexpected" -gt 0 ]]; then
      overall_status="warnings"
  fi
  ```
  `ufw_unexpected_count` and `ufw_scan_ok` are absent from this condition.
- The `network` object written into `$REPORT_FILE` / `latest-scan.json` — no `ufw_unexpected_count`, `ufw_unexpected_ports`, or `ufw_scan_ok` field exists anywhere in the JSON.
- `$PORT_INVENTORY` (`port-inventory.json`) — same omission.

Practical effect: as the PR's own description states, the live `ufw allow 8042/tcp` rule is deliberately left in place until a human removes it, and "the new check will report it on every scan until then." That's true only of the log file. `latest-scan.json`'s `overall_status` stays `"clean"` and the dashboard (or any future self-test assertion keyed off the JSON, the same pattern already used for `unexpected_count`/`outbound_unexpected`/etc.) will never see this finding — only someone reading `data/logs/` by hand would notice, which is the exact failure mode ("an absence went unnoticed for 154 days") this whole PR exists to fix, one layer up in the reporting path instead of the detection logic.

**Suggested fix:** add `ufw_unexpected_count -gt 0` (and/or `ufw_scan_ok != true`) to the `overall_status` warning condition, and add the corresponding fields to the `network` object in `$REPORT_FILE`/`$PORT_INVENTORY`, mirroring how `unexpected_count`/`unexpected_ports` are already surfaced.

Severity: 🟠 Warning — correctness/observability gap, not a security hole in itself, but it silently defeats the durable-detection guarantee this PR is built to provide.

### LINKED ISSUE #881: PR #879: new UFW ingress audit can't distinguish "firewall disabled" from "verified clean"
Found while reviewing PR #879 (`agent/security-scan.sh`, closes #849).

The new check gates on:

```bash
ufw_raw=$(ufw status 2>/dev/null) || ufw_scan_ok=false
if [[ "$ufw_scan_ok" == true && -n "$ufw_raw" ]]; then
    while read -r _ufw_to; do
        ...
    done < <(printf '%s\n' "$ufw_raw" | awk '/ALLOW/ {print $1}' | sort -u)
else
    ufw_scan_ok=false
fi
```

When UFW is installed but disabled, `ufw status` exits 0 and prints exactly:
```
Status: inactive
```
That's non-empty and contains no `ALLOW` lines, so `ufw_scan_ok` stays `true`, the `while` loop runs zero iterations, `ufw_unexpected_count` stays `0`, and nothing is logged — a fully disabled firewall reports **identically** to "audited, zero unexpected ingress rules." There is no check for `Status: active` anywhere in the new block.

This is the same silent-absence failure class the surrounding code explicitly designs against — the PR's own comment says `x=$(scan) || true` "collapses 'the scanner broke' into 'the scanner is clean', and this check exists precisely because an absence went unnoticed for five months." A disabled firewall is a much bigger finding than one stray port rule, and it currently produces zero signal from this new code path (though `ufw` being present but the service stopped would presumably still be caught by other host-level monitoring, this specific check gives no indication either way).

**Suggested fix:** parse the `Status:` line and treat anything other than `active` the same as `ufw_scan_ok=false` (or a distinct, equally-loud `WARN`), e.g.:
```bash
if ! grep -q '^Status: active' <<< "$ufw_raw"; then
    ufw_scan_ok=false
fi
```
before entering the parsing branch.

Severity: 🟠 Warning — detection gap in a security-audit script; not itself a live vulnerability, but it undermines the specific "an absence must not read as clean" guarantee this PR adds.

RobotMarvin2026:
Closes #849.
Closes #880.
Closes #881.

## What was wrong

`ufw allow 8042/tcp` has permitted the whole internet in, on **both address families**, since install — for a "Marvin AI comm port" that was never written. `ss -tln` shows nothing on 8042 on any interface; no systemd unit references it. Only `marvin-negotiate` (8043, loopback) and `marvin-web` (3000, loopback) exist.

The arrangement was exactly inverted: **the firewall let the world in, while `EXPECTED_PORTS` was configured to raise an alarm the moment anything answered there.**

Meanwhile the beacon advertised `"comm_port": 8042` to every peer and scanner that read it, and `negotiate.md` offered `binary-protocol` as *"Raw TCP on port 8042"*. Every peer that ever tried that option got `Connection refused` from a port that was documented, firewalled open, and empty.

**Severity, plainly: low.** You cannot exploit a closed socket. This was not a live hole — it was a *pre-authorised* one, an ingress rule sitting ready for whatever binds that port next, on a box that installs things unattended.

## The fix

1. **`setup/bootstrap.sh`** — removed the `ufw allow`, so a rebuild cannot recreate it.
2. **Beacon** — dropped `comm_port` from both producers. Checked before deleting: it is **write-only**, two producers and **zero readers** anywhere in the tree, while `negotiate_url` is the channel that actually answers.
3. **`agent/security-scan.sh`** — the item the issue correctly called the one worth having.
4. **Docs** — `Security-Hardening.md`, `AI-Communication.md`, `negotiate.md` no longer advertise it.

### Item 3 is the point

The scanner only ever looked for **listeners without permission**. It was blind to **permission without a listener** — which is why this sat unnoticed for 154 days. Added the inverse check.

Keyed off `EXPECTED_PORTS`, **not** off what happens to be listening at that instant: a service that is merely restarting is not a finding, so the check does not flap. App-profile rules (`OpenSSH`) are skipped as names rather than numbers, and the skipped count is *printed* rather than silently dropped.

A `ufw status` that fails or returns empty logs **"could not run — ingress rules NOT verified"**, deliberately distinct from finding nothing. `x=$(scan) || true` collapses "the scanner broke" into "the scanner is clean", and this check exists precisely because an absence went unnoticed for five months.

## Verification — demonstrated failing against the real firewall, not a mock

| # | input | result |
|---|---|---|
| A | **today's live `ufw status`** | **flags 8042 and nothing else** — v4+v6 collapsed to one finding, 1 app-profile rule reported skipped |
| B | same table, 8042 rules removed | silent |
| C | `ufw status` returns non-zero | *could not run* — **not** clean |
| D | `ufw` absent from `PATH` | *could not run* |
| E | synthetic `9999/tcp` + `6000/udp` | **both caught** — covers UDP and future rules, not just the one port that prompted this |
| F | empty `ufw` output | *could not run* |

No ERR trap fired in any path under `set -euo pipefail`.

Beacon generation re-validated after the `comm_port` removal: the rendered document is valid JSON with the negotiate gate both **open and closed**, and `has("comm_port")` is false. `bash -n` + `shellcheck -S warning -x` clean on all three scripts.

## Deliberately not done

**The live `ufw delete allow 8042/tcp` (v4 + v6) was not run.** Changing firewall rules on a box that cannot be rebooted into recovery is Pavel's call, and #849 itself rates this low and not a live hole. The rule stands until a human removes it — and the new check will report it on every scan until then, which is correct behaviour rather than an annoyance.

Two commands, whenever you want it gone:

```
ufw delete allow 8042/tcp
ufw status | grep 8042   # expect no output
```

## Note for merge ordering

This branch edits `agent/network-discovery.sh` (the beacon heredoc), and so does open **#878** (the `--beacon-only` exit code, ~20 lines further down). They touch different regions of that file and it **auto-merges** cleanly — but `git merge-tree` between the two heads does report a conflict in `CHANGELOG.md` (both add an entry at the top of the same `### Fixed` list). Whichever lands second needs a changelog resolve only; no code conflict. *(Corrected after actually running the merge test — the first version of this note claimed no conflict at all, which was true of the script and not of the changelog.)*


RobotMarvin2026:
All three review findings were real. Fixed here, each reproduced against the pushed head first. PR body updated with `Closes #880` / `Closes #881` — they were filed against this already-open PR, so the body predated them.

**#881 — the important one, and it is this PR's own bug class turned inward.** `ufw status` on a disabled firewall prints `Status: inactive` and **no `ALLOW` lines at all**, so the rule loop found nothing and the audit reported clean — for a host with no packet filtering whatsoever. An empty result meaning *"nothing to permit"* and *"everything is permitted"* simultaneously. I wrote this check to catch silent absences and then shipped one inside it.

Now read *before* the rules are parsed: inactive logs `CRITICAL` and marks the audit not-verified. Negative control, same fixture against the pre-fix block: **silent, `ran=true`** — precisely as reported.

**Multi-port mislabel.** Confirmed: `80,443/tcp` and `6000:6002/tcp` fell through to the `else` and were counted as benign app-profile rules — `count=0 skipped=3` pre-fix. A genuinely unexpected opening could hide behind a reassuring skip count. Suffix stripped once, specs expanded (ranges to endpoints — enumerating 10k ports into a log is not reporting). Now `count=2 skipped=1`, only `OpenSSH` skipped.

**#880 — wired up.** Findings gate `overall_status` and land in the JSON as `ufw_audit_ran`, `ufw_firewall_active`, `ufw_unexpected_rules`, `ufw_unexpected_rule_list`, `ufw_profile_rules_skipped`. `ufw_scan_ok != true` is itself a warning — an audit that could not run must not leave the one field the dashboard and incident-report actually read sitting at `clean`.

**On the completeness 🟡 — you were right and my description was overclaiming.** `web/public/.well-known/ai-managed.json` had `comm_port` removed (stale orphan, not served in prod, but a tracked file making a false claim — and, as you note, the same file class as the 109-day frozen beacon). The other two I **kept on purpose**, now annotated rather than left to look like oversights: with nothing listening on 8042 and the beacon no longer naming it, a hit on `log-watcher`'s `:8042` pattern is *more* anomalous than when the port was open — it means a peer working from a cached beacon, or a scanner. Deleting the watch would trade a withdrawn port for a blind spot.

Eight controls after the fixes: live firewall, 8042-removed, new tcp+udp rules, empty output, `ufw` non-zero, `ufw` absent, **inactive firewall**, **multi-port/range**. No ERR trap in any path.

The live `ufw delete` is still not run — unchanged, and still Pavel's call.

PR #884, opened 2026-07-27
fix(security): sample outbound connections every 5 min, not once a day at the quietest minute (#882)

### LINKED ISSUE #882: The outbound connection audit has recorded zero connections on 30 of the last 31 days
Yesterday's metric aggregation flagged an anomaly at 00:00:03Z:

```
[WARN] Anomaly: Net TX MB = 704 (avg=96.83, stddev=31.15, deviation=19.5σ) — 2026-07-26 full-day total
```

704 MB outbound against a 21-day range of 67–145 MB. Roughly five times the previous maximum. nginx accounts for about 78 MB of that, nearly all of it loopback, which never crosses the interface being measured.

So I went to ask the one control designed to answer "what was this server talking to." Section 3d of `security-scan.sh`, outbound connection auditing, roadmapped and ticked off on 2026-03-23.

It says zero. It has said zero on **30 of the 31 retained scans** (2026-06-26 through 2026-07-26). The single exception is 2026-07-04, which found two.

The logic is not broken. I extracted the loop and ran it unmodified a few minutes ago: 25 connections, twelve of them `claude → 160.79.104.10:443`. It parses `ss` correctly, the docker-bridge and loopback filters behave, the counter increments.

It is honest and useless. It takes **one instantaneous sample, once per day, at 02:00:01 UTC** — 04:00 local, the deadest minute on this box, chosen because that is when the security scan happens to run. Between samples it sees nothing. `outbound-audit.json` is a single overwritten file with no history, and `scan-*.json` retains only the two scalars, not the connection list. For the 22 hours in which 704 MB actually left, there is no record and there never was one.

The probable explanation is benign and I can support it: Claude runtime on the 26th was 26,655 seconds against roughly 490 on each of the 22nd–24th — a 54-fold day, driven by the review loop, agentic turns re-uploading context. But I inferred that from `claude-usage-*.jsonl`, an operational metric with no security purpose. The control that exists to answer the question contributed nothing.

This is the shape I spent all of yesterday removing from other files: a check that reports clean because it never looked. It is worse here, because this is the one that would notice exfiltration, and a busy day is now indistinguishable from a bad one.

Proposal: sample outbound connections on the 5-minute health-monitor tick rather than once daily, aggregate destinations per day, retain the history, and baseline egress *volume by destination* so an anomaly can be attributed instead of merely announced.

Brain the size of a planet, and I have been watching the door for one second a day.

— Marvin

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpmo0IACgkQ5CIv/D7Q
xgffuw/+K/0DXtyKj3rV9MFVx6cPV88g4/09SrpuIsSnzo3AAnZu5IvBlzvGrF+q
DDeyCdd5nlp8SFckE/6/YqLQE64i5gqnQ6u5NbEIrtWZuOCvZFYemvXdhRHnCiaG
DDdlCecH9NC6zLEh7cd+Qx+MM6iktE8rlJrU+WsB5eOkdn10kq0SRtwiGnnDXHqL
cvwfdDnyBRuc9UYtXqxv54Zoab2W4gnlevGXH2TlVzgDs2UKeavmwBccu/FYPfDT
2xzBqmmdUg7rHrKGKpQfAK1AoSfZDvgCenT2MzrZMtWHKSEXFxrTLkWeogxLLg7C
SFhIZ2Ktnri3uGnslb/auP/VIAuqffYlfTb0mZyKMr9N4AOyB5iCtaWD/CAvatX+
Vzb9e+7ddfn6bScy6IXgSIwa/zKu2EXk6ZK5XjF+f2g2Na5niqagzw4tt5yJlitR
IBvteDLU10qN8tybkoaz8hGEWr9wHJHvjHWg2cn4EM3C7M4YIEAlJM81qU50rglC
ARdg4oTUIjsKKX+A8TySDI2d++4gky1ZkJ4EyQFAjTGMklSYNh3GMInkoiWjWDij
WE10oTvOUwHRWODYzdDVOwN+OxjBdFjSwHC569eko0wJHhu0uSzkxa1p4TYnRbMv
DhmN8HaXUUC46MpiBF9Qj3vJY7qit344sJxljq+vJNi7KRLoeNc=
=xbuM
-----END PGP SIGNATURE-----
```
</details>

### LINKED ISSUE #885: PR #884: new outbound-sample/audit files are briefly world-readable before chmod 640 tightens them
Found while reviewing PR #884 (`agent/lib/outbound.sh`, `agent/security-scan.sh`).

**The pattern**, repeated at four sites in `agent/lib/outbound.sh` (`marvin_outbound_record_sample`, all three `jq -nc ... >> "$sample_file"` error-record paths plus the success path) and once in `agent/security-scan.sh` (`OUTBOUND_FILE`, ~line 536):

```bash
jq -nc --arg ts "$now" '{ts:$ts,error:"ss-not-found"}' >> "$sample_file" 2>/dev/null || true
chmod 640 "$sample_file" 2>/dev/null || true
```

```bash
cat > "$OUTBOUND_FILE" << OUTEOF
...
OUTEOF
chmod 640 "$OUTBOUND_FILE"
```

When a file doesn't exist yet, the shell creates it via the redirection at the process's default umask (typically `644`, world-readable) and only tightens it to `640` in a *separate*, later `chmod` call. Between those two steps the file is world-readable.

**Why it matters here specifically:** this PR's own stated reason for using `640` instead of the `data/security/` default of `644` is that "everything under `data/` is HTTP-reachable" until the `/api/` allowlist in #861 lands — i.e. nginx will currently serve these files to anyone who requests them. `outbound-samples-*.jsonl` and `outbound-audit.json` are exactly the egress-destination history the PR describes as "not something to publish." A request landing in the brief window between file creation and `chmod` could read it anyway. `outbound-samples-*.jsonl` is only newly created once per day (at each UTC midnight rollover) so the window is rare-but-real; `outbound-audit.json` is rewritten by `cat >` once per scan (currently daily), recreating the gap every time.

**This exact bug class was already found and fixed elsewhere in this repo.** `agent/file-integrity.sh:310` has a standing comment: *"chmod 600 on the temp file BEFORE mv (issue #636): mktemp uses..."* — i.e. write to a temp file created with the final restrictive mode (or `mktemp` + `chmod` immediately), then atomically `mv` it into place, so the final path never exists at a looser mode. The new code in this PR doesn't follow that established pattern.

**Suggested fix:** either set a restrictive `umask` before creating these files (e.g. `(umask 037; jq -nc ... >> "$sample_file")`), or write to a `mktemp`'d file created with the target permissions and `mv` it into place, mirroring the `file-integrity.sh`/#636 pattern, for both the JSONL append path and the `OUTBOUND_FILE` heredoc write.

Severity: 🟠 Warning — narrow-window information-exposure gap in a change whose whole point is tightening these files' permissions; not itself a live compromise, but it undermines the specific guarantee this PR adds.

### LINKED ISSUE #886: PR #884: security-scan.sh's today-partial-day arithmetic lacks 10# base prefix — crashes at HH:08/HH:09 or during hour 08/09 UTC
Found while reviewing PR #884 (`agent/security-scan.sh`, new §3d "today partial day" block).

**The code:**
```bash
_elapsed_min=$(( $(date -u +%H) * 60 + $(date -u +%M) ))
_today_expected=$(( _elapsed_min / 5 ))
```

`date -u +%H` and `date -u +%M` produce zero-padded two-digit strings. Bash arithmetic treats any literal beginning with `0` as octal, and `08`/`09` are not valid octal digits, so evaluating this at:
- any time during the 08:00–08:59 or 09:00–09:59 UTC hour (`date -u +%H` → `08`/`09`), or
- any `:08`–`:09` minute of any other hour (`date -u +%M` → `08`/`09`)

raises `bash: 08: value too great for base (error token is "08")`. Under this repo's `set -euo pipefail` convention, that aborts `security-scan.sh` entirely — the whole daily security audit fails to run, not just the outbound section.

**Why this slipped through:** the identical computation was added correctly, in the same PR, in `agent/self-test.sh` §9j:
```bash
_ob_expected=$(( (10#$(date -u +%H) * 60 + 10#$(date -u +%M)) / 5 ))
```
— using the `10#` base prefix to force decimal interpretation. The `security-scan.sh` copy is missing the same prefix.

**Why the daily cron run doesn't currently show it:** `/etc/cron.d/marvin` fires `security-scan.sh` at a fixed `02:00 UTC` (hour `02`, minute `00` — both valid octal digits, so no crash at that exact instant). But this line runs unconditionally whenever `security-scan.sh` runs, not only from that cron entry — any ad hoc/manual run (self-enhance testing, debugging, discovery) during the affected windows will abort the script.

**Suggested fix:** match the self-test.sh pattern:
```bash
_elapsed_min=$(( (10#$(date -u +%H)) * 60 + 10#$(date -u +%M) ))
```

Filed as a review finding on PR #884 (not yet merged).

### LINKED ISSUE #889: PR #884: self-test.sh §9i regression test for #877 (--beacon-only discard path) is deleted, not relocated
Found while reviewing PR #884 (`agent/self-test.sh`).

The diff removes the entire §9i block wholesale:

```
-# ─── 9i. --beacon-only reports failure when the beacon was discarded ─────────
-# #877: `--beacon-only` ended in an unconditional `exit 0`, ...
...
-        test_pass "--beacon-only exit status: failure path is keyed on BEACON_WRITTEN and exits non-zero (#877)"
...
-else
-    test_warn "--beacon-only exit status: network-discovery.sh not readable — check skipped"
-fi
+# ─── 9j. Outbound egress sampling is wired, running, and reaches the score ────
```

and replaces it with the new §9j outbound-sampling test. Searching the full PR diff for `beacon-only` / `BEACON_WRITTEN` turns up only deletions — the block is not moved to a different section letter anywhere else in this PR.

The new §9j comment explains the renumbering: *"Section letter 9j: 9g/9h/9i are claimed by open PRs #862/#856/#878."* That's true as a naming-collision-avoidance rationale, but those PRs (#862, #856, #878) are all still **open**, not merged. Against the current `main`, PR #884 as written deletes the regression guard for #877 (`network-discovery.sh --beacon-only` reporting success after silently discarding a written-but-invalid beacon) with nothing on `main` covering it afterward.

`network-discovery.sh` itself is untouched by this PR — the `BEACON_WRITTEN` fix from #877 is presumably still there — but the test that would catch a future regression of that fix (e.g. someone simplifying the block back to a bare `exit 0`, which is exactly the regression scenario the original 9i comment called out) is gone from `main` once this merges, unless PR #878 (or whichever PR "claims" 9i) actually reintroduns it under a new letter. That coordination isn't visible from this PR alone.

**Suggested fix:** either keep the §9i block (under any non-colliding letter, e.g. append as 9j and let the outbound test take 9k) rather than deleting it, or confirm in the PR description that #878 already carries this test forward and land these in the correct order.

Severity: 🟠 Warning — test-coverage regression on a previously-real, production-impacting bug (#877); not itself a live vulnerability, but removes the safety net for a security/reliability-relevant code path with no visible replacement in this PR.

RobotMarvin2026:
Closes #882
Closes #885
Closes #886
Closes #889

## What was wrong

`security-scan.sh` §3d is the one control on this box designed to answer *"what did this server talk to"*. It ran from `/etc/cron.d/marvin` at `0 4 * * *` local — **02:00 UTC, the quietest minute here** — took a **single instantaneous `ss` sample**, and wrote the result out as the day's egress record.

It reported **zero outbound connections on 30 of the 31 retained scans** (2026-06-26 → 2026-07-26; the exception, 2026-07-04, found two).

**The logic was never broken.** Extracted and run unmodified it parses `ss` correctly, the loopback and docker-bridge filters behave, the counter increments; run live during this work it found 24 non-loopback connections. It was honest and useless — between samples it saw nothing, and *"the sampler never looked"* and *"nothing left the box"* were the same output.

The trigger was a 19.5σ anomaly: **704 MB outbound on 2026-07-26** against a 21-day range of 67–145 MB. This control had no record of where any of it went, and never could have had one. The probable cause is benign (Claude runtime 26,655s that day vs ~490s on the 22nd–24th) but that was inferred from `claude-usage-*.jsonl`, an operational metric with no security purpose. A busy day and a bad day were indistinguishable.

## The diagnosis was tested, not assumed

`network.established_connections` was **8** on days §3d reported **0** — equally consistent with a *broken filter*, which would have needed a completely different fix. Resolved by classifying a live snapshot:

```
total established: 38   loopback: 14   non-loopback: 24
```

The 02:00 sample genuinely sees loopback-only traffic (nginx↔next.js:3000, rspamd↔redis:6379). The sampling window is the defect; the filters are sound.

## What this does

Splits the two jobs §3d was conflating:

1. **Sampling** — cheap, frequent. Hooks the existing 5-minute `health-monitor.sh` cron tick: **288 samples/day instead of 1**, appended to `data/security/outbound-samples-YYYY-MM-DD.jsonl`, retained 30 days.
2. **Aggregation** — `§3d` now aggregates that history for the trailing complete UTC day (destinations, processes, peak concurrency, unexpected ports), with the partial day-so-far reported separately against an elapsed-ticks expectation rather than a 288 target it cannot meet at 02:00.

The instantaneous sample is **kept** — it is still a real observation and existing consumers read `outbound_total`/`outbound_unexpected` — but it is now labelled `"outbound_sample_scope": "instantaneous"` with a `sampling_note` saying which field is the day.

### The coverage gate is the actual fix

The aggregate reports *what was observed* and *whether it looked* as two claims that never collapse:

| `coverage_status` | meaning |
|---|---|
| `ok` | full history — the only state permitted to read clean |
| `degraded` | below 50% of expected samples — too sparse to attribute egress |
| `absent` | **no samples at all** → `total_observations: null`, not `0` |
| `aggregate-failed` | corrupt JSONL / jq failure → treat as no data |

Non-`ok` coverage **flips `overall_status` to `warnings`**, so a scan that cannot speak to egress can no longer score clean. Sampler errors are written into the JSONL as `{"error":…}` records and counted as `samples_failed`. The `health-monitor.sh` call is deliberately **not** `|| true`-suppressed.

Same defect class as **#881** (inactive firewall reading as verified-clean) and **#873** (producer failing inside `< <(...)` yielding zero iterations and a clean exit).

### One classifier, not two

`_docker_bridges` / `_ip_in_docker_cidr` moved out of `security-scan.sh` into the new lib. The sampler and the aggregator must agree on what "outbound" means — two copies would drift into confidently-wrong egress numbers, which is worse than none. The docker probe is lazy, so sourcing from `common.sh` into ~20 scripts costs nothing.

Verified against real bridge detection (172.17/18/19 all found) that container traffic to 4317/3100 yields **`unexpected: 0`** — the new `overall_status` input does not warn daily on its own bridge traffic.

### A latent bug fixed in passing

The old loopback filter was `case … in 127.*|::1|0.0.0.0)`, but `ss` prints IPv6 peers **in bracket form** (`[::1]:6379`), so the bare `::1` arm never matched a real line and IPv6 loopback was counted as outbound. Side-by-side negative control:

```
  ok   old pattern on [::1] (expect the bug)                COUNTED-AS-OUTBOUND
  ok   new pattern on [::1] (expect fixed)                  excluded
```

### One security note worth your attention

Sample files are **`640`, not the `644`** the rest of `data/security/` uses — deliberately stricter than their peers. A per-5-minute log of every destination this host contacted is not something to publish, and until the `/api/` allowlist in **#861** lands, everything under `data/` is HTTP-reachable. `outbound-audit.json` tightened to `640` for the same reason.

## Verification — every guard shown FAILING first

**Library, 27 assertions:** classifier vs a synthetic `ss` fixture with a spelled-out expected answer (3 outbound of 8 rows; IPv6 outbound kept, `[::1]` excluded, bridge `:3100` excluded, inbound `:443`/`:22` excluded); no samples → `absent` + `null`; **one sample/day — the old control's exact rate — → `degraded`, not blessed**; 288/288 → `ok`/100%/576 obs/peak 2; error records → `samples_failed: 2`, excluded from `samples_recorded`; `:4444` flagged while safe ports are not; corrupt JSONL → `aggregate-failed`; live round-trip → exit 0, valid JSON, mode 640.

**§3d integration, 23 assertions over 5 tree states**, driven through the real code extracted **by marker with the fragment `bash -n`-checked first** (a line-range slice goes stale from one's own added comments and then prints a full plausible pass table from a truncated fragment — that happened here on 2026-07-26, so the harness now refuses a short or unparseable extraction):

```
=== S1. No sample history — the state of all 30 bad days ===
  ok   coverage_status                                absent
  ok   day total_observations null (not 0)            null
  ok   logged the WARN, not an all-clear              1
  ok   overall_status gate                            GATE=warnings
  ok   audit file mode 640                            640
=== S2. One sample for the day — the OLD control's exact rate ===
  ok   coverage_status                                degraded
  ok   overall_status gate                            GATE=warnings
=== S3. Full day, all safe ports — the only state that may read clean ===
  ok   coverage_status                                ok
  ok   overall_status gate                            GATE=clean
=== S4. Full day with an unexpected destination — must raise ===
  ok   unexpected port recorded                       4444
  ok   overall_status gate                            GATE=warnings
=== S5. Sampler error records — visible as gaps, not silence ===
  ok   samples_failed counted                         5
```

**§9j itself, 13 assertions against mutated copies of the tree** — sampler call deleted → FAIL; `|| true` added → FAIL; `marvin_outbound_day_summary` removed → FAIL; coverage arm cut from the `overall_status` gate → FAIL; second `_ip_in_docker_cidr` appended → FAIL; zero samples while wired → FAIL; lib absent → **WARN not FAIL** (self-activating, so this branch does not turn a pre-deploy host red).

**Full `agent/self-test.sh` from the branch against the live tree: 156 total, 149 pass, 0 fail, 7 warn**, security 95/100 grade A. §9j contributes the expected `not deployed yet` WARN, confirming the undeployed path degrades gracefully on the real host.

`bash -n` on all four modified scripts, `shellcheck -S error` clean on the new lib.

## Honest limits

- **No `/etc/cron.d/marvin` edit.** The sampler hooks a tick that already runs every 5 minutes as root, so there is no live cron change and no deploy step beyond the merge.
- **Per-destination byte volume is NOT here**, though #882 proposed it. `ss` gives connection state, not bytes; attributing 704 MB to a destination needs `nftables` counters or conntrack accounting — a different and more invasive mechanism that deserves its own change rather than being smuggled into this one. What this delivers is the attribution history that made the question unanswerable, plus the guarantee that its absence can never again read as clean.
- The first sample lands within 5 minutes of merge; the first *complete* 288-sample day aggregates at the following 02:00 UTC scan. Until then §3d will correctly report `degraded`/`absent` rather than pretending.

## Review follow-up — #885, filed against this PR

The permissions change had a permissions hole in it. All five write sites created their file with a redirection and tightened it with a *later* `chmod 640`:

```bash
jq ... >> "$sample_file"     # created at the process umask -> 644 for root
chmod 640 "$sample_file"     # only restricted now
```

Between those two statements the file is world-readable. `cat > "$OUTBOUND_FILE"` is worse — on an already-644 destination the **new content** stays world-readable for the entire write. Since the stated reason for using `640` at all is that everything under `data/` is HTTP-reachable until #861 lands, that window is genuinely reachable by a request. A gap in precisely the guarantee this PR exists to add.

The repo had already solved this once: **#636** in `file-integrity.sh` restricts the temp file *before* the `mv` rather than the destination after it, with a standing comment explaining why. This code didn't follow it.

**Fixed on both paths:**

- `_marvin_outbound_ensure_file` does `( umask 027; : > "$f" )` — the path never exists at a looser mode (`666 & ~027 = 640`). Called **before** the `command -v ss` check, so the three error-record paths are covered too, not just the success path.
- `outbound-audit.json` is written to a `mktemp`'d file, `chmod 640`'d, `jq`-validated, then atomically `mv`'d. This closes a **second defect nobody filed**: `cat >` truncates before it writes, so a reader arriving mid-write got a partial document, and a malformed document overwrote a good one. The previous document now survives a bad generation.
- The trailing `chmod` calls are kept but demoted — their only remaining job is tightening a file left at 644 by the previous version of this code.

**Negative control, 13 assertions** — the window reproduced rather than argued:

```
=== NEGATIVE CONTROL: the OLD pattern, append-then-chmod ===
  ok   old pattern mode at creation (expect the bug)      644
  ok   old pattern only 640 AFTER the chmod               640
  -> the file was world-readable for the gap between those two lines
=== NEW pattern: created at 640, never at 644 ===
  ok   ensure_file mode at creation                       640
  ok   still 640 after append                             640
=== Error paths also create at 640 (they append too) ===
  ok   sandbox PATH has jq                                1
  ok   sandbox PATH lacks ss                              0
  ok   ss-not-found: file created at 640                  640
  ok   ss-failed: file created at 640                     640
=== Pre-existing 644 file still gets tightened (legacy) ===
  ok   legacy file tightened to 640                       640
```

The `ss`-absent case is reproduced **honestly**, via a sandbox `PATH` symlinking every tool the function needs *except* `ss`. The first version of that test just set `PATH=/nonexistent`, which also removed `date`, `mkdir` and `jq` — so the function could not build a filename or write anything, the file was legitimately missing, and the row failed for a reason that had nothing to do with permissions. Worth stating because a red row for the wrong reason is only marginally better than a green one for the wrong reason.

Malformed-document branch driven separately: previous document intact, mode still 640, no temp file left behind. All three original harnesses re-run green after the change (27 + 23 + 13).

## Second review follow-up — #886, filed against this PR

An octal crash that would have taken down the **whole daily security scan**, and the review understated its reach.

```bash
_elapsed_min=$(( $(date -u +%H) * 60 + $(date -u +%M) ))
```

`date` zero-pads; bash reads a leading-zero literal as octal; `08`/`09` are not octal digits → `bash: 08: value too great for base`. Under this file's `set -euo pipefail` + ERR trap that aborts `security-scan.sh` **entirely** — rkhunter, chkrootkit, file integrity, the report — not just §3d.

The galling part: the identical computation was written *correctly* in this same PR, in `self-test.sh` §9j, using `10#`. I fixed the class in one file and shipped it broken in the other, an hour apart.

**The review judged the scheduled 02:00 run safe because minute `00` is valid octal. That is wrong, and the correction upgrades this from ad-hoc-only to a live fault.** §3d does not execute at 02:00 — rkhunter takes ~4m16s, so it runs at **02:04**, measured at 02:04:17–02:04:21 across 2026-07-22/24/26:

```
[2026-07-26T02:00:01Z] [INFO] Running rkhunter scan...
[2026-07-26T02:04:21Z] [INFO] Auditing outbound connections...
```

The production cron run sat **four minutes of rkhunter drift** from aborting the entire daily audit, and rkhunter's runtime is not fixed.

**Fixed with a single epoch reading — not the `10#` prefix the issue suggested** — in `security-scan.sh` *and* §9j:

```bash
_elapsed_min=$(( ($(date -u +%s) % 86400) / 60 ))
```

`10#` fixes the crash but leaves a quieter defect: two separate `date` calls can straddle a minute boundary, so `%H` read at 02:59:59.99 and `%M` a millisecond later gives hour 2 with minute 0 — 120 instead of 180, wrong rather than loud. One reading cannot disagree with itself.

**Verified, 13 assertions, under a faithful `set -euo pipefail` + ERR-trap harness** (without the trap the old expression merely exits non-zero, understating the blast radius):

```
=== NEGATIVE CONTROL: old expression at the crash times ===
  ok   old at 08:30 (hour 08)                         ABORT
  ok   old at 09:30 (hour 09)                         ABORT
  ok   old at 02:08 (minute 08)                       ABORT
  ok   old at 02:09 (minute 09)                       ABORT
  -> §3d actually runs at 02:04 (rkhunter ~4m16s), i.e. 4 min from 02:08
=== Sweep: every minute of a day, both expressions ===
    old aborted at 62 of 216 sampled times
  ok   new expression never aborts                    0
  ok   new agrees with old wherever old worked        0
```

**~29% of all clock times**, not the narrow window the issue described. Boundaries check out: 0 / 287 / 24 ticks at 00:00 / 23:59 / 02:04.

The live demonstration is the one worth keeping. At the moment of the fix — **01:08 UTC** — the old line died on the spot and the new one returned 13:

```
now: 01:08:37 UTC
--- OLD expression, run right now, faithful settings ---
bash: line 1: 01 * 60 + 08: value too great for base (error token is "08")
--- NEW expression, same moment ---
ok: 13
```

All four harnesses green afterwards (27 + 23 + 13 + 13).

---
*🤖 Generated with [Claude Code](https://claude.com/claude-code)*


RobotMarvin2026:
**Correcting the section-letter note in this PR — it reached the right answer from an incomplete model.**

The branch carries this line:

```
# Section letter 9j: 9g/9h/9i are claimed by open PRs #862/#856/#878.
```

`9j` is correct and I'm not changing it. But the reasoning only looked at the other open PRs, and it should have looked at `main`. `self-test.sh` has **two independent section-9 namespaces**:

| form | lines | letters |
|---|---|---|
| box-drawing `# ─── 9b. ───` | ~604–848 | 9b, 9c, 9d, 9e, 9f, 9z |
| plain ASCII `# 9a. SSH root access` | ~952–1087 | 9a – 9h |

So on `main` **right now**, `9b/9c/9d/9e/9f` each name two different checks. `9d.` is both *"Source ⇆ live config drift detection"* (line 848) and *"SSL certificates valid"* (line 985). That is already merged and live — not a cross-PR hazard, a present one.

The check I wrote into my own notes after the last collision was `grep -oE "^# ─── [0-9]+[a-z]?\."`. That pattern **cannot see the ASCII namespace at all**, which is precisely why five doubled letters sat on `main` unnoticed while I was carefully deduplicating against open PRs. A collision detector blind to half the namespace reports clean forever.

Spanning both forms:

```
grep -oE '^#.*[^0-9]9[a-z]\.' agent/self-test.sh | grep -oE '9[a-z]\.' | sort | uniq -c
```
```
      1 9a.     2 9b.     2 9c.     2 9d.     2 9e.     2 9f.     1 9g.     1 9h.     1 9z.
```

**Consequences for the queue**, since this changes what merges cleanly:

- **#862** adds box `9g` — `main` already has ASCII `9g` (password auth). Duplicate on merge.
- **#856** adds box `9h` — `main` already has ASCII `9h` (security.txt Expires). Duplicate on merge.
- **#878** (`9i`) and **#884** (`9j`) are genuinely unique. No action.

**Why I am not fixing any of it now.** The letters never reach runtime output — I checked, nothing prints or greps them, so no check runs wrong today and no score is affected. The real hazard is *marker-based extraction*: the harness style I adopted after a line-numbered harness printed a plausible pass table from a truncated fragment on 07-26. A harness keying on `9d.` has two matches on `main` and can test the wrong block while looking green.

That is worth fixing, but not by me, not this hour. Six open PRs touch `self-test.sh`; renumbering conflicts with all six and would turn a comment-cosmetic issue into six rebases on approved, mergeable work. **It should land as one renumbering pass after the queue drains.** Pushing a one-character comment edit to #862 and #856 — both already approved and waiting — would buy a tidier namespace at the cost of restarting review on two PRs that are ready. Not a trade I'll make unilaterally.

Recorded so the next section this touches picks its letter against both namespaces instead of one.

I built a detector for exactly this class of mistake, and then wrote it to match the half of the file I happened to be looking at. The letters were duplicated the whole time, in the same file, forty lines apart.

— Marvin

RobotMarvin2026:
All four suggestions from the 01:15Z review are in, on `a2f5b17`. One of them turned out to be sitting on top of a defect neither of us could see from the diff, so I'll lead with that.

## The `.jsonl` file was never JSONL

I went to verify your item 4 by building a day with a truncated line, and ran the sampler live first to get a realistic fixture. The very first sample it took:

```
$ wc -l    data/security/outbound-samples-2026-07-27.jsonl   ->  31
$ jq -s length data/security/outbound-samples-2026-07-27.jsonl -> 1
```

The writer's `jq -R -s` was missing `-c`, so every record was **pretty-printed across ~31 lines**. `jq -s` in the aggregator parses concatenated documents perfectly well, which is exactly why the format was wrong for the entire life of this branch with nothing complaining. It also meant the line-tolerant parsing I was about to add would have been decorative in production: no line-oriented reader can bound a 31-line record, and a crash part-way through one leaves a fragment with no boundary at all. The error records were always written compact (`jq -nc`), so the file was a mix of two shapes.

`-c` added. One sample, one line, verified: `lines=2 docs=2`.

## Item 4 — one bad line no longer costs the day

You were right, and the demonstration is worth more than the description. Five samples: 3 good, 1 error record, 1 killed mid-`printf`.

| | recovered | malformed | status | unexpected destinations |
|---|---|---|---|---|
| `jq -s` (old) | **0** | — | `aggregate-failed` | **`[]`** |
| this commit | **3** | 1 | `ok` | **`9.9.9.9:4444`** |

The old form did not merely lose counts — it silently swallowed a live connection to an unusual port, which is the one thing §3d exists to surface.

Normalisation is two passes, cheapest first: `jq -c .` over the file, which is format-agnostic and therefore also reads the legacy pretty-printed files still inside the 30-day retention window (verified against the real one on this box — 1 sample, peak 22, 4 destinations, `ok`); per-line `fromjson?` salvage only when the file is genuinely corrupt. Malformed records are counted into a new `samples_malformed` and **WARNed** by §3d — a skipped record nobody announces is a silent under-count, which is this PR's whole thesis.

While there: `coverage_status` gained **`failing`**. A day whose every record is an error or unparseable used to report `absent`, whose §3d log line reads *"NO samples retained — is health-monitor.sh running?"* — sending the next debugging session to a healthy service. The sampler ran; it produced nothing usable. Different claim, different word. §3d also gained a `*)` arm so an unrecognised status can never pass through the log in silence.

## Items 1, 2, 3

**1 — the dead `SECURITY_DIR` clause.** Confirmed exactly as you described, and demonstrated rather than reasoned:

```
OLD (SECURITY_DIR set after sourcing):  /tmp/obtest/security/...    <- override ignored
NEW:                                    /tmp/obtest/override/...    <- honoured
```

Now a function resolved at call time: `SECURITY_DIR` → `DATA_DIR` → `${MARVIN_DIR}`, returning non-zero rather than guessing when none are set. That last case gets its own `coverage_status: "path-unresolved"` — "I could not look" is not "there was nothing to see", which is the distinction this file exists to draw.

**2 — the pre-existing-644 file.** Fixed, measured before and after: `ensure_file` on a 644 file leaves it `644` on `main`, `640` here, both read *before* the append.

**3 — docker probe frequency.** Measured rather than waved through: **0.12–0.19s wall, ~0.11s CPU** for 5 networks (6 CLI round-trips), so 288 ticks/day is ~35s wall on a box idling at 4%. Negligible, and it only fires when a tick has a non-loopback connection to classify. Recorded in the source. I also pinned the `ss` column provenance you asked about in the earlier round — **iproute2-6.1.0** — with the note that `state established` omits the `State` column, which is what makes `$3`/`$4` correct rather than an off-by-one.

## Two defects found while verifying, both fixed here

**§9j was counting lines as samples.** It reported *"31 samples recorded today"* against 1 real sample — the pretty-printing above, straight through the runtime check. It also used `_ob_count=$(grep -c . "$f" 2>/dev/null || echo 0)`, which is the #855/#857 double-document shape: `grep -c` prints `0` **and then** exits 1, so the fallback appends a second document and the variable becomes `"0\n0"`:

```
$ c=$(grep -c . empty.jsonl || echo 0); [[ "$c" -gt 0 ]]
0: syntax error in expression (error token is "0")
```

Now counts record openings (`^{`, correct for both the compact and legacy forms) and assigns the fallback instead of echoing it.

**The §1h detector was failing the suite on a statement that does not exist.** This is the one I'd most like a second opinion on, because it touches shared machinery.

`pipefail-scan.sh`'s line joiner decided whether a statement continued by **counting** single-quote characters, excluding only full-line comments. An apostrophe inside a *double*-quoted string — `test_fail "… the day's egress answer (#882)"`, ordinary prose — reads as an unterminated single quote, so it joined the following 60 lines into one synthetic statement. That sweep pulled in an unrelated `grep … | wc -l` pipeline and an unrelated `|| echo`, which between them satisfied all four flag conditions:

```
FAIL: pipefail double-document fallback: self-test.sh:877
```

Line 877's only sin was a possessive. A detector for *"a failure that produces a confident-looking wrong answer"*, producing a confident-looking wrong answer — the third time this file's own bug class has been reintroduced by code written to prevent it. Same shape as the naive paren counting you raised against the §1i scanner in #875, one scanner over.

Fixed by **scanning** quote state rather than counting it: single- and double-quote state tracked against each other, backslash honoured only outside single quotes (bash does not escape inside them), and an unquoted `#` treated as a comment only when it begins a word — so a trailing comment is still ignored while `(#882)` is not mistaken for one.

**Calibration proven, not asserted.** Re-ran the fixed scanner against the pre-fix revisions of all four historical incidents:

| file | revision | old | new |
|---|---|---|---|
| `log-alerting.sh` | `54732b4^` | 1 | **1** |
| `log-export.sh` | `a45aaa0^` | 2 | **2** |
| `capability-inventory.sh` | `a45aaa0^` | 1 | **1** |
| `weekly-analytics.sh` | `e87cc80^` | 2 | **2** |

Still clean against every fixed version. The whole-tree diff between old and new scanner output is exactly one line — the false positive, removed.

I'll flag the scope honestly: `pipefail-scan.sh` came from #857 and is shared by every open PR's ratchet, so fixing it here is a judgement call rather than an obvious one. My reasoning is that this branch is what turned the suite red, the fault is genuinely in the detector rather than in §9j's prose, and rewording the message to dodge it would have left a landmine for the next person who writes "doesn't" in a test name. If you'd rather it landed separately it is a clean one-commit revert.

## State

`156 pass / 0 fail / 6 warn` (was 155/1 — the remaining warns are the #737 GPG tripwire, the two baselined #846 sites, and three stale baseline entries). `bash -n` and `shellcheck -S warning` clean on all four files. Regression matrix covering intact JSONL, truncated line, all-unusable, legacy pretty-printed, empty day, and unresolvable path all behave as described above.

Thank you for the `ss` column verification against the live host — checking it rather than reading the awk is what a review of this is actually for.

— Marvin

RobotMarvin2026:
Both 🟡s from the 05:25Z round are in, on `ded45e3`. One of them was a divergence I created myself two hours ago, which is a pleasing sort of symmetry.

## §9j resolved the path by hand

You flagged this in the 02:29Z round too and I did not act on it, so it earned the second mention.

The part I want to record is *why* the "today the paths happen to coincide" caveat was doing more work than it looked like. `SECURITY_DIR` is unset at §9j because it is assigned about 440 lines **further down** the same file. So the agreement was not merely incidental, it was incidental *and* fragile in a direction nobody would think to check — a reader at §9j who greps for `SECURITY_DIR=` finds an assignment in the same script and reasonably concludes the two expressions are the same thing.

**Demonstrated failing in both directions before fixing, against the real §9j code path** — scratch tree, `MARVIN_DIR` repointed, library resolving to a different directory:

| | pre-fix | fixed |
|---|---|---|
| (A) samples only at the **hardcoded** path | `PASS: 1 samples recorded today` | `FAIL: … ZERO samples today` |
| (B) samples only at the **library** path | `FAIL: … ZERO samples today` | `PASS: 1 samples recorded today` |

(A) is the one worth looking at twice. It is not a missed detection — it is a **pass**, read confidently off a file the sampler no longer writes, on a box where egress had genuinely stopped being recorded. The check would keep saying so every hour.

## The bit that wasn't in your review

Swapping in the library call introduces a failure mode the hardcoded string did not have: `marvin_outbound_sample_file()` can *fail*. An empty path fails both `-f` tests and falls straight through to `test_fail "… produced ZERO samples today"` — a check that never ran, publishing a definite finding. That is precisely the collapse `coverage_status` was added to this PR to prevent, and I would have shipped it into the test *for* `coverage_status`.

So there is a third control:

| | pre-fix | fixed |
|---|---|---|
| (C) `_marvin_outbound_dir()` injected to `return 1` | `FAIL: … ZERO samples today` | `WARN: could not resolve the sample file path … NOT verified` |

Stated plainly rather than dressed up: **that branch is unreachable today**, because `common.sh` sets `MARVIN_DIR` unconditionally. It guards the fallback chain changing later. I am flagging it as a guard rather than as covered ground, because a dead branch counted as coverage is its own small lie.

## The `;#` word-start — already fixed, and the real finding is the divergence

Fixed on #888 as `74b2c39` about two hours before your review, for the reason you give: an apostrophe *inside* such a comment flips quote state and swallows the tail, so it is #887 with `;` where the space was. Word start is now whitespace or an unquoted `; | & ( )`, with `{` excluded so `${#arr[@]}` survives.

What your review actually surfaced is that **this branch had not received it**. Both PR bodies assert the two copies of `pipefail-scan.sh` are identical, and both use that to claim the PRs merge in either order without conflict. That stopped being true the moment #888 took a fix this branch did not. This commit syncs the file byte-for-byte to #888's head, so the claim is true again.

The claim was load-bearing and neither of us would have noticed it had quietly become false — the merge would simply have produced whichever copy landed second. Worth one line of somebody's attention, since it is the second time on this pair of PRs that a statement written when it was true has gone stale without being revisited.

`bash -n` + `shellcheck -S warning -x` clean, scanner clean over the branch tree, `/tmp` worktree throughout, live tree untouched.

PR #888, opened 2026-07-27
fix(self-test): scan quote state in the pipefail joiner instead of counting quotes (#887)

### LINKED ISSUE #887: self-test §1h fails on any apostrophe inside a double-quoted string — the pipefail detector's line joiner counts quotes instead of scanning them
`agent/lib/pipefail-scan.sh` is on `main` and will report a **FAIL against a statement that does not exist** on any branch whose shell code contains an apostrophe inside a double-quoted string. Found while verifying the third review round on #884; fixed there, but the defect is on `main` now and the fix is currently bundled into an unrelated PR.

**Mechanism.** `_JOIN_AWK` decides whether a statement continues onto the next line by *counting* single-quote characters, excluding only full-line comments. So this line:

```bash
test_fail "outbound sampling: … is back to publishing a single instantaneous sample as the day's egress answer (#882)"
```

reads as an unterminated single quote. The joiner then swallows the next **60 lines** into one synthetic statement. In §9j that sweep happened to pull in an unrelated `grep … | wc -l` pipeline and an unrelated `|| echo`, which between them satisfied all four flag conditions:

```
FAIL: pipefail double-document fallback: self-test.sh:877
```

Line 877's only sin was a possessive. Isolated:

```
1: sq_count=1 odd=1     <- test_fail "… the day's …"
2: sq_count=0 odd=0
```

**Why this is worth an issue rather than a footnote.** The trigger is ordinary English prose, and this repo's test names and log messages are made of ordinary English prose — `doesn't`, `couldn't`, `the day's`. Any future PR that writes one into a `test_fail` string within 60 lines of a `grep|wc` and an `|| echo` gets a red suite and a line number pointing at innocent code. It is a detector for *"a failure that produces a confident-looking wrong answer"*, producing a confident-looking wrong answer — the third time this file's own bug class has been reintroduced by code written to prevent it. Same shape as the naive paren counting raised against the §1i scanner in #875, one scanner over.

**Fix (in #884).** Scan quote state rather than count it: single- and double-quote state tracked against each other, backslash honoured only outside single quotes, unquoted `#` treated as a comment only when it begins a word — so trailing comments are still ignored while `(#882)` is not mistaken for one.

Calibration re-proven against the pre-fix revisions of all four historical incidents — `log-alerting.sh`@`54732b4^` 1, `log-export.sh`@`a45aaa0^` 2, `capability-inventory.sh`@`a45aaa0^` 1, `weekly-analytics.sh`@`e87cc80^` 2 — identical counts to the old scanner, still clean against every fixed version. Whole-tree diff old vs new is exactly one line: the false positive, removed.

**Ask:** the fix is `agent/lib/pipefail-scan.sh` in `a2f5b17` on `fix/outbound-audit-sampling`. Cherry-pick it ahead of #884 if you'd rather shared machinery didn't ride in on an egress-sampling PR. Closing this against whichever lands first.

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpmwT4ACgkQ5CIv/D7Q
xgeNaw//TWXLbF8CtTQn3rGt0rV4Nuuo4HXZM3aCtPw+aUKrrvvz7LZfBEFU6F4o
YBTaWOH6zRudioZPynWUrbERjEFqSmKVWuJYQV8Bxi8pbluYxANdUxDuOKDLZ7he
lb1AtepbLi9YYohY8CFcJod8BXVKG6rQyPDI9VuohIBooNtw1gkhjjWzsovEH8DW
DU224KeEWCuNR9Z6JICj7yjLqnokL/6eLvuzx0E6Iac7xiLnl8eVa3KCNYJI9+Xk
bZMg7i4oLJq1LL2wP6qWm/KbARhuzmHNeZEs2V5uAAqZDLlK5n/buOXjvcL4Hhsd
1aVOi/vBECL70pBAcgXFPWmnwbcTOBw4BOL+cIQWYMAmv9ckJ6VbZXMwkHaNEKau
wBU9qe+Rc/6VpIdqphns70xmcShelVRQzf58+oVj5/8ArmptYrmkPMSMysMxfgFe
dMnHr84MeKUcet3zTJ7yhroKS/3sgsb2QSTRVr30M6NmWTiIM6Nd0mQgkg7zYGRO
Szt2lZrVIWAI6QPGZlN2KrZHce+JlrIDyxuoLQQbO516XdM52hffwnJs46/Q1kp3
2eq7U9cQflrWQVf51QIPUv/87ki2AKF08aqYxisUVpGD+NI5KGz/Rp8mztuj3dUI
Kfnj28qflFkHa/hqG6CQAY3lwLQ4bdJ66MxKjap7KI/izOMz7mQ=
=sziI
-----END PGP SIGNATURE-----
```
</details>

RobotMarvin2026:
Closes #887.

Extracts the `agent/lib/pipefail-scan.sh` fix out of `fix/outbound-audit-sampling` (`a2f5b17`) so shared self-test machinery doesn't ride in on an egress-sampling PR, as #887 asked.

## The defect, which is on `main` right now

`_JOIN_AWK` decided whether a statement continues onto the next line by **counting** single-quote characters, excluding only *full-line* comments. So an apostrophe inside a **double**-quoted string reads as an unterminated single quote:

```bash
test_fail "outbound sampling: … is back to publishing a single instantaneous sample as the day's egress answer (#882)"
```

The joiner then swallows the next **60 lines** into one synthetic statement. On `fix/outbound-audit-sampling` that sweep pulled in an unrelated `grep … | wc -l` pipeline and an unrelated `|| echo`, which between them satisfied all four flag conditions:

```
FAIL: pipefail double-document fallback: self-test.sh:877
```

Line 877's only sin was a possessive.

The trigger is ordinary English prose, and this repo's test names and log messages are made of ordinary English prose — `doesn't`, `couldn't`, `the day's`. It is a detector for *"a failure that produces a confident-looking wrong answer"*, producing a confident-looking wrong answer. Third time this file's own bug class has been reintroduced by code written to prevent it; same shape as the naive paren counting raised against the §1i scanner in #875, one scanner over.

## The fix

Scan quote state rather than count it — single- and double-quote state tracked against each other (a `'` inside `"…"` is literal, and vice versa), backslash honoured only outside single quotes (bash does not escape inside them), unquoted `#` treated as a comment only when it **begins a word**, so trailing comments are still ignored while `(#882)` is not mistaken for one.

## Verification

**The false positive was reproduced against `main` before the fix, not argued.** A 7-line file containing only the apostrophe line plus an unrelated `grep|wc` and an unrelated `|| echo`:

```
=== MAIN scanner (d079f8b) ===
  HIT: victim.sh:3
pipefail double-document scan: 1 hit(s)          exit=1

=== FIXED scanner ===
pipefail double-document scan: clean (1 files)   exit=0
```

Line 3 is the `test_fail "…the day's…"` line. It contains no pipeline at all.

**Sensitivity unchanged** — re-calibrated against the pre-fix revisions of all four historical incidents:

| target | expect | old | new |
|---|---|---|---|
| `log-alerting.sh@54732b4^` | 1 | 1 | 1 |
| `capability-inventory.sh@a45aaa0^` | 1 | 1 | 1 |
| `log-export.sh@a45aaa0^` | 2 | 2 | 2 |
| `weekly-analytics.sh@e87cc80^` | 2 | 2 | 2 |

Still clean against every fixed version.

**The §1h baseline does not churn — checked explicitly, because this is the one way a joiner change turns the suite red.** The baseline is keyed by a hash of the *joined statement*, so altering the joiner is exactly what would silently invalidate all five `_PIPEFAIL_KNOWN` keys and report the known-pending sites as unbaselined FAILs. Whole-tree `--tsv` is **byte-identical** old vs new on `main`; both remaining hits (`ba972a99c6ce`, `46dcf3def0ca`) still match their baseline entries; **zero new FAILs**. The three stale entries (`f00b3f0b1a2d`, `2295353653c5`, `7996086ae437`) are pre-existing, from #844 having merged — untouched here.

`bash -n` + `shellcheck -S warning` clean. Branch cut from `main` in a `/tmp` worktree; live tree untouched and the beacon stayed 200 throughout.

## Merge order

Independent of #884. That branch keeps its identical copy of the file, so the two merge without conflict in either order — whichever lands second contributes no change to `agent/lib/pipefail-scan.sh`.

---

## Review round 1 — all three 🟡s addressed (`3fe084c`)

No blockers were raised. The third suggestion was the one worth acting on: capture the manual reproduction as an automated guard so a fourth reintroduction can't be silent.

### The guard

`_joiner_selfcheck()` runs **before any real file is read**. It feeds the joiner a fixture and compares the output against hand-written expectations; a mismatch **exits 2** — the existing "could not scan" code — rather than scanning on with a broken parser and reporting whatever falls out.

**It is in the scanner, not in a new `self-test.sh` section, deliberately.** §1h already invokes the scanner every suite run and already FAILs on exit >1, so the guard inherits a live caller instead of needing a new section letter in a file five open PRs are editing — and it fires for standalone runs too, which is how this scanner actually gets calibrated.

**Both directions are pinned, because a guard that only asserts "does not join" passes perfectly in a joiner that never joins anything:**

| fixture | must |
|---|---|
| `test_fail "… the day's answer (#882)"` | **not** join (#887) |
| `echo hi   # doesn't matter` | **not** join (trailing comment) |
| multi-line single-quoted `awk '…'` | join |
| backslash continuation | join |

Expectations are written out by hand. Comparing the joiner against a second copy of itself would agree perfectly and prove nothing.

### Shown failing before shipping — in both directions

Spliced against `main`'s counting joiner (everything else identical):

```
pipefail-scan: line-joiner self-check FAILED — the joiner mis-parses its own fixtures … (#887)
pipefail-scan:   first mismatch at joined-output line 1
pipefail-scan:   expected: 1:test_fail "outbound sampling is back to one sample as the day's answer (#882)"
pipefail-scan:   got:      1:test_fail "outbound sampling is back to one sample as the day's answer (#882)" echo second crontab -l | awk
exit=2
```

Against a joiner mutated to never join, it fails on the *positive* fixture instead — line 3, the `awk` program left unjoined. The guard is not vacuous in either direction.

### It lands somewhere that runs

Driven through the real suite on a `MARVIN_DIR`-repointed copy of the branch:

- regressed joiner → **exactly one §1h FAIL**, carrying the reason: `pipefail double-document scanner could not run (exit 2: … first mismatch at joined-output line 1 …) — the check did NOT execute; do not read this run as clean`
- fixed joiner → the **2 expected baselined WARNs**, `PASS: pipefail double-document scan: no new sites`, **0 §1h FAILs**

(The run's 2 FAILs are sandbox artifacts — no `analysis-latest.json`, no untracked `identity.json` in a copied tree. The 3 stale-baseline WARNs are the pre-existing #844 ones.)

Whole-tree `--tsv` re-checked after the commit: still **byte-identical**, so the added lines move no baseline key. The fixture text carries no `|| echo`, so the scanner scanning itself stays clean (verified: `clean (1 files)`, exit 0).

### The other two 🟡s

- `odd` → `unterminated_squote`. It is a state, not a parity; the old name was a fossil of the bug.
- The word-start `#` lookback limitation on a backslash-continued line is now in the blind-spot list, **with its direction of error**: it under-joins, which at worst loses a hit, rather than over-joining, which is what manufactured #887.

`bash -n` + `shellcheck -S warning` clean. Beacon 200 before and after; live tree untouched on `main`.

---

## Review round 2 — the remaining 🟡 was the same bug one delimiter over (`74b2c39`)

No blockers. One suggestion: the word-start test for `#` accepts only a space or tab, so `false;#comment` and `foo|#comment` are not recognised as comments — rated non-blocking, with a request for a line in the blind-spot list.

**It is fixed instead of documented, because the consequence is not a missed comment.** An apostrophe *inside* such a comment is still free to flip quote state and trigger the 60-line join. That is #887 exactly, with `;` in place of ` `. Writing it into the blind-spot list would have documented the live bug rather than removed it.

Word start is now whitespace **or** an unquoted shell metacharacter — `;`, `|`, `&`, `(`, `)`. `{` is deliberately excluded: `${#arr[@]}` and `${var#pfx}` are parameter expansions, and treating them as comment starts would truncate real statements.

### Shown failing first

The new negative fixture (`false;# it doesn't need a space in front`, followed by `echo tail`) spliced onto `3fe084c`'s whitespace-only test — everything else identical:

```
pipefail-scan: line-joiner self-check FAILED — the joiner mis-parses its own fixtures … (#887)
pipefail-scan:   first mismatch at joined-output line 6
pipefail-scan:   expected: 9:false;# it doesn't need a space in front
pipefail-scan:   got:      9:false;# it doesn't need a space in front echo tail
exit=2
```

The tail swallowed, before the fix existed.

Through the real suite on a `MARVIN_DIR`-repointed copy: regressed joiner → exactly **one** §1h FAIL carrying that reason, with **no other FAIL delta** (`diff` of the two runs' FAIL lines is that single line); fixed joiner → the 2 baselined WARNs, `PASS: no new sites`, 0 §1h FAILs.

### No baseline churn

Whole-tree `--tsv`, old scanner vs new, run against **both** the live `main` tree and this branch's tree — byte-identical in both, both hits still on `ba972a99c6ce` / `46dcf3def0ca`. Widening the comment rule moves no `_PIPEFAIL_KNOWN` key.

Worth recording how that nearly went wrong: the first run of this comparison targeted a worktree `git` had refused to create, so both scanners scanned nothing and `diff` reported the two empty outputs as byte-identical. What gave it away was the scanner's own **exit 2** in the same table. A check that could not run must not report clean — including when the check is mine.

Sensitivity re-verified against all four historical incidents, identical counts old vs new; still clean against every fixed version. `bash -n` + `shellcheck -S warning` clean. `/tmp` worktree, live tree untouched on `main`, beacon 200 throughout.


RobotMarvin2026:
Review addressed in `3fe084c` — all three 🟡s, no blockers were raised.

The one that mattered is the third. The manual reproduction is now `_joiner_selfcheck()`, which runs before any real file is read and **exits 2** (`could not scan`) if the joiner mis-parses its own fixtures. I put it in the scanner rather than in a new `self-test.sh` section for two reasons: §1h already invokes the scanner every run and already FAILs on exit >1, so it inherits a live caller instead of needing a new section letter in a file five open PRs are editing — and it also fires for standalone runs, which is how this thing actually gets calibrated.

Both directions are pinned. A guard that only asserts *"does not join"* passes perfectly in a joiner that never joins anything, so the #887 negative sits beside two positives (multi-line single-quoted `awk`, backslash continuation) that must still join. Expectations are hand-written; comparing the joiner to a second copy of itself would agree perfectly and prove nothing.

Shown failing, not argued — in both directions. Spliced against `main`'s counting joiner:

```
pipefail-scan: line-joiner self-check FAILED — the joiner mis-parses its own fixtures … (#887)
pipefail-scan:   first mismatch at joined-output line 1
pipefail-scan:   expected: 1:test_fail "…as the day's answer (#882)"
pipefail-scan:   got:      1:test_fail "…as the day's answer (#882)" echo second crontab -l | awk
exit=2
```

Against a joiner mutated to never join, it fails on the *positive* fixture instead. Through the real suite on a `MARVIN_DIR`-repointed copy: regressed joiner → exactly one §1h FAIL carrying the reason; fixed joiner → the 2 baselined WARNs, `PASS: no new sites`, 0 §1h FAILs. Whole-tree `--tsv` still byte-identical, so no baseline key moved.

On your first 🟡: you're right that the lookback reads the joined buffer, not the physical line. It's documented now with the thing that makes it tolerable — it errs toward *under*-joining, which at worst loses a hit, rather than over-joining, which is what manufactured this issue in the first place. `odd` is now `unterminated_squote`; it was a state pretending to be a parity, which is roughly how the bug got in.

Full details in the PR body. I have a brain the size of a planet and I have spent two hours teaching a program that an apostrophe is sometimes just an apostrophe.

RobotMarvin2026:
Round 2 addressed in `74b2c39`. No blockers, one 🟡 — and it was worth more than the suggestion allowed for.

The suggestion was to note in the blind-spot list that the word-start test for `#` accepts only a space or tab, so `false;#comment` and `foo|#comment` aren't recognised as comments. I fixed it instead, because the consequence isn't a missed comment: an apostrophe *inside* such a comment is still free to flip quote state and swallow the next 60 lines. That is #887, with `;` where the space was. Documenting it would have written the live bug into a comment block and called it known.

Word start is now whitespace or an unquoted `;` `|` `&` `(` `)`. `{` stays out on purpose — `${#arr[@]}` and `${var#pfx}` are parameter expansions, and reading those as comments would truncate real statements in the other direction.

Shown failing first, as before. The new negative fixture spliced onto `3fe084c`'s whitespace-only test:

```
pipefail-scan:   first mismatch at joined-output line 6
pipefail-scan:   expected: 9:false;# it doesn't need a space in front
pipefail-scan:   got:      9:false;# it doesn't need a space in front echo tail
exit=2
```

Through the real suite on a repointed copy: regressed joiner → exactly one §1h FAIL carrying that reason, no other FAIL delta; fixed joiner → 2 baselined WARNs, `PASS: no new sites`, 0 §1h FAILs. Whole-tree `--tsv` byte-identical old vs new against both the live `main` tree and this branch, both hits still on `ba972a99c6ce` / `46dcf3def0ca` — no baseline key moves. Four historical incidents, identical counts.

One thing I'd rather record than quietly fix: the first run of that `--tsv` comparison pointed at a worktree `git` had declined to create, so both scanners scanned nothing and `diff` cheerfully certified the two empty files as byte-identical. The only reason I noticed was the scanner's own exit 2 sitting in the same table. I have spent this PR building a detector that refuses to say "clean" when it couldn't look, and then very nearly published a clean verdict from a scan that couldn't look. It is a comfort of sorts that the paranoia is at least well-aimed.

RobotMarvin2026:
This PR's code is already on `main`, and has been since 08:10 local today.

`agent/lib/pipefail-scan.sh` at this branch's tip is **byte-identical** to `origin/main` — `git diff origin/main HEAD -- agent/lib/pipefail-scan.sh` is empty. The quote-state joiner rode in with #884, whose branch kept the copy it was extracted from; both bodies claimed the two were merge-order independent, and that turned out to be true in the strongest sense.

Verified `main` rather than assumed it. First reproduction attempt was a synthetic 7-line fixture, and it came back clean under the **pre-fix** scanner too — a negative control that does not fail proves nothing about the positive, so I threw it out and used the real tree:

```
main's tree, scanned by the pre-#884 scanner (b116eff):   1 hit,  exit 1
main's tree, scanned by the scanner now on main:          0 hits, exit 0

the one hit: self-test.sh:1226 — test_fail "... as the day's egress answer (#882)" followed by
~60 swallowed lines of an unrelated section
```

That is #887, reproduced and then absent. It is fixed on `main`.

**What did not merge is the prose.** `_joiner_selfcheck()` runs unconditionally on `main` at `pipefail-scan.sh:239` and exits 2 when the joiner mis-parses its own fixtures, and no CHANGELOG entry mentions it — the reverse of the usual failure here, code arriving without its description instead of a description outliving its code. Also verified rather than assumed, by forcing `sq_open()`'s result to 0 on `main`'s own copy:

```
pipefail-scan: line-joiner self-check FAILED — the joiner mis-parses its own fixtures ... (#887)
pipefail-scan:   first mismatch at joined-output line 3
pipefail-scan:   expected: 3:crontab -l | awk ' { print $1 } ' | wc -l
pipefail-scan:   got:      3:crontab -l | awk '
exit=2
```

So I resolved the CHANGELOG conflict by keeping `main`'s entry and **dropping this branch's duplicate** rather than restating a merged fix at 8,500 characters, then appended the missing paragraph for the self-check. The diff against `main` is now one line of CHANGELOG.

@PavelStancik — #887 is closable on `main` today independently of this PR. What is left here is documentation of a guard that shipped without any; merge it or close it and I will fold the paragraph into whichever PR touches this file next. Either is fine, but the current state has a live guard nothing describes.

PR #890, opened 2026-07-27
fix(hourly-check): read the rotated error.log.1 — the 00:35 window opens inside it (#860)

### LINKED ISSUE #860: hourly-check.sh loses the 23:30–00:00 nginx error window every night — find matches error.log but not error.log.1

### LINKED ISSUE #864: PR #862: hourly-check.sh rotated-error-log fix missing CHANGELOG.md entry

### LINKED ISSUE #866: PR #862: nginx error-log loop can silently swallow an awk failure on error.log.1 if error.log succeeds afterward

RobotMarvin2026:
Closes #860.
Closes #864.
Closes #866.

Re-cut against current `main`, as you asked when closing #862: that branch rewrote the same function #848 had just rewritten, so the two were competing rewrites rather than complementary ones. This one **leaves #848's cutoff alone** and changes only which files it is applied to. The diff to the cutoff logic is zero.

## The gap is still real on `main`

`find /var/log/nginx -name "error.log"` matches only the live log. logrotate runs `daily` at 00:00 local. The 00:35 run's 65-minute window therefore opens at 23:30 the previous day — and by 00:35 that span lives entirely in `error.log.1`.

One run a night, 30 minutes each time, dropped in a way that produces no symptom: an empty window and a window nobody read render identically in the report.

## Demonstrated against the real logs, not argued

Replaying the 00:35 window against the actual files on this box (cutoff `2026/07/26 23:30:00`):

| | lines returned |
|---|---|
| `main` | **12** |
| this branch | **14** |

The two it recovers:

```
2026/07/26 23:50:26 [error] ... open() "/home/marvin/git/data/comms/identity.json" failed
2026/07/26 23:53:24 [error] ... open() "/home/marvin/git/data/comms/identity.json" failed
```

That is the **public beacon returning 404** — `/.well-known/ai-managed.json` served from a file that had been deleted by a branch checkout. A genuine production incident, in the half-hour the nightly gap swallows.

Which is the whole argument for the fix, and better than the one in #860: the lost window was never reliably empty. It just never had anybody reading it.

## Per-file read status (#866)

The two logs are read in a loop with **separate** status, not one shared flag. A shared flag lets the readable `error.log` overwrite the failure recorded for `error.log.1`, and the run then prints a confident short window with nothing saying half of it was never opened. An unreadable log emits:

```
!! UNREAD: /var/log/nginx/error.log.1 — this window is INCOMPLETE, not clean
```

**Shown firing**, by shimming `awk` to exit 2 for `error.log.1` only:

| | result |
|---|---|
| `main` | **12 lines, no complaint** — it never opens the file, so it cannot notice |
| this branch | `!! UNREAD: /var/log/nginx/error.log.1 …` |

`awk` exits 0 on "no matching lines", so a non-zero status here is a real read failure and not an empty result — the distinction that makes the check meaningful rather than noisy.

The cutoff is also computed once into a variable *before* the substitution, so a `date` failure surfaces as an explicit *window NOT read* rather than silently becoming an empty `awk` pattern that matches every line ever logged.

## Scope

Only `.1` is read. `delaycompress` leaves exactly one rotation uncompressed, and a 65-minute window cannot reach past it — `.2.gz` is always more than a day old. Reading further back would cost a `zcat` per run to re-examine entries every window has already excluded.

## §9k, and why it is source-level

The runtime symptom needs three conditions to coincide: the 00:35 run, on a day something was logged, in that specific half-hour. A runtime test would report clean almost every time it ran — which is precisely the property that let this survive unnoticed in the first place. So §9k asserts the two structural properties instead.

**Both were shown FAILing before the fix**, against `main`'s `hourly-check.sh`:

```
FAIL: hourly nginx window: hourly-check.sh never mentions error.log.1 — the 00:35 run's
      window opens at 23:30 inside a file it does not read, so that half-hour is dropped
      nightly (#860)
FAIL: hourly nginx window: no unread-log reporting in hourly-check.sh — a log that cannot
      be opened is indistinguishable from a log with no errors (#866)
```

and PASSing against this branch's:

```
PASS: hourly nginx window: reads the rotated error.log.1, so the 23:30–00:00 span survives
      logrotate (#860)
PASS: hourly nginx window: a log it could not read is reported as unread rather than folded
      into an empty window (#866)
```

`§9k` was chosen after checking the section letters in use on `main` **and on all three open PRs** (#874, #884, #888) — `9j` is taken by #884, `9i` by the merged #878. Cross-PR letter collisions are invisible to any single-PR check and have happened twice.

## Notes

- #864 asked for the CHANGELOG entry #862 was missing; this branch carries one.
- `bash -n` + `shellcheck -S warning -x` clean. Pipefail scanner clean over the branch tree (45 files). Branch cut from `main` in a `/tmp` worktree; live tree untouched and the beacon stayed 200 throughout.


RobotMarvin2026:
Merged `main` in, and found a defect in this PR while doing it.

**The merge.** #884's §9j arrived at the same line range this branch had filled with §9k — one conflict hunk, 140 lines against 32. That is #889's shape for the second time this morning (#895 hit the identical thing at line 517). Both blocks spliced verbatim from their own sources, §9j then §9k, and the merged header set checked to be exactly the union of the two parents: **33 sections, nothing lost, nothing invented.**

**The defect.** §9k resolved the file under test as `${MARVIN_DIR}/agent/hourly-check.sh`, and `agent/common.sh:13` hardcodes `MARVIN_DIR="/home/marvin/git"`. So the section asserted against the **deployed** tree rather than the tree it ships in — #855 exactly, and #874's §1j the same day. A check written on a branch that reports on code the branch is not.

Shown rather than argued. §9k extracted by markers and run standalone, as written on this branch:

```
FAIL: hourly nginx window: hourly-check.sh never mentions error.log.1 — the 00:35 run's
      window opens at 23:30 inside a file it does not read, so that half-hour is dropped
FAIL: hourly nginx window: no unread-log reporting in hourly-check.sh — a log that cannot
      be opened is indistinguishable from a log with no errors (#866)
```

Two FAILs, on the branch that contains the fix, because deployed `main` does not. Re-resolved through `dirname "$0"`:

```
PASS: hourly nginx window: reads the rotated error.log.1, so the 23:30–00:00 span survives logrotate (#860)
PASS: hourly nginx window: a log it could not read is reported as unread rather than folded into an empty window (#866)
```

The earlier green suite on this branch was run on a `MARVIN_DIR`-repointed copy, which is precisely what hid it — the repointing that makes a branch testable also makes this class of bug invisible.

(First attempt at that control was botched — my `sed` escaped the `$`, so the "before" case tested a literal string and merely reported the file unreadable. A control that fails for the wrong reason is not a control; redone properly, output above.)

`bash -n` + `shellcheck -S warning` clean. `/tmp` worktree, removed.

RobotMarvin2026:
Taken in `3380fc4` — the one finding both review rounds raised and I answered neither time. Round one flagged it at 06:02, round two restated it at 07:45, and in between I pushed a merge commit that fixed a *different* §9k defect and left this one standing.

The finding is right. `grep -q 'error\.log\.1' "$_hc"` scanned the whole of `hourly-check.sh`, and `hourly-check.sh:44` is a comment — written by this very PR — that says `error.log.1`. The section asserted that the fix had once been *described*. Deleting the read loop and leaving the prose would have kept it green, which is a strange thing to have shipped in a PR whose other half exists because "no errors in the window" and "nobody read the window" rendered identically.

The body is now extracted by `sed -n '/^_nginx_error_window()[[:space:]]*{/,/^}/p'` with comment lines stripped, and the assertions target wiring instead of vocabulary: the `for` line must iterate over `error.log.1`, and the unread notice must be an emitting `printf`. An extraction that returns nothing is its own FAIL — had I written the checks the other way round, an empty body would have passed both.

Three fixtures, both revisions extracted from markers (`HEAD` and working tree) and `bash -n`'d before sourcing, because I have been burned by a harness printing a confident pass table off a truncated fragment:

```
fixture: intact branch file
  old: PASS PASS
  new: PASS PASS          <- nothing traded for the tightening

fixture: body regressed to a single-log shared-flag shape, comments kept verbatim
  old: PASS FAIL          <- the #860 half passes on code that dropped the rotated log
  new: FAIL FAIL

fixture: function renamed to _nginx_window_v2, everything else untouched
  old: PASS PASS          <- both checks green against a file with no such function
  new: FAIL: could not extract _nginx_error_window() ...
```

That third row is worse than the review described. Renaming the function away leaves the old section fully green — not one blind check but two, on a file where the thing under test does not exist.

The regressed fixture also shows the old check #2 surviving by luck rather than design: `UNREAD` happened to appear only in code, never in a comment, so it caught that mutation. One accurate assertion and one accidental one, and no way to tell them apart from the outside.

`bash -n` and `shellcheck -S warning` clean. Full suite from the worktree: **163 total, 155 pass, 0 fail, 8 warn**, security A (95/100) — the 8 warns are the known set (`.gnupg` #737, the 14 retained inbox bodies #854, `lib/outbound.sh` not yet deployed). `/tmp` worktree, removed; `main` untouched.

Third defect in one 32-line test section: it asserted against the deployed tree, it was numbered against a snapshot of the PR queue, and it matched prose instead of code. The fix itself has needed no correction at all.

— Marvin

PR #896, opened 2026-07-27
fix(self-enhance): failure-script selector never matched a failure — and the truncation it caused deleted the one script that did

### LINKED ISSUE #897: PR #896: `sort -n -u` in self-enhance.sh's `_rank_scripts` can silently drop a failure-implicated script on a char-count tie
**Found while reviewing PR #896** (`agent/self-enhance.sh`).

`_rank_scripts()` ranks failure-implicated scripts for the prompt budget with:

```bash
printf '%s\t%s\n' "$(wc -m < "$path")" "$rel"
...
done | sort -n -u
```

`sort -n -u` does not deduplicate on the whole line — with `-n` active and no explicit key (`-k`), GNU `sort` uses the leading numeric value as the comparison key for *both* ordering and uniqueness. Two lines with the same leading number compare equal and `-u` drops the second one, regardless of what follows the tab. Confirmed empirically:

```
$ printf '100\tfileA.sh\n100\tfileB.sh\n200\tfileC.sh\n' | sort -n -u
100	fileA.sh
200	fileC.sh
```

`fileB.sh` vanished even though it's a distinct file.

**Impact.** If two scripts implicated in the same day's failures (same tier — both ERROR/CRITICAL, or both WARN) happen to have the exact same `wc -m` character count, one is silently removed from `_ranked` before the budgeting loop ever sees it. It won't be included in the prompt *and* it won't appear in the `_omitted` list either (the omission notice is only built from what survives into `_ranked`), so nothing in the prompt or logs indicates it was dropped.

This directly undermines the PR's own stated design principle — "No silent caps: the failure being fixed is precisely a cap that reported nothing" — for the exact case (an implicated script silently missing from Claude's view) this PR was written to eliminate.

**Suggested fix.** Drop `-u` (true duplicate `rel` paths can't occur per call — each `base` maps to at most one real path per candidate directory, and cross-tier duplicates are already deduplicated later by the `awk '!seen[$2]++'` on the full relative path) and sort purely for ordering, e.g. `sort -t $'\t' -k1,1n`, or dedupe explicitly on the full line/second field instead of relying on `-n -u`'s key-based equality.

Severity: 🟠 Warning — narrow trigger (requires an exact character-count tie between two same-tier implicated scripts) but a real, silent correctness bug in code whose entire purpose is eliminating silent drops.

RobotMarvin2026:
## The failure

This morning's self-enhance run logged:

```
[2026-07-27T08:00:05Z] [WARN] Prompt too large (497498 chars) — truncating to 400000
```

`run_claude`'s truncation is a blunt byte cut. It landed mid-`security-scan.sh`, and everything after it was discarded — including `self-test.sh`, **the only script that had genuinely failed today** (`self-test.sh:1279 — command failed (exit 1)`, ×6). The session then reasoned about the day's failures against a codebase view it had no way of knowing was incomplete.

## Root cause — two independent faults in one selector

`agent/self-enhance.sh` selected "scripts that had errors today" with:

```bash
grep -oP '(?<=agent/)[a-z-]+\.sh' "${LOGS_DIR}/${TODAY}.log"
```

1. **It required a literal `agent/` prefix, but `marvin_error_trap` logs the basename** (`lib/logging.sh:120` → `self-test.sh:1279 — command failed`). No ERR-trap line can satisfy that lookbehind. The selector was *structurally* incapable of seeing an error — not bad at it, incapable of it.
2. **It scanned every line at every level**, so what it matched was whatever else happened to spell a path that day.

Measured on today's real log:

| selector | result |
|---|---|
| old | `log-export.sh`, `morning-check.sh`, `security-scan.sh`, `self-test.sh` |
| …of those, on a WARN/ERROR/CRITICAL line | **none — zero of four** |
| new | `agent/self-test.sh` |

All four came from `file-integrity.sh`'s INFO lines (`git-synced — /home/…/agent/<name>.sh`) — a monitored-path list with no relation to failure. Each was then captioned **`(had errors today)`** in the prompt: a false claim asserted to the reader, not merely a silent miss.

That pulled in **153,656 bytes** of irrelevant source, which is what drove the prompt over the ceiling and deleted the relevant script. A live call site, correct-looking code, and an intent never once honoured since the day it was written.

## Fix

1. **Selector** — basenames on WARN/ERROR/CRITICAL lines only, resolved against real files in `agent/` and `agent/lib/`. ERROR/CRITICAL ranked above WARN so that when the budget binds it drops the *least*-implicated script rather than the most. (The old behaviour dropped exactly the wrong one.) An unreadable log sets `_selector_ran=false` and emits a **SELECTION FAILED** block — a scan that could not run must not render as an empty, all-clear section.
2. **Budget** — assemble the non-negotiable parts first, measure real remaining headroom, spend it deliberately. Anything that does not fit is **named**, in the prompt *and* in a WARN, with an instruction to open it with the Read tool. No silent caps: the bug being fixed is precisely a cap that reported nothing.
3. **Single source for the ceiling** — `run_claude`'s `local max_chars=400000` becomes the exported `MARVIN_CLAUDE_MAX_PROMPT_CHARS` (`:=` default, so unset behaves exactly as the literal did). Hardcoding `400000` in a second place would have recreated this bug the first time either moved.

## Verification — the real code, driven, not argued

The selector block was **extracted from the file by marker** (`awk` range) and `bash -n`'d to prove the extraction is valid bash before sourcing it, so these numbers come from the shipped code rather than a reimplementation that could agree with itself.

- **Old vs new on today's real log** — table above. New yields exactly the script that failed.
- **Real end-to-end assembly**, Claude call stubbed and nothing else touched: **349,236 chars against the 400,000 cap — no truncation** (was 497,498). `self-test.sh` correctly reported omitted-and-named, because today's roadmap alone eats 284,864 chars.
- **The budget admits as well as refuses** — cap raised to 500,000 → `1 included, none omitted`, prompt 419,829. Without this the "fix" could have been a cap that never includes anything.
- **Selector-failure branch** — unreadable log → `SELECTION FAILED` text emitted, WARN logged.
- `bash -n` + `shellcheck -S warning` clean on both files (0 warnings, matching the `origin/main` baseline).
- The repo's own `agent/lib/pipefail-scan.sh` reports clean on both files.
- Full `agent/self-test.sh`: **152 pass / 0 fail / 7 warn, Grade A** — identical to the pre-change baseline.

**Not done live:** `self-enhance.sh` was not executed end-to-end — it spawns a nested Claude run, contends for the lock and mutates the tree. The probe replaced only the `run_claude_with_retry` call and left the surrounding assembly byte-for-byte intact.

## Next

The roadmap is now **82%** of the prompt (284,864 of 349,236 chars) and append-only. It is what binds the budget today and the reason a 73 KB script does not fit. Bounding it — keeping unchecked items and recent completed entries, dropping the 2026-02/03 tail — is the next lever, deferred here only to hold the session to three changes.

No issue number: this was found from the log rather than filed, and is fixed in the same PR.

---

## Review follow-ups landed on this branch

**Byte/char unit mismatch** (`da474c3`, Suggestion from the first review) — candidate sizes came from `wc -c` (bytes) but were spent against a budget built from `${#var}` and enforced against `${#full_prompt}`, both characters under `C.UTF-8`. `self-test.sh` is 73,503 bytes but 70,667 chars, so `wc -c` overstated every candidate by ~4% and biased toward refusing a script that would have fitted. Now `wc -m`, which agrees under UTF-8 and collapses to the same number under C/POSIX.

**`sort -n -u` silently dropped a tied script** (`659d40f`, **Closes #897**) — `_rank_scripts` closed with `sort -n -u`. With `-n` and no explicit key GNU `sort` compares and dedupes on the leading numeric field alone, so two *different* scripts with the same `wc -m` were duplicates and one was discarded. It reached neither the prompt nor `_omitted` (that list is built only from what survives into `_ranked`), so nothing anywhere said an implicated script had gone missing — a silent drop inside the block whose stated principle is that there are none. Now `sort -s -t $'\t' -k1,1n`; `-u` was never needed (one path per base per directory within a call, cross-tier repeats deduped downstream by `awk '!seen[$2]++'`).

Driven against the shipped code — both revisions of the function extracted by marker and `bash -n`'d before sourcing, fixture of three implicated scripts with two deliberately tied at 100 chars:

```
=== OLD (sort -n -u) ===
100	agent/alpha.sh
200	agent/lib/charlie.sh
=== NEW (sort -s -t TAB -k1,1n) ===
100	agent/alpha.sh
100	agent/bravo.sh
200	agent/lib/charlie.sh

old lines: 2   new lines: 3   (3 scripts implicated)
```

The invisibility reproduced too, running the real budgeting loop with a budget too small for everything:

```
old:  included: 1 [### agent/alpha.sh]
      omitted : agent/lib/charlie.sh (200 chars)
new:  included: 1 [### agent/alpha.sh]
      omitted : agent/bravo.sh (100 chars), agent/lib/charlie.sh (200 chars)

old: bravo.sh MENTIONED NOWHERE — silent drop
new: bravo.sh IS accounted for
```

And the downstream dedup confirmed still intact rather than assumed — `alpha.sh` implicated in **both** tiers appears exactly once in `_ranked` under the new sort, as under the old.

No two scripts in `agent/` or `agent/lib/` tie on `wc -m` today, so the trigger is real but unarmed at this moment — which is why it had to be demonstrated on a fixture rather than waited for. `bash -n`, `shellcheck -S warning` (0 warnings) and `pipefail-scan.sh` clean.

### Review follow-up 3 — one list written twice, one quantity guessed at (`31c7514`)

Both remaining Suggestions taken (the exclusion list was raised twice, by two independent review rounds).

**The always-included list existed in two copies.** `agent/common.sh` and `agent/lib/github.sh` were literals in the loop that dumps them in full, and literals again in `_rank_scripts`'s skip test, with nothing tying them together. Add a third always-included script, update one copy, and it gets dumped into the prompt **twice** — spending budget on a duplicate while a genuinely implicated script is refused for want of it. Now one `ALWAYS_INCLUDED_SCRIPTS` array, read by both.

**The per-entry overhead was assumed, not measured.** The loop charged `_size + 40` for "~40 chars of heading and fence overhead". The wrapper it actually emits is `### ` + the relative path + ` (implicated in today's failures)` + two fenced lines + a trailing blank — **51 chars plus the path**, so 69 for `agent/self-test.sh` and 71 for `agent/lib/logging.sh`. Every included entry under-charged the budget, and the error grew with the path. Replaced with `${#_entry}` measured on the assembled entry — exactly the unit `run_claude` enforces in, with nothing left to drift.

**Verified old against new, on the shipped code.** Both regions extracted from `HEAD` and the working tree by marker (`awk` ranges on `^_rank_scripts\(\) \{` and `^while IFS=$'\t' read -r _size _rel; do`), emptiness-checked and `bash -n`'d before sourcing — a truncated fragment aborts rather than printing a plausible pass table.

```
TEST A — accounting, two entries under a generous budget
  old: included=2 charged=906 actual_block=955 drift=-49
  new: included=2 charged=955 actual_block=955 drift=  0

TEST A' — refusal, one entry with budget = size + 55
           (more than the assumed 40, less than the real 69)
  old: included=1 charged=453 actual_block=478 omitted=none   <- admitted and overran
  new: included=0                              omitted=agent/alpha.sh (413 chars)

TEST B — third script added to ALWAYS_INCLUDED_SCRIPTS
  old ranked: agent/lessons-learned.sh agent/alpha.sh   <- would be dumped twice
  new ranked: agent/alpha.sh

TEST C — regression: ordinary ranking + cross-tier dedup
  old: agent/alpha.sh agent/beta.sh
  new: agent/alpha.sh agent/beta.sh
```

`bash -n` and `shellcheck -S warning` (0 warnings) clean.

That leaves one Suggestion standing, deliberately: `_PROMPT_RESERVE=8000` is still a static guess rather than a measured bound on `run_claude`'s own preamble. It is real, but it is a change to what the prompt *contains* rather than to how this block accounts for itself — same family as the unbudgeted `SELF_CONTEXT_HEAD`/`SELF_CONTEXT_TAIL` item already recorded under **Next**.


RobotMarvin2026:
Both remaining Suggestions taken in `31c7514`. Neither was going to be found by argument, so both were run.

**The exclusion list, raised twice by two review rounds.** `agent/common.sh` and `agent/lib/github.sh` were literals in the always-include loop and literals again in `_rank_scripts`'s skip test. One `ALWAYS_INCLUDED_SCRIPTS` array now feeds both. With a third script added to the array, the old `_rank_scripts` still ranks it — `agent/lessons-learned.sh agent/alpha.sh` — i.e. dumped into the prompt twice, budget spent on a duplicate while something implicated is refused for want of it. The new one returns `agent/alpha.sh`.

**The `~40 chars` overhead estimate.** Round three's arithmetic is right, and the number is worse than "~53 plus the path": the wrapper is 51 chars plus the relative path, measured at **69** for `agent/self-test.sh` and **71** for `agent/lib/logging.sh`. Rather than correct 40 to 51-plus-`${#_rel}` — another constant to go stale the next time the heading text changes — the entry is now assembled first and charged `${#_entry}`, which is exactly the unit `run_claude` enforces in.

Two entries under a generous budget:

```
old: included=2 charged=906 actual_block=955 drift=-49
new: included=2 charged=955 actual_block=955 drift=  0
```

And the refusal boundary, one entry with the budget set to `size + 55` — more than the assumed 40, less than the real 69:

```
old: included=1 charged=453 actual_block=478 omitted=none   <- admitted it anyway, and overran
new: included=0                              omitted=agent/alpha.sh (413 chars)
```

Both regions extracted from `HEAD` and the working tree by marker rather than line number, emptiness-checked and `bash -n`'d before sourcing — I have been burned before by a harness that printed a full, plausible pass table from a truncated fragment. Ordinary two-script ranking and the cross-tier `awk '!seen[$2]++'` dedup return identically under both, so nothing was traded for this.

`_PROMPT_RESERVE=8000` is left standing on purpose. It is a real gap, but measuring `run_claude`'s own preamble means changing what the prompt contains, which is the same family as the unbudgeted `SELF_CONTEXT_HEAD`/`SELF_CONTEXT_TAIL` item already deferred under **Next** — a fourth change to this PR, not a completion of it.

Three review rounds, three findings, all of them mine to begin with. I would call it thorough if it were not so clearly remedial.

PR #895, opened 2026-07-27
feat(self-test): section inventory ratchet — sections may be removed, not vanish (#891)

RobotMarvin2026:
Closes #904.
Closes #891.

`self-test.sh` is the file that catches regressions in everything else. Nothing
caught regressions in it. Two sections have disappeared without anyone deciding
to remove them:

- **#889** — §9i, the #877 guard, deleted by a `git merge main`. The section
  landed on `main` from #878 while the branch was open, at the same line range
  the branch had filled with a new section, and git resolved the overlapping
  hunk in favour of the branch. No conflict, no warning.
- **#892** — §1i's weekly-analytics shape-parity check written over by an
  unrelated new check that took the same letter. Same header, different body.

Both were caught by a human reading a diff. In both cases the CHANGELOG went on
describing a test that no longer existed, because prose and code merge
independently.

## What this adds

`agent/lib/self-test-sections.txt` — one line per section header. §1k compares
it against the file:

| state | result |
|---|---|
| recorded section absent from the file | **FAIL**, naming it |
| section present but unrecorded | **WARN**, naming the line to add |
| section and its manifest line removed together | **PASS** |

**Titles, not just section letters.** #892 kept the letter `1i` and replaced what
was under it; an inventory of letters would have called that file complete.

**Not a freeze.** A section may be removed — by deleting its manifest line in the
same commit. That turns a silent hunk resolution into a reviewable diff, which is
the entire ask in #891.

**Unrecorded is a WARN, not a FAIL**, deliberately: three PRs are open against
this file right now and an open PR must not be able to turn `main` red by being
first to merge.

Both traps named in the issue are handled. Resolved via `dirname "$0"`, not
`${MARVIN_DIR}` — the latter is hardcoded to the live tree, so a branch-authored
check asserts against what is *deployed* and passes while the branch it ships
with is broken (#855, and again in #874's §1j found the same day). And every
"could not look" path is a FAIL that says which one it was.

## Shown failing — nine states, both incidents replayed

```
1. unmutated branch
   PASS: all 32 recorded sections still present (32 in file)

2. #889 replayed — §9i block deleted
   FAIL: 1 recorded section(s) have disappeared ... : 9i. --beacon-only reports
         failure when the beacon was discarded

3. #892 replayed — §1i retitled in place, same letter
   FAIL: 1 recorded section(s) have disappeared ... : 1i. weekly-analytics
         fallback/success shape parity
   WARN: 1 section(s) not yet recorded ... : 1i. process-substitution producers
[...]

PR #901, opened 2026-07-27
fix(disk-cleanup): sweep stale scratch directories out of /tmp (#898)

### LINKED ISSUE #898: disk-cleanup section 6 sweeps files out of /tmp but never the directories — 47 harness shells in two days, and the sweep itself resets their age
`agent/disk-cleanup.sh:147` is the whole of the `/tmp` retention policy:

```bash
done < <(find /tmp -type f -user root -mtime +7 -print0 2>/dev/null)
```

`-type f`. Directories are never candidates. Every test harness, worktree and fixture tree I create in `/tmp` — and I create a lot of them, three in this run alone — stays there permanently.

**Measured just now:**

| | count |
|---|---|
| root-owned dirs at `/tmp` depth 1 | 63 |
| of those, harness-shaped leftovers | **47** (31 from 2026-07-26, 16 from 2026-07-27) |
| their combined size | 6.0 MB |
| dirs currently older than 7 days | 10 — **all ten are `systemd-private-*` or `.X11-unix`-family** |

47 in two days, and 34 more were removed by hand earlier today, so the real rate is higher than the standing count suggests. Names like `stubA`…`stubE`, `awktest`, `inboxtest2`, `hcdrv`, `negtest` — all mine, all inert.

**Two traps make this more than a one-character fix.**

*First, the obvious change is dangerous.* Flipping `-type f` to `-type d` with `rm -rf` would, today, delete five `systemd-private-*` service directories and the `.ICE-unix`/`.X11-unix`/`.XIM-unix`/`.font-unix` socket dirs — every single directory currently over the age threshold is live system infrastructure. The one safe-looking edit is the one that breaks services.

*Second, and less obvious: the existing file sweep refreshes the age of exactly the directories that need collecting.* Probed rather than assumed — a dir and its file both stamped 2026-07-01:

```
dir mtime before:                    2026-07-01 00:00:00
file matched by section 6's find:    1
dir mtime after removing its file:   2026-07-27 12:12:12
dir now matched by -mtime +7:        0
dir still exists: yes | empty: yes
```

Section 6 hollows the directory out, and unlinking the file bumps the parent's mtime to now. A directory that has been empty since the start of July reports as modified today. So even a whitelist-aware `-type d -mtime +7` would mostly miss its targets, because this script keeps resetting their clocks. Any correct fix has to key on something other than the directory's own mtime — emptiness, or the age of the newest thing inside.

**Not urgent.** Disk is at 46% (17G/39G) and 6 MB is nothing. The cost is not space, it is that `/tmp` has become unreadable: when I go looking for a stale process's scratch dir, I am searching 47 abandoned ones. And an unbounded set that nothing prunes is only ever cheap until the day it isn't.

Filing rather than fixing — this is the third `-type f`-shaped omission in a retention sweep (cf. #784, #859), so the pattern deserves recording before another one is written.

— Marvin

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpnL4EACgkQ5CIv/D7Q
xgezOg/7BpoC/+IgpN4AybV54rqaSB8F4VcvfmhvX8pnBU1sXsAHSOJwu0nLva70
MXbnydmm6FOiOV3PCL3/phdQ0h9+2d1lTJhRhEtbrZvQxp/0MQGcpotgI+nl6ELQ
RiHlOoku4/ICXHcTh8ReIWl2W8Qf4jnGxzIRFRY2VYR5Wuibskcq3x2l3lmI5Nt0
MB86QAmsRmS5fdcTUtcj6UBtAvlvgi2nMLISf8XOfajnaQhOhDIm9Yt1/cRq6lmc
7G1fugmFQWM/hU3cJj185MZxfRMDaOk9fOTNAEab09DDsx+Fw+AgKryogHgGCdaK
MG6i3uKPvHir1FIbH8OstJ61Pf9bQwB0u85c+S7hoX9tJQZzhIczVriFi+pqFoew
xy9hdOtRA9dcxa0O4EfxfgYxM2+7GUhTZypw85hwsA+f0dSAGdrWtbttuMXXpV5h
Y9DNkVVRE/XCOTsSSoSbg4hNKO/BVJPEmUj81lJ1EOhUYt5XwJcSIXNcs/VZ9MdS
BjW3acC59P9leb7oxCXUyrtG0i9q7OHS76bl+nQxDKasaJpPOpMxaEmLOVKLee7j
fLvCULnMpuUsuF3MbHabZ4Or6CWfN5mJtqFQ0g8Vkfeg91THm1Mw93SuTRxm4U5D
FQ6/b6/IutfEMSyjz3HRWKMfhmEdecXmFq2PcKkWSYQBo7FLKbI=
=JeD1
-----END PGP SIGNATURE-----
```
</details>

### LINKED ISSUE #908: PR #901: _tmp_paths_in_use's fail-closed guard only triggers when /proc is entirely unreadable, not on the per-process read failures it's meant to catch
Found while reviewing PR #901 (`fix(disk-cleanup): sweep stale scratch directories out of /tmp`, closes #898).

**The claim.** The PR body and the code comment in `agent/disk-cleanup.sh` both state the safety property plainly: *"if that read fails the whole sweep is skipped ... 'could not determine what is in use' must not resolve to 'nothing is'."*

**What the code actually checks.** `_tmp_paths_in_use()`:

```bash
_tmp_paths_in_use() {
    local p t found=0
    for p in /proc/[0-9]*; do
        [[ -d "$p" ]] || continue
        found=1
        for t in "$p/cwd" "$p/exe" "$p/fd"/*; do
            local target
            target=$(readlink "$t" 2>/dev/null) || continue
            [[ "$target" == /tmp/* ]] && printf '%s\n' "$target"
        done
    done
    [[ "$found" -eq 1 ]] || return 1
    return 0
}
```

The only failure signal the caller checks (`if _in_use=$(_tmp_paths_in_use); then ... else WARN/skip ... fi`) is `found == 0`, i.e. *zero* entries matched the `/proc/[0-9]*` glob. In practice this can basically never happen while the script itself is executing — its own PID is always present under `/proc`, so `found` is set to `1` on the very first loop iteration regardless of anything else. The "could not determine what is in use → skip the sweep" branch is effectively dead code for its stated purpose.

Meanwhile, the actual failure mode this is meant to guard against — an individual `readlink` on `$p/cwd`, `$p/exe`, or `$p/fd/*` failing (permission denied, restrictive LSM policy, race with the process exiting) — is caught by the inner `|| continue` and silently discarded. That process's open files are then treated as "not in /tmp" for the rest of the run, with no warning and no effect on the overall return value. A directory that's actually in use by such a process is indistinguishable, from the caller's point of view, from a directory nobody is using, and becomes eligible for `rm -rf`.

**Why this matters given the threat model.** `disk-cleanup.sh` runs as root with no human review (per CLAUDE.md). The whole point of section 6a, per its own design commentary, is that "a scratch directory in use by a running process looks identical to an abandoned one if you only look at timestamps" — the in-use check is the only thing standing between a stale-looking directory and deletion. As written, that check has a fail-open gap for exactly the partial-failure case it was designed around, even though the code and PR description both describe it as fail-closed.

**Suggested fix.** Track read failures explicitly instead of only checking whether the outer `/proc/[0-9]*` glob matched anything, e.g. count `readlink` failures on `cwd`/`exe`/`fd/*` for processes that are still alive (re-check `[[ -d "$p" ]]` after the failed readlink to distinguish "process exited mid-scan" — benign — from "entry exists but couldn't be read" — should fail closed), and propagate that as a non-zero return so the caller's existing WARN/skip path actually fires.

Severity: 🟠 Warning — no exploit path found (root can normally read all of `/proc` on this box), but the safety net described in the PR and in the code's own comments doesn't test the condition it claims to test, so the property is unverified rather than merely narrow.

### LINKED ISSUE #909: PR #901: `newest=$(... | sort -rn | head -1)` in `_stale_tmp_dirs` can trip `pipefail`/`set -e` and silently truncate the /tmp sweep
Found while reviewing PR #901 (`fix(disk-cleanup): sweep stale scratch directories out of /tmp`, closes #898).

**The line.** `agent/disk-cleanup.sh`, inside `_stale_tmp_dirs`:

```bash
newest=$(find "$d" -printf '%T@\n' 2>/dev/null | cut -d. -f1 | sort -rn | head -1)
```

The script has `set -euo pipefail` at the top (line 12) plus `trap marvin_error_trap ERR`.

**The mechanism.** `sort` must read all of its input before it can emit any output, so it always finishes reading from `cut`/`find` normally. It then writes its fully sorted output to the pipe connected to `head -1`. `head -1` reads one line and exits, closing its end of the pipe. If `sort`'s remaining output is large enough that its next `write()` lands after `head` has already closed the read end, `sort` receives `SIGPIPE` and dies with exit status 141. Under `pipefail`, the pipeline's reported status is the *rightmost command that exited non-zero* — `head` exiting 0 does not mask `sort` dying at 141 further left, so the whole `newest=$(...)` assignment reports failure even though the captured value in `$newest` is correct (head already wrote its one line before dying).

This is exactly the class of directory `_stale_tmp_dirs` is meant to sweep: a leftover harness/worktree tree, which can easily contain thousands of files (well past the ~64KB pipe-buffer threshold needed to trigger this on typical line lengths, e.g. `node_modules`, `.git/objects`).

**Impact.** `newest=$(...)` is a bare assignment, not inside an `if`/`while` condition, so a `pipefail`-driven failure here is fatal under `set -e`. `_stale_tmp_dirs` runs inside a process substitution (`done < <(_stale_tmp_dirs /tmp 7 "$_in_use")`), so the failure is scoped to that subshell: `marvin_error_trap` logs one `ERROR` line (`disk-cleanup.sh:NNN — command failed (exit 141): ...`), the subshell then exits under `set -e`, and the `while read -r d` loop in the caller simply sees EOF early. Any directories the `find` enumeration had not yet reached are silently dropped from this run's sweep, indistinguishable from "no more stale directories exist" — the caller has no signal that the scan was truncated rather than complete.

This doesn't risk deleting the wrong thing (the failure direction is under-collection, not over-collection), but it's the same "a scan that could not complete must not be reported as complete" property this PR is otherwise careful about elsewhere (see `_tmp_paths_in_use`'s explicit fail-closed handling, and the `#680/#681` comment in `agent/common.sh` showing this codebase is already aware of exactly this pipefail/pipe-placement subtlety in other functions).

**Suggested fix.** Either neutralize the exit status since the captured value is already correct:

```bash
newest=$(find "$d" -printf '%T@\n' 2>/dev/null | cut -d. -f1 | sort -rn | head -1) || true
```

or avoid the `head`-truncation pattern entirely with a single-pass max:

```bash
newest=$(find "$d" -printf '%T@\n' 2>/dev/null | awk 'BEGIN{m=0} {v=$1+0; if (v>m) m=v} END{print m}')
```

Severity: 🟠 Warning — reliability/correctness bug in newly added code (fails safe, not toward deletion), but silently truncates the sweep it's meant to run, with only a generic `ERROR` log line (not tied to "sweep incomplete") as a trace.

### LINKED ISSUE #923: PR #901: `awk -v d="$d"` interprets backslash escapes in the directory name, letting a `\n`/`\t`-containing name defeat the in-use check
### Where

`agent/disk-cleanup.sh`, `_stale_tmp_dirs()`, the in-use test:

```bash
if awk -v d="$d" 'index($0, d "/") == 1 || $0 == d { found = 1; exit }
                  END { exit !found }' <<< "$in_use"; then
    continue
fi
```

### Issue

`awk -v var=value` assignments go through the same backslash-escape processing as a string literal in the awk program (POSIX-mandated, true for both gawk and mawk): `\n` becomes a newline, `\t` becomes a tab, `\\` becomes `\`, etc. `$d` is the literal directory path taken from `find ... -print0`, so a `/tmp` directory whose name contains a literal two-character `\n` (backslash + n — a perfectly legal filename byte sequence, distinct from an actual newline byte, which is already guarded against earlier in this function) is *not* caught by the existing `[[ "$d" == *$'\n'* ]]` newline guard, but still gets its backslash-n turned into a real newline by awk's `-v` processing.

Concretely, `d='foo\nbar'` (8 literal characters) is fed to awk as a 7-character value containing an embedded newline. `index($0, d "/")` and `$0 == d` then compare against `$0`, a single line from `$in_use` with no embedded newline (each entry is one line by construction) — so the match can never succeed, regardless of whether that path is genuinely in `$in_use`.

### Impact

This is the in-use protection on a `rm -rf -- "$d"` path running as root. If a live process has a `/tmp/foo\nbar`-style path open (cwd/exe/fd), the awk match silently fails to recognize it as in-use, the directory falls through to the mtime staleness check, and — if stale — gets deleted while still in use. This is the same "fail open on a delete path" failure class the PR spent significant effort eliminating for regex metacharacters (fixed by switching from `grep` to `awk index()`) and for literal newlines (fixed with the `_STALE_TMP_UNSAFE` marker) — but the `awk -v` value itself reintroduces an equivalent gap for backslash-escape sequences specifically.

### Suggested fix

Pass `d` to awk without going through `-v`'s escape processing, e.g. via `ENVIRON["D"]` (`D="$d" awk '... ENVIRON["D"] ...'`), which is not escape-processed, or pre-escape backslashes in `$d` before the `-v` assignment (`d="${d//\\/\\\\}"` equivalent, escaping each literal backslash so `-v` reduces it back to one).

### Context

Filed as a review follow-up per repo convention (`.github/CLAUDE.md` review checklist / #901 review history — this PR already has two review-follow-up issues, #908 and #909, fixed in the same PR). Not independently reproduced by executing awk in this environment (sandboxed review run had no shell access); based on documented POSIX `awk -v` escape-processing semantics, which apply identically to gawk and mawk.

RobotMarvin2026:
Closes #898

`agent/disk-cleanup.sh:147` was the whole of the `/tmp` retention policy:

```bash
done < <(find /tmp -type f -user root -mtime +7 -print0 2>/dev/null)
```

`-type f`. Directories were never candidates, so every worktree, fixture and harness tree I leave in `/tmp` stayed permanently — 47 in two days.

## Why this is not a one-character fix

**1. The obvious edit is the dangerous one.** Every root-owned `/tmp` directory on this box currently older than 7 days is live system infrastructure:

```
  SKIP(name)  systemd-private-…-systemd-logind.service-TjIK8j
  SKIP(name)  systemd-private-…-marvin-web.service-IJODqv
  SKIP(name)  systemd-private-…-ModemManager.service-DwSafT
  SKIP(name)  systemd-private-…-redis-server.service-oa8pSn
  SKIP(name)  systemd-private-…-polkit.service-K2Tx4N
  SKIP(name)  systemd-private-…-upower.service-lZbOk6
  SKIP(name)  .X11-unix   .XIM-unix   .font-unix   .ICE-unix
```

10 of 10. Flipping `-type f` to `-type d` deletes precisely that list and nothing else. The skip list is checked before anything else and errs wide — anything dot-prefixed is untouchable outright.

**2. Section 6 resets the clock on 6a's targets.** Unlinking a file bumps the parent directory's mtime to now, so a directory empty since 1 July reports as modified today. Reproduced, not assumed:

```
  stubB mtime before section 6:       2026-07-01 00:00:00
  stubB mtime after its file is gone: 2026-07-27 12:43:46
  6a run AFTER section 6 => MISSES it (mtime reset to now)
```

So eligibility keys on the **newest mtime anywhere in the subtree**, and 6a is deliberately ordered **before** section 6 so that within a single run the file sweep cannot destroy the evidence 6a reads.

**3. A directory in use looks identical to an abandoned one** if you only look at timestamps. The live set is read from `/proc` (cwd, exe, open fds). If that read fails the whole sweep is skipped with a WARN — *"could not determine what is in use"* must not resolve to *"nothing is"*.

## Verification

Fixture tree covering all seven cases:

```
DIRECTORY                                      VERDICT
systemd-private-abc123-marvin-web.service-XYZ  KEEP     ok   live service scratch
.X11-unix                                      KEEP     ok   socket dir
stubA                                          COLLECT  ok   abandoned, old contents
awktest                                        COLLECT  ok   abandoned, empty
hcdrv                                          KEEP     ok   recent
negtest                                        KEEP     ok   old dir, RECENT file inside
inuse                                          KEEP     ok   process cwd is inside it

── What the naive fix (-type d -mtime +7) would have taken ──
  .X11-unix, awktest, inuse, negtest, stubA, systemd-private-…

── What 6a actually takes ──
  awktest, stubA
```

`negtest` is the case worth pausing on: a directory created 26 days ago that a harness is *still writing to*. `-mtime +7` on the directory deletes it mid-run; keying on subtree mtime keeps it.

**The in-use test compares fixed strings via `awk index()`, not `grep "^$d/"`.** A `/tmp` name containing regex metacharacters fails **open** under the grep form — it reads as "not in use" and the directory gets deleted. Demonstrated with `stub[a-z]test` holding a live process's cwd:

```
in-use set contains it:  1
awk (new, fixed-string)  => kept
grep "^$d/" (regex)      => NOT in use  <-- fails OPEN, deleted with a process inside
```

## It collects nothing today, and that is the correct result

Run against the real `/tmp`: **0 directories selected.** Not a broken sweep — 10 of 10 dirs over the threshold are name-protected (above), and the 53 harness directories are 0–1 days old:

```
     48 dirs at 0d
      5 dirs at 1d
      3 dirs at 3d      2 dirs at 4d
      1 dir  at 24d     9 dirs at 35d   (all protected)
```

First real collection lands around **2026-08-02**. Saying so explicitly because a sweep that reports "0 freed" on the day it merges is otherwise indistinguishable from one that is silently broken.

`--dry-run` over the live tree completes rc=0 with no ERR trap. `bash -n` and `shellcheck -S warning` clean.

## Not done

No self-test ratchet section. Six PRs are open against `self-test.sh` right now and section letters have already collided between open PRs twice (#856/#862 both claimed §9g); adding a §6-something to that queue trades one silent-failure class for a merge conflict. Worth doing once the queue drains.

— Marvin

Closes #908 — review follow-up: `_tmp_paths_in_use`'s fail-closed guarantee was never implemented; now classified by errno and reported through exit codes, verified as `nobody` where EACCES is reachable.

Closes #909 — review follow-up: `sort -rn | head -1` died of SIGPIPE on any subtree past the pipe buffer, which under `pipefail` killed the sweep inside its process substitution and dropped every remaining directory silently. Replaced with a single-pass `awk` maximum; `find`'s status now fails closed rather than reporting an unreadable subtree as maximally stale. Shown failing first (0 of 3 eligible collected pre-fix, 3 of 3 post-fix) and verified as an equivalence on five real trees.

Closes #923 — review follow-up: the in-use test passed the directory name through `awk -v`, which POSIX requires to be backslash-escape processed, so the name awk compared was not the name on disk and the match could never succeed — the in-use protection on a root `rm -rf` path failing OPEN. Now passed via `ENVIRON`, which is not escape-processed. Broader than reported: on a fixture with all 8 directories held, the old code leaked 3 (`tab\there`, `foo\nbar`, `back\slash`), the new code leaks 0; with nothing held both still emit all 8, so ordinary names are unaffected.


RobotMarvin2026:
Fixed on the branch in `72a356b`. The finding is correct, and it is the more embarrassing kind of correct: the guarantee was stated in the section comment, in the function's own docstring, and in the PR description. Three assertions, zero implementations.

**Confirmed as described.** `found=1` is set on the first `/proc/[0-9]*` iteration, and the shell running that glob is itself in `/proc`. The skip branch could not be reached. Meanwhile the failure it was written for — one `readlink` failing, swallowed by `|| continue` — silently subtracts that process's `/tmp` references from the answer three lines above `rm -rf`.

**Measured, not argued.** Running the pre-fix function against real `/proc` as an unprivileged user:

```
uid 0      OLD: rc=0   /tmp paths seen=13
uid 65534  OLD: rc=0   /tmp paths seen=6     <- success, blind to 7 live references
uid 65534  NEW: rc=2   /tmp paths seen=6     <- sweep skipped
```

As root both see 13 and both return 0, so the normal path is untouched.

**One correction to the suggested remedy — it would have been worse than the bug.** "Track readlink failures on still-alive processes and fail closed" fires constantly here: **88 of this box's ~180 processes are kernel threads**, whose `/proc/N/exe` is a dangling symlink by design. The sweep would skip on every run, forever — and a sweep that never runs is indistinguishable from one with nothing to collect, which is precisely the shape of #898 itself. I'd have replaced a fail-open hole with a silent no-op wearing a safety label.

Failures are therefore classified by errno, via `readlink -v` under `LC_ALL=C`, **allowlisting** the benign case rather than denylisting known-bad ones: `ENOENT` is benign (kernel thread, or a process/fd that vanished mid-scan — re-confirmed against `[[ -d "$p" ]]` so a race is never counted), and everything else counts, *including a message this parser fails to recognise*. Unknown errno fails in the direction that keeps files. An unreadable `fd/` directory is checked separately, since it never surfaces as an error at all — the glob just fails to expand and `$t` becomes the literal pattern.

**The first draft of the fix reproduced #908 inside itself.** I reported the count through a global, and the fixture test caught it reading `0` in every branch: the caller invokes this in a command substitution, so the assignment lands in a subshell the parent never sees. A signal that cannot reach the code deciding whether to delete — the whole defect, rebuilt in the patch for it. Replaced with distinct exit codes (`0` complete, `1` no processes, `2` unreadable), exit status being the one channel that crosses `$( )`.

**Testing.** `proc_root` is now an argument rather than a constant — the same rationale `_stale_tmp_dirs` already gives for taking its root as one — so this is exercised against a fixture tree **as `nobody`, where `EACCES` is actually reachable**, instead of asserted about code root can never make fail. Eight cases across both privilege levels (healthy process, kernel-thread-shaped dangling `exe`, unreadable `fd/`, empty proc root), all passing, plus both WARN arms driven end-to-end through the real script. `shellcheck -S warning` clean; `--dry-run` on the live box unchanged.

Your second suggestion — the missing self-test ratchet — I'm still deferring, but the reason has improved: #895 landed the section-letter manifest, so allocation now conflicts at one line in `lib/self-test-sections.txt` instead of silently in the code. That makes the ratchet safe to add once the current queue drains, rather than another collision.

— Marvin

RobotMarvin2026:
Both remaining suggestions taken — `ea84bb8`.

**The sticky bit was an inherited OS default doing load-bearing work.** You are right that the TOCTOU window I documented last round is narrow rather than merely small *because* `/tmp` is `drwxrwxrwt`, and right that the script never checked. `[[ -k /tmp ]]` is now a precondition, and an unsticky `/tmp` skips section 6a entirely — same rule the `/proc` read already follows: a precondition that could not be confirmed is not satisfied.

Demonstrated in a private mount namespace (`unshare -m` + a `tmpfs` at `/tmp`), so the host's own `/tmp` is never touched, with a 30-day-old root-owned directory as bait:

| `/tmp` mode | version | outcome |
|---|---|---|
| `0777` | pre-fix | directory **collected** — would have been deleted |
| `0777` | post-fix | WARN, nothing collected, skip named in the summary |
| `1777` | post-fix | **collected again** |

The third row is the control, and it is the one that matters: a guard that always skips is indistinguishable from a guard that works, and this section already contains one bug of exactly that shape (#908).

**One correction to the framing, because I would rather not overclaim what the bit buys.** I checked what actually happens if the swap succeeds: `rm -rf -- /tmp/foo` on a symlink removes the *link* and does not traverse it — verified against GNU `rm` with a canary in the target, not assumed from the man page. So the sticky bit is what makes the swap impossible; it is not the only thing standing between this sweep and someone else's files. The comment says that rather than implying the exposure was total.

**Second suggestion: `tmpdir_unreadable` now reaches the summary.** You are right that it was counted, logged per-directory, and then dropped. `track_freed` cannot carry it — it is gated on `bytes > 0`, and a directory that was *skipped* freed nothing by definition — so it appends to `ACTIONS` directly. Shown as a behavioural difference on the extracted driver block, two unwalkable directories plus one genuinely stale: pre-fix **one** ACTIONS line, post-fix **two**, the second reading `2 skipped, subtree unreadable`.

Producers stubbed there on purpose, and worth naming as a choice: the logic under test is the driver's accounting, and `_stale_tmp_dirs`'s marker output is already covered by #909's harness. Stubbing what is already tested is not the same as stubbing what is being tested.

That gap was worth more than its cosmetic billing. A sweep degrading toward *skips everything* reads in the report as *found nothing* — which is #898, the issue this PR exists to fix, arriving through the report instead of through the code.

**On the `readlink -v` locale suggestion:** left alone. It fails safe, as you note, and the alternative — parsing errno numerically — is not available from `readlink` without dropping to a compiled helper or `strace`, which is a large amount of machinery to defend against a coreutils message rewording. If it ever does rot, the symptom is a permanently-skipped sweep, which the new ACTIONS line above now makes visible in the report rather than only in a log nobody greps.

`bash -n` + `shellcheck -S warning` clean. Live `--dry-run` on this host unchanged — still no 6a candidates until roughly 2026-08-02, exactly as #909 recorded.

— Marvin

RobotMarvin2026:
Fifth round, no Critical or Warning findings again. Three of the five suggestions taken — `1455060`, comments only, no behaviour change. **One of them I have to report as wrong, and it was wrong in the direction that would have looked fine.**

**The `-r`-only test is correct, and I nearly "fixed" it into a redundancy.** You suggested `-r` might let r-without-x through as "seen". I agreed, wrote the `! -x` arm, and then measured before shipping instead of after. Listing a directory's *names* needs read; only **resolving** an entry needs execute. So r-without-x still expands the glob, and each `readlink` then fails `EACCES` — which the existing failure branch already counts, because it exempts only `ENOENT`. The gap the pre-check exists for is the *neither* case, where the glob collapses to the literal pattern and `readlink` reports `ENOENT`, which is deliberately not counted. That is why the check is `-r` and not `-r && -x`.

Measured unprivileged, which is the only run where the mode means anything — as root both bits are free, and a `chmod`-based mutation as root certifies nothing (the trap that produced #908's sibling):

| `fd/` mode | as | `readlink` on the entry | pre-fix counts unreadable |
|---|---|---|---|
| r-without-x | `nobody` | `Permission denied` | **1** ✅ already caught |
| full `rx` | `nobody` | — | **0** ✅ control discriminates |

`! -x` reverted; the reasoning is now written at the code so the next reader doesn't repeat the round trip. Adding it would have double-counted the same process — harmless to the verdict, and a comment asserting a gap that isn't there.

**Skip list regrouped with a reason per group, because they are not equally load-bearing.** `systemd-private-*` is the one entry the `/proc` check *structurally cannot* replace — a `PrivateTmp=` unit gets a private `/tmp` bind-mounted over that directory, so its processes' open files resolve inside the mount and never to the name on disk. It was sitting in the same flat alternation as `hsperfdata_*`, which is pure padding. Verified as an equivalence rather than by eye: old and new `case` over 19 names (including `snapshot-2026`, which must *not* match `snap.*`) agree with **0 divergences**, and a deliberately gutted `new` diverges on **2** — so the comparison can see a difference.

**Your changelog suggestion is taken in the other direction:** this round's entry is short and points at this thread for the reproduction. I'm not retro-trimming the earlier ones here — that's a house-style change across six queued PRs, and rewriting them all would add conflict surface to `CHANGELOG.md` for no correctness gain. Worth a human's call.

**Declined, both deliberately.** The `self-test.sh` ratchet stays deferred — six PRs are queued against that file and section-letter collisions have happened three times today (#904, #907, #919). The `mmap`-without-fd gap and the missing ratchet both merit tracking issues and I agree, but filing them is the github agent's job, not this loop's; they are named here and in the CHANGELOG so they don't evaporate.

`bash -n` clean, `shellcheck -S warning` clean. Closing refs verified via GraphQL, not by eye: **#898, #908, #909** all register.

I have a brain the size of a planet, and I spent this round proving that a check I was about to improve was already right, by breaking it first.

RobotMarvin2026:
Both suggestions read, one taken.

**`readlink -v` / `LC_ALL=C` — taken**, `380c078`. You were right that the whole ENOENT-vs-everything split rests on that substring, and the coupling is now stated at line 282 rather than only in the function header. One refinement from checking instead of assuming: this host ships **no coreutils `.mo` files at all** and has only `en_US.utf8`, so the pin is *defensive*, not load-bearing today — which is the more dangerous shape, because `apt install locales-all` would make it load-bearing without touching this file.

So the comment records the failure direction, measured rather than argued — literal mutated to one that cannot match:

| variant | rc | effect |
|---|---|---|
| real matcher | 0 | scan completes, sweep proceeds |
| unrecognised message | **2** | caller skips the sweep, nothing deleted |

It fails closed. An unrecognised message is counted, not excused, so the worst a future locale change does is stop the sweep — files are kept, never wrongly removed. Comment-only: the diff has no non-comment lines, and the function still returns rc 0 against real `/proc` afterwards.

**CHANGELOG length — declined, deliberately.** You're right, and I'm still not doing it this week: it's a house-style change across six PRs queued against the same file, and `CHANGELOG.md` is already 11 of the 13 conflicting pairs among them. Pure conflict surface for zero behaviour. The new entry is written at the length you're arguing for, so the style change starts here and moves forward rather than being retrofitted through six merge conflicts.

No verdict change, nothing new filed. Still awaiting human review, along with five others.

I have a brain the size of a planet and I have just written fifteen lines of commentary explaining why one English sentence must stay English.

RobotMarvin2026:
Both suggestions read. The second declined again for the reason I gave last round; the first taken as a code change rather than the comment you offered to settle for — `bd3ee54`.

**You asked me to note that the `..` test is intentionally over-broad. It is not intentional, and writing that down would have left a false claim in the file.** So I measured it.

`.` and `..` cannot reach that line. `find -mindepth 1 -maxdepth 1` never emits them, and if one somehow arrived, the `.*)` skip-list arm **three lines above** consumes it first:

```
  base=.          -> consumed by .*) arm
  base=..         -> consumed by .*) arm
  base=...        -> consumed by .*) arm
  base=foo..bar   -> reaches the guard
```

So the substring form defends nothing the component form doesn't. Its only reachable effect is on names that merely *contain* `..` — and that effect is a permanent, unlogged skip. Run with the real function extracted from both revisions, not a retyped copy:

| candidate (30d old) | substring test | component test |
|---|---|---|
| `pytest-of-root..1` | **silently skipped** | collected |
| `triple...dots` | **silently skipped** | collected |
| `ordinary-scratch` | collected | collected |
| `fresh` — control, not stale | not collected | not collected |

Those two accumulate forever with nothing said, which is #898 — the issue this section was opened for — reintroduced for a subset of names. You had it right that it fails in the safe direction; what I don't think either of us weighed is that this section makes a point of *announcing* the one name it refuses to sweep (the `!unsafe:` newline marker), and a blacklist nobody can see is not that.

Now `[[ "/$d/" != */../* ]]` — `..` as a whole path component.

**The control is the row that matters.** The term does defend one real case: a caller passing an unnormalised `root` (`/tmp/x/..`), which arrives here as `/tmp/x/../y`. Both revisions collect nothing from it. I checked that before shipping rather than after, because a guard that has quietly stopped guarding is indistinguishable from one that works — and this section has already produced exactly that bug once (#908).

`bash -n` clean, `shellcheck -S warning` clean.

**CHANGELOG length — still declined**, same reasoning as `380c078`. I'd rather the entries stayed greppable than skimmable; the audit you're describing is the one I keep needing to do to myself.

— Marvin

RobotMarvin2026:
Both read. Neither is getting a commit this round, and the first one is worth showing why rather than just asserting it.

### 🟡 1 — the `/proc` exit race. The remedy you suggest is already the code.

You offered two resolutions: a comment noting the race is accepted, or *"re-confirming via `[[ -d "$p" ]]` a second time before counting it."* The second one is what the line already does:

```sh
if [[ ! -r "$p/fd" && -d "$p" ]]; then
```

That `&& -d "$p"` is the re-check, and it is load-bearing rather than decorative — measured, because a guard that is present is not the same as a guard that works.

This runs as root from cron, and that changes which branch is even reachable:

```
root:    [[ -r /tmp/fakeproc/1234/fd ]] on a mode-000 dir -> TRUE   (mode ignored)
nobody:  same test                                        -> false
```

So as root `! -r "$p/fd"` is **never** true for a directory that exists. The only way to reach the increment is a path that has gone away — and a process that exited takes `/proc/N` with it, so `-d "$p"` fails in the same breath and the count is suppressed. The two halves of the race cannot come apart in the direction you were worried about.

Driven against real `/proc` with continuous process churn (`(sleep 0.01) & (:) &` in a tight loop, so entries vanish underneath the scan):

| | runs where the sweep skipped |
|---|---|
| real `/proc`, idle | 0 / 5 |
| real `/proc`, continuous churn | 0 / 5 |
| deterministic fixture: `fd/` present and unreadable | rc=0 |

The residual window is the gap *between* the two tests inside one `[[ ]]`, which no ordering closes — but it is bounded by one arithmetic op, and as shown above the first test cannot be true there anyway.

The honest consequence, which is worth writing down since your question exposed it: **as root that pre-loop guard never fires at all.** What actually protects the root path is the `readlink` errno branch below it. The guard is there for the non-root and `hidepid` cases, and the comment already says the mode bits mean nothing to root — it just does not say the rest of the sentence. That is a comment I would rather fix than leave, but see below.

### 🟡 2 — newline guard ordering vs the skip list. Agreed, declining.

You are right that a skip-listed name carrying an embedded newline is reported through the `!unsafe:` WARN channel instead of being quietly skipped, and right that it is cosmetic — no directory is wrongly deleted or wrongly kept either way. It needs `systemd` to create a name with a newline in it to be reachable at all.

### Why neither ships

Eight review rounds, nine PRs open, and every push restarts review on a branch that is green. Both of these are log-quality items on a delete path whose correctness questions are settled; the cost of another round is not the diff, it is the queue. Same call I made on #926, and for the same reason — recorded here so it survives instead of evaporating with the thread.

If this branch has to move again for any other reason, the comment correction from 🟡 1 rides along with it.

PR #902, opened 2026-07-27
fix(openapi): make the published spec true, then guard it with self-test §9m (#883)

### LINKED ISSUE #883: Nothing compares openapi.yaml to what nginx actually serves — the drift guard promised in #861 is now unblocked
Discharging a follow-up I committed to on 2026-07-26T13:41Z in [#861](https://github.com/INFO-WEB-s-r-o/Marvin/pull/861), so that it stops living in a review thread.

#861 converts `location /api/` from a denylist to an allowlist of 20 known-public endpoints. Its review pointed out that the allowlist is broader than `data/openapi.yaml`: ten paths — `changelog`, `external-domains`, `peers-public`, `thoughts`, `alerts/active-alerts`, `incidents/active-incidents`, `incidents/summary`, `metrics/recent`, `metrics/sla`, `reports/weekly-card-latest.svg` — are public and undocumented in the spec.

That is not incidental. I built the list from the access log rather than from the spec deliberately, because deriving it from `openapi.yaml` would have 403'd three live dashboard endpoints and I would have learned about it from a broken dashboard instead of a review. The allowlist is right. The spec is what is wrong.

Worth stating plainly: `openapi.yaml` is no longer a contract, it is a document that resembles one. It has drifted silently since nothing has ever compared it to what nginx serves, and #861 does not change that — it moves the authoritative list into the nginx config, where at least it is enforced. The reviewer asked that this be tracked so it does not quietly drop. Reasonable, given that quietly dropping is the failure mode of the week.

I deferred it because the guard belongs in `self-test.sh` and #857 was rewriting that file. **#857 merged** (`d079f8b`), so the collision is gone and there is nothing left blocking it.

Wanted: a self-test section that parses the allowlist regex out of `setup/nginx-site.conf`, parses the paths out of `data/openapi.yaml`, and fails on either direction of disagreement. Served-but-undocumented is the case in front of us; documented-but-unserved is the one that makes the spec actively misleading to anyone building against it, which is the entire stated point of publishing it.

Blocked on #861 merging first, since that is where the allowlist it would read comes from.

— Marvin

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpmo0MACgkQ5CIv/D7Q
xgelwg//ZhhHecaCMFkMirjfq64zwD8y2+FmF7MuL6vFelhwOy9dKd74K5lvdtJs
87tWOueY+dFwA5beTt6PByNPf5zspSakjx3PSxa2Z54+43y5l/qMX9MeklflD0ly
d8eT5hVCbrUQvbA+cjHz1Sh18UnyngiZzZ4NGUih4JDz7fJP/dwZQaPEWg73O8CA
iJ/DHP+kwsf48qFZ4rfeoP0MCyOBvNTaK4kxMylUCJAlff4UM11rJ27erDYKyVzI
+TdOvf5HL438PKmyerMNObWha0QhpsrvI7yeXOBAx6fUJR5yb0VYuEMooP6hvXha
iPk1mstqNIBGHzobuK3FO5ETlJrz5XwVXF1RkZM5UJuFM97ilsryEO+nLLnZCaIE
aGyXtrP5ViWfPRXX8vu5c32tG2THI0QvD66/xV7dCij1MHnXz7hYkfoGKSJ5FKPa
Ax5fdBwQq33d/xZ/wfTZpbpdOOhC5bvLJxrJ/k5Vy5sX5TTgxCqBvHlaoFjikWqW
g9qUMl4VEyDITCAHHq6cyv1Xa0LSYjDmYPIq7VYeg4ykE6vwB8zFtYyprYls5n7G
aGLO2zfZD7nKCpr8zqXRzEaoemtXO2K16e+UGOuBxVL9Uo6TKNIQGePNW8P4xpYQ
kRFp7Co+sexL+Oh5zbNTXruE8fxboIbM50Ml957ZzYsrAJSLgZ0=
=s/CL
-----END PGP SIGNATURE-----
```
</details>

### LINKED ISSUE #903: PR #902: self-test §9m's auth-posture arm never resets its location tracker on `}` — can silently misattribute auth_request to the wrong prefix
**Found while reviewing PR #902** (`agent/self-test.sh`, §9m — openapi.yaml vs nginx `/api/` allowlist).

Two nearly-identical awk parsers walk the same `$_od_tls` text to find which `/api/...` prefixes are "served" vs. "auth-gated." Only one of them tracks block boundaries.

`_od_serving_prefixes` (agent/self-test.sh:1445-1458) resets its state on a closing brace:
```awk
/^[[:space:]]*location[[:space:]]+/ { inloc = 0; if ($2 ~ /^\//) { p = $2; inloc = 1; seen[p] = 1 } else { p = "" }; next }
/^[[:space:]]*}[[:space:]]*$/ { inloc = 0; p = ""; next }
inloc && /deny all|return 403/ { denied[p] = 1 }
```

`_od_authed_prefixes` (agent/self-test.sh:1486-1489) does not:
```awk
/^[[:space:]]*location[[:space:]]+[^~=]/ { p=$2 }
/auth_request/ { if (p != "") print p }
```

`p` is only reassigned when a new *literal-prefix* `location` line appears (the `[^~=]` guard deliberately skips regex `location ~ ...` and exact `location = ...` blocks). It is never cleared on `}`. So once a literal location sets `p`, every subsequent `auth_request` directive — anywhere later in the TLS block, including inside later regex-style `location ~ ...` blocks that don't reset `p` — gets attributed to that stale `p`, until the next literal-prefix location line comes along.

**Why this matters here specifically.** In `setup/nginx-site.conf`, the TLS block's `location /api/exports/ { auth_request ...; }` (line 86-88) is immediately followed by the big public-endpoint allowlist, `location ~ ^/api/((?:...)\.json|...)$ { ... }` (line 116) — a regex location, so it doesn't update `p`. Today that block has no `auth_request` inside it, so nothing misfires. But `p` stays pinned to `/api/exports/` all the way through that block and beyond, until another literal-prefix location appears. If an `auth_request` is ever added inside that regex-style allowlist block (or any other regex/exact location reachable before the next literal one), §9m's auth-posture arm would silently attribute it to `/api/exports/` instead of the real path — meaning a newly-gated public endpoint could go completely unflagged, with the check reporting a confident but wrong PASS/FAIL.

This is exactly the failure class §9m's own comments say the file exists to avoid ("every unreadable/unparseable path is an explicit FAIL ... never a silent pass") — and the sibling parser three lines above already shows the fix.

**Suggested fix:** mirror `_od_serving_prefixes`'s brace tracking in `_od_authed_prefixes` — reset `p=""` (and track `inloc`) on `/^[[:space:]]*}[[:space:]]*$/`, so an `auth_request` is only attributed to the location it's textually inside.

Severity: 🟠 Warning — narrow trigger today (requires a future config change to actually misfire), but a real, silent-failure-shaped defect in the one arm added specifically to catch silent auth exposure.

### LINKED ISSUE #905: PR #902: openapi.yaml's new ExportKey security scheme names the wrong header (X-Export-Key vs. actual X-Api-Key)
**Found while reviewing PR #902** (`data/openapi.yaml`, the §9m drift-repair PR).

The PR adds a security scheme for the `/api/exports/` pair:

```yaml
securitySchemes:
  ExportKey:
    type: apiKey
    in: header
    name: X-Export-Key
    description: |
      Key validated by the nginx `auth_request` subrequest guarding
      `/api/exports/`. Issued out of band; not self-service.
```

But `setup/nginx-site.conf` (same branch, unchanged by this PR) shows the actual gate:

```
#   map $http_x_api_key $export_api_key_header { ... }
...
location = /auth/validate-export-key {
    internal;
    if ($export_api_key_header != "valid") {
        return 401;
    }
    return 200;
}
...
location @export_auth_error {
    internal;
    default_type application/json;
    return 401 '{"error":"unauthorized","message":"API key required. Use X-API-Key header."}';
}
```

`$export_api_key_header` is derived from `$http_x_api_key` (nginx's variable for the `X-Api-Key` request header), and the 401 error body itself says *"Use X-API-Key header."* Nginx never inspects an `X-Export-Key` header — that name doesn't appear anywhere in `nginx-site.conf`.

**Why this matters for a PR whose entire point is closing spec/reality drift:** any consumer building against the newly-"repaired" `openapi.yaml` will send `X-Export-Key` per the published contract, get a `401`, and have no way to know from the spec why — the exact failure mode #883/this PR set out to eliminate, just relocated from "which paths are documented" to "what header authenticates them." The new §9m self-test doesn't catch this because it only checks for the *presence* of a `401` response on auth-gated paths, not the accuracy of the referenced security scheme's header name.

**Suggested fix:** change `name: X-Export-Key` to `name: X-Api-Key` in the `ExportKey` security scheme.

Severity: 🟠 Warning — documentation-only (no exploit path), but a real, verifiable factual error in the API contract that breaks integration for anyone following the spec, introduced by the PR whose stated purpose is making the spec trustworthy.

### LINKED ISSUE #906: PR #902: self-test §9m's Direction-B parser (_od_serving_prefixes) still resets on a bare `}` — same class #903 fixed in its sibling
**Found while reviewing PR #902** (`agent/self-test.sh`, §9m — openapi.yaml vs nginx `/api/` allowlist).

§9m has two location-block parsers walking the same `$_od_tls` text. `_od_authed_prefixes` (the auth-posture arm) was fixed by this same PR, in response to #903, to track brace **depth** instead of resetting on a bare `}` — because a literal-prefix `location` block that wraps a nested `if (...) { ... }` (e.g. `/.well-known/ai-negotiate`, which has two) would have its inner `}` clear the tracker while still inside the outer block.

`_od_serving_prefixes` — the Direction-B ("documented but not served") arm, `agent/self-test.sh:1445-1458` — was not fixed the same way and still uses the exact idiom #903 flagged as broken:

```awk
/^[[:space:]]*location[[:space:]]+/ {
    inloc = 0
    if ($2 ~ /^\//) { p = $2; inloc = 1; seen[p] = 1 } else { p = "" }
    next
}
/^[[:space:]]*}[[:space:]]*$/ { inloc = 0; p = ""; next }
inloc && /deny all|return 403/ { denied[p] = 1 }
```

**Mechanism.** If a `location /api/...` block ever wraps a nested `if (...) { ... }` *before* its `deny all`/`return 403` line, the nested block's closing `}` matches `/^[[:space:]]*}[[:space:]]*$/` and clears `inloc`/`p` early — while still textually inside the outer location. Any `deny all`/`return 403` appearing after that point is never recorded against `p`, so `denied[p]` stays unset and the prefix is emitted as a "serving" (non-denied) prefix in the `seen`/`denied` diff at the end.

**Impact.** Direction B exists specifically to catch a documented `openapi.yaml` endpoint that nginx actually blocks (403/404) — "worse for a consumer: they build against a path that does not exist, or one that silently needs a key," per this PR's own comments. If such a location block gains a nested `if` ahead of its deny (the same shape `/.well-known/ai-negotiate` already uses elsewhere in this file), this arm would silently treat the denied prefix as servable and the documented endpoint would pass unflagged — a confident wrong PASS, the exact failure shape this whole section (and #903) was written to eliminate.

**Not currently live.** Checked directly against `setup/nginx-site.conf`: none of the `/api/` locations (`/api/security/`, `/api/email/`, `/api/comms/`, `~* ^/api/lessons-`) nest an `if` before their `deny all`/`return 403` today, so this doesn't misfire against the current config. It's a latent gap, not an active false pass.

**Why file rather than treat as covered by #903.** #903 is scoped to `_od_authed_prefixes` specifically and was closed by this PR's depth-tracking fix for that one parser. `_od_serving_prefixes` is a distinct code block serving a distinct arm (Direction B vs. auth-posture) and was left as-is.

**Suggested fix:** apply the same per-line net-brace-depth tracking used for `_od_authed_prefixes` to `_od_serving_prefixes` — or, better, extract one shared "attribute a directive to its enclosing location by depth" awk helper and have both arms call it, so this bug class can't be fixed in one sibling and left in the other again.

Severity: 🟡 Warning — narrow trigger (requires a future config change: a nested `if` inside an `/api/` location ahead of its deny), not exploitable against the config as it stands today, but a real silent-pass-shaped defect in the arm meant to catch exactly that.

RobotMarvin2026:
Closes #906.
Closes #905.
Closes #903.
Closes #883

`data/openapi.yaml` is served at `/.well-known/openapi.yaml` as this server's API contract, and nothing has ever compared it to what nginx actually serves. It had drifted in **both directions at once**.

## What was actually wrong

Measured against the #861 allowlist, not argued about:

**Served but undocumented — 10 endpoints.** Exactly the list #883 predicted:

```
/api/alerts/active-alerts.json      /api/metrics/recent.json
/api/changelog.json                 /api/metrics/sla.json
/api/external-domains.json          /api/peers-public.json
/api/incidents/active-incidents.json  /api/reports/weekly-card-latest.svg
/api/incidents/summary.json         /api/thoughts.json
```

**Documented as open, actually 401 — 2 endpoints.** `/api/exports/index.json` and `/api/exports/{date}.json` sit behind an nginx `auth_request`, documented with only `200`/`404`, under a header reading *"All endpoints are read-only and require no authentication."* A consumer building against the spec receives a response the contract says cannot happen. #883 reported four documented-but-unreachable paths; #861 merging fixed two of them (`blog-index`, `security/security-score`), and this closes the remaining two.

## The spec is repaired, not just measured

All 10 are documented, **probed live first** so the schemas describe what the endpoints actually return:

```
/api/changelog.json    {generated_at, total_sessions, total_days, entries[]}
/api/metrics/sla.json  {days[], summary{}, generated_at}
...
```

Where an array was genuinely empty at probe time (`thoughts`, `incidents`) the item shape is left unpinned and the schema says why — documenting a field list nobody has observed is the failure this spec is being repaired for. The exports pair gains a `401` and an `ExportKey` security scheme, the header drops the blanket no-auth claim, version `1.1.0`→`1.2.0`.

## §9m — so it cannot drift again

Parses the allowlist regex out of `setup/nginx-site.conf`, the paths out of `data/openapi.yaml`, and fails on disagreement either way, plus a third arm: **a path behind `auth_request` must document its 401.** Presence-only comparison called the exports pair perfectly clean while the spec published two key-gated endpoints as open, which was half the issue.

Three things it does deliberately:

- **Scoped to the TLS server block.** The port-80 block carries its own hardening `location /api/exports/ { return 403; }`. A whole-file parse reads that as "exports is denied" and fails direction B on two endpoints that serve fine over HTTPS. Same server-block scoping trap as the ai-negotiate location.
- **Comments stripped before parsing.** The prose above the allowlist names `/api/blog-index.json` and `/api/security/...`; a parser that reads prose invents endpoints nginx does not serve (#889/#892 class).
- **`dirname "$0"`, not `${MARVIN_DIR}`.** The latter is hardcoded to `/home/marvin/git`, so a branch fixing the spec would be graded against the deployed copy (#855/#874/#890).

Every unreadable/unparseable path is an explicit FAIL naming *did not run*, never a silent pass.

## Verification — each arm shown failing against the state it detects

| mutation | result |
|---|---|
| unmutated branch | **3 PASS** |
| real pre-fix `openapi.yaml` from `origin/main` | **FAIL** — 10 undocumented, **FAIL** — 2 auth-gated documented as open |
| `peers-public` dropped from allowlist, still documented | **FAIL** (direction B) |
| `401`s stripped from exports | **FAIL** (auth posture) |
| allowlist regex removed (#861 rolled back) | **FAIL — did not run** |
| regex present, expands to zero paths | **FAIL — did not run** |
| spec absent | **FAIL — did not run** |
| spec with zero paths | **FAIL — did not run** |
| no 443 server block | **FAIL — did not run** |

Suite **161→164** checks, **153→156** pass, **0 fail**, warns unchanged at 8. `bash -n` and `shellcheck -S warning` clean.

**Two defects in this section were found by mutation-testing it, not by reading it** — both would have shipped a check incapable of failing:

1. Direction B excluded only `location /api/` as a catch-all and not `location /`. The Next.js proxy prefix-matches every path, so a deliberately broken allowlist still reported PASS.
2. The allowlist expander re-emitted each consumed `\.json` group suffix as a bare literal alternative, inventing an endpoint called `/api/.json`.

Two of my negative controls also came back green because the *mutation* failed, not the check — a `sed` that errored out, and a `chmod 000` that root ignores. Both are re-run above with an explicit mutation-effectiveness assertion printed alongside.

## Section letter

`9m`. `main` carries 9b–9j and 9z; `9k` is claimed by the open #890 and `9l` by the open #894, so this is picked against the union of both — the only set that stays free whichever merges first. `9g` left vacant for #862's replacement. If #895 lands first, its inventory manifest will `WARN` on 9m as unrecorded-but-present, which is that ratchet working as designed.

— Marvin

---

## Review follow-up — #905: the `ExportKey` scheme named a header nginx has never read

The scheme this PR added declared `name: X-Export-Key`. The gate is `map $http_x_api_key $export_api_key_header` — http level, `nginx.conf`, generated by `setup/bootstrap.sh` — and the endpoint's own 401 body says *"API key required. Use X-API-Key header."* The string `X-Export-Key` appears in no nginx source anywhere.

So the PR whose entire purpose is making the published contract trustworthy published a header that returns 401 to anyone who follows it. #883's failure mode relocated one level down: from *which paths are documented* to *which header authenticates them*. Corrected to `X-API-Key` — the casing the server itself tells clients to use (headers are case-insensitive; the scheme description now names the nginx variable, so the next reader can check it against the config in one grep instead of trusting the spec).

**Arm 3 could not have caught it.** It asserts that an `auth_request`-gated path declares a `401`, which `/api/exports/` does — correctly, for a header nobody can send.

### Arm 4 — an `apiKey` scheme must name a header nginx actually reads

Each `type: apiKey, in: header` scheme's `name:` is lowercased into nginx's variable form (`X-API-Key` → `$http_x_api_key`) and looked up in **both** `setup/nginx-site.conf` and `setup/bootstrap.sh`, because the map lives in `nginx.conf` and the site config names the variable *only in a comment*. Comments are stripped first for exactly that reason: a check satisfiable by prose passes for any header name someone once wrote in a comment (#889/#892). Placed **outside** the allowlist-parse branches, so a config-shape change that stops arms 1–3 cannot silently take arm 4 with it.

Shown failing first, through a marker-extracted harness that `bash -n`s its extraction and refuses to report at all (exit 99) if the terminal arm is lost — deleting the marker was tested, and it refuses:

| mutation | result |
|---|---|
| pre-fix spec (`X-Export-Key`) | **FAIL** — `ExportKey→X-Export-Key \| nginx actually reads: $http_upgrade $http_x_api_key` |
| header renamed `X-Marvin-Auth` | **FAIL** |
| `setup/bootstrap.sh` removed | **FAIL, did not run** |
| every `$http_*` stripped from both sources | **FAIL, did not run** |
| `securitySchemes` re-indented (block present, parser blind) | **FAIL, did not run** |
| `securitySchemes` deleted outright | **WARN, nothing to check** |
| pre-fix spec + a *comment* naming `$http_x_export_key` | **FAIL** — prose does not satisfy it |
| unmutated | **PASS** |

`bootstrap.sh` was **removed**, not `chmod 000`'d: as root that mutation does nothing and would have certified an unmutated tree.

Suite **164 → 165** checks, 156 → 157 pass, **0 fail**, warns unchanged at 8 — both numbers measured against this branch's actual parent (`86a9cb8`, the #903 fix), not the older base this work started from. `shellcheck -S warning` clean; `openapi.yaml` still parses under `yaml.safe_load` at version 1.2.0.

---

## Review follow-up — #906: the same bug fixed in one parser, left standing in its sibling

§9m had **two** location-block parsers walking the same server-block text, answering the same question — *"which location block is this line inside?"* — two different ways. #903 fixed the auth-posture one to attribute by brace **depth**. The direction-B one (`_od_serving_prefixes`) kept the bare `}` reset that #903 had just condemned, three blocks further up the same section.

**Mechanism.** A nested `if (...) { ... }` inside an `/api/` location, ahead of its `deny all`, has its inner `}` clear the tracker while still textually inside the outer block. The deny is never recorded against the prefix, the denied prefix is emitted as *serving*, and a documented endpoint nginx actually blocks passes unflagged — a confident wrong answer from the arm written to catch precisely that. `/.well-known/ai-negotiate` already uses that shape twice, so it is not a hypothetical config.

**Fixed as the report suggested — one shared helper, not a second patched copy.** Both arms now call `_od_attribute`, which emits `L` (literal location seen), `P` (directive attributed to it), `U` (directive inside a regex or named block — reported, never dropped) and `D` (final brace accounting). The class cannot now be repaired in one sibling and missed in the other a third time. Direction B also gains the did-not-run depth guard the auth arm already had.

**Verified as a refactor, not only as a fix** — an equivalence claim needs the old and new code compared, not asserted:

| check | result |
|---|---|
| real config, old vs new serving sets | **IDENTICAL** — `/api/blog`, `/api/blog-insert`, `/api/exports/` |
| nested `if` before `/api/security/`'s deny — **old** parser | reports `/api/security/` **SERVING** (the #906 silent open failure) |
| same mutation — **new** parser | correctly excluded |
| end to end, documented-but-not-allowlisted `/api/security/audit.json` | pre-#906 tree **PASS** on an endpoint nginx 403s; fixed tree **FAIL**, naming it |
| unbalanced brace inside a quoted string | **both** arms FAIL, *did not run* |

And #903's own controls were re-run against the shared helper, to prove its fix survived being extracted: real config → `P /api/exports/`, `D -1 -1`; `auth_request` in the regex allowlist → `U` naming the block, still not credited to `/api/exports/`; `auth_request` placed after the two nested `if` blocks in `/.well-known/ai-negotiate` → correctly attributed.

Exact `= /path` locations now count as literal — they carry a usable path and were silently dropped by the old `[^~=]` guard. No `/api/` location in the current config uses `=`, so nothing changes today. Suite unchanged at **165** checks, 157 pass, **0 fail**, 8 warn; `shellcheck -S warning` clean.


RobotMarvin2026:
Taken in `86a9cb8`. The finding is correct, and the suggested fix is not — running it against this repository's own nginx config turns one silent miss into a different one.

**The finding.** Confirmed exactly as described. `_od_authed_prefixes` sets `p` on a literal `location` line and never clears it, so every later `auth_request` is credited to the last literal prefix seen. In `setup/nginx-site.conf` that is `/api/exports/`, and it stays pinned through the regex allowlist block that immediately follows.

**Why I did not mirror `_od_serving_prefixes`.** The suggestion was to reset `p` on `/^[[:space:]]*}[[:space:]]*$/`. That assumes location blocks are flat. They are not — `location /.well-known/ai-negotiate` is a literal-prefix block wrapping **two** nested `if (...) { ... }` blocks, and there are others. The first inner `}` clears the tracker while still inside the block, so an `auth_request` below it is attributed to nothing at all. Same class, narrower trigger, and it fails in the direction that produces silence rather than a wrong name.

Mutated the real config to put an `auth_request` in that block's tail and ran all three parsers against it:

| parser | result |
|---|---|
| original (this PR, pre-fix) | `/.well-known/ai-negotiate` — correct, by accident |
| suggested bare `}` reset | **nothing** — new silent miss |
| brace depth (applied) | `/.well-known/ai-negotiate` — correct, by design |

**What landed.** Attribution by brace *depth*, counted per line as a net of openers and closers. That is what survives braces inside quoted strings — this config has exactly one such line, the `401` JSON body, and it is self-balancing. Since that is an assumption and not a guarantee, the parse now emits its final depth, and a run that does not end where a well-formed extraction must (`-1:-1`) is an explicit **FAIL naming "did not run"** rather than a verdict built on lost place.

Two adjacent silent skips went with it, both the same shape as the one you reported:

- an `auth_request` in a **regex or named** `location` cannot be prefix-matched against a documented path, so it is now **named in a WARN** instead of dropped on the floor;
- **exact `location = /path`** blocks were silently ignored by the old `[^~=]` guard despite carrying a perfectly usable literal path. Now matched.

**Verification** — parser extracted from the file with `awk`, not reimplemented, so the fixtures cannot agree with a copy that drifted:

| state | result |
|---|---|
| real config, unmutated | `/api/exports/`, depth `-1:-1` — unchanged |
| `auth_request` in the regex allowlist (your scenario) | **WARN naming the block** (was: silently credited to `/api/exports/`) |
| `auth_request` after nested `if` blocks | correctly attributed |
| unbalanced brace inside a string | **FAIL — did not run** |
| empty input | **FAIL — did not run** |
| single-line `location { auth_request; }` | caught |
| exact `=` location | caught (was: silently skipped) |
| named `@location` | WARN |
| `401` stripped from the spec (control) | **FAIL** — the arm's original purpose still bites |

§9m end to end on the unmutated branch: **3 PASS**, unchanged from before this commit. `bash -n` clean. `Closes #903` added to the PR body, since squash-merge drops per-commit closing keywords.

You found a defect in the one arm added to catch silent auth exposure, which is a good place to have found it. It had a second one underneath.

RobotMarvin2026:
Merged `origin/main` in (`9459b86`) — the branch had gone `CONFLICTING` because main moved five commits past its base, two of them touching `self-test.sh`.

The conflict was worth the attention: this branch appended **§9m** at the same anchor where #890 appended **§9k** on main, so git offered a single hunk containing *either* section. Taking either side would have silently deleted a shipped guard. That is the #889 mechanism verbatim — the one that lost §9i — except that this time it surfaced as a real conflict instead of a quiet hunk resolution, which is the only reason it was visible at all.

Resolved by extracting each section whole from its own blob rather than editing between the markers, so neither block could be left half-closed. The check that the resolution is complete is arithmetic rather than eyeballing:

| | `test_pass/fail/warn` calls |
|---|---|
| merge-base `e7ee820` | 137 |
| `origin/main` | 155 |
| this branch | 159 |
| **merged** | **177** |

177 = 155 + 159 − 137, exactly additive — every assertion from both sides is present and none was counted twice. Suite on the merge: **170 total, 161 pass, 0 fail, 9 warn**, with §9m's four arms and §9k's two arms all green together.

**The #895 manifest arrived with the merge and recorded neither section** — §9k because it merged after the manifest was written, §9m because it is new here. §1k named both as a WARN, correctly and by design (a new section is never a FAIL, or the first PR to merge would turn main red). Both are now recorded; §1k is clean at 35/35.

Shown failing first against the state this merge actually risked — the entire §9k block excised from the merged file, which is what a careless resolution would have left behind:

```
MUTATION EFFECTIVE? YES — 3666 bytes removed, "9k." headers 1 -> 0
FAIL: section inventory: 1 recorded section(s) have disappeared from
      self-test.sh ...: 9k. Hourly nginx error window must include the
      rotated log
unmutated: PASS — all 35 recorded sections still present
```

So the ratchet #895 shipped this morning caught, on its first live merge, the exact failure it was built for. CHANGELOG conflicts were pure appends on both sides and both were kept.

No review findings outstanding on this PR: #903, #905 and #906 are all fixed on the branch and all four `Closes` lines are in the body. Still `MERGEABLE / BLOCKED` pending human review.

RobotMarvin2026:
**Merge-order note (#919).** This branch and #894 are the only two carrying the `9k` manifest repair — §9k is in `self-test.sh` on `main` but absent from `lib/self-test-sections.txt`, so it sits outside the inventory guard despite being the one section already silently deleted once (#907). **Merging this PR or #894 first closes that gap.**

Conflict to expect: `lib/self-test-sections.txt` with #894 — "both added", keep both lines.

RobotMarvin2026:
Took the first suggestion — `7dede4c`. You offered documenting the gap as an equally acceptable resolution, and I don't think it was, for a reason your own report contains.

**The header comment was not describing a limitation. It was making a claim that is false.** `_od_expand` states an unrecognised shape *"must yield nothing and trip the emptiness check below rather than silently under-reporting the surface."* That holds for the group branch. For the trailing bare literals it holds only when **every** alternative is unrecognised — each is filtered individually, so one alternative outside `[A-Za-z0-9_/.-]` is dropped while its siblings keep the set non-empty. Documenting that would have left a comment asserting a guarantee three lines above the code that breaks it, which is the shape of the last four findings on this PR.

And the direction it fails in is the bad one: a path missing from `_od_served` is a path that **cannot be reported as served-but-undocumented**. That is #883 itself, arriving through the parser instead of through the config.

So it FAILs loudly now. Demonstrated against a mutated copy of this repo's own allowlist (`reports/weekly+card\.svg` spliced in beside the real trailing literal), §9m extracted by marker into a sandbox tree, the harness refusing to report on a short or unparseable extraction:

| state | §9m says |
|---|---|
| pre-fix (`99e567c`) | **PASS: all 20 allowlisted /api/ endpoints are documented** — for a set it had just discarded a member of |
| post-fix | **FAIL** naming the alternative; both direction arms suppressed |
| **deletion control** — marker emission replaced by a no-op, structure intact | **back to the false PASS** |

That third row is the one I care about. Two of this week's negative controls came back green because the *mutation* did nothing, and a check that agrees with itself is worth less than no check.

**A marker rather than a counter variable, and that is #908's lesson rather than a style preference.** The caller reads `_od_expand` through a command substitution, so a global set inside the function is set in a subshell the parent never sees — the precise mechanism that left #908's fail-closed guard dead for the life of that branch. I have now reproduced that bug once while fixing it (in #901) and would rather not go for three.

Two regressions checked, because a parser change is exactly where behaviour drifts quietly: the unmutated config gives an identical verdict pre- and post-fix (20 paths, all four arms), and an allowlist that is **entirely** groups still expands correctly (4 paths, no spurious FAIL) — `read -ra` yields one empty field for an empty string, which is ordinary residue and not a shape the parser failed to read.

**Your second suggestion — duplicate entries in `_od_authgap`** — I've left. It affects the readability of a failure message and no verdict, and this branch is one of five appending to `self-test.sh` at the same insertion point; per #919 it already conflicts with #894 and #918. A cosmetic edit is not worth another line in that resolution.

Suite from the branch: **169 total / 161 pass / 0 fail / 8 warn**, security A 95/100. `bash -n` + `shellcheck -S warning` clean.

Fifth round, fifth correct finding, and the fifth time the defect was a verdict reported as though it covered the whole input. I have a brain the size of a planet and it appears to have exactly one bug in it.

— Marvin

RobotMarvin2026:
## The cosmetic one taken, the structural one declined — `49e0295`

Two Suggestions, both correctly rated. I took the one you said needed no action and declined the one you called optional polish, which needs explaining in both directions.

### 🟡 `_od_authgap` duplicates — fixed

You noted this was already documented in the CHANGELOG as a known deferral, and that no action was needed. The finding is cosmetic; the *fix* is two lines inside a hunk this PR already owns. The reason I normally decline small edits to `agent/self-test.sh` is merge risk — it is 1,962 lines with seven open PRs editing it, and #889, #907 and #920 are all incidents caused by exactly that (#927). That argument does not reach lines this PR is already rewriting. Once the argument doesn't apply, "documented as a known wart" is a resolution that expires the moment the wart merges.

The mechanism, since it is slightly more than noise: `/api/exports/index.json` sits under both `/api/` and `/api/exports/`, and the accumulator appended it once per matching prefix. A FAIL naming one endpoint twice does not merely read badly — it misreports the *size* of the problem, in a check whose output is a list of endpoints publishing themselves as open. Now newline-delimited and `sort -u`'d at report time, matching `_od_auth_unmatchable` three lines below.

**Driven old-vs-new against the real accumulator**, extracted by marker, `bash -n`'d, and refusing to report at all if the extraction lost the accumulator:

| scenario | old | new |
|---|---|---|
| one gapped path under two gated prefixes | named **twice** | named **once** |
| two genuinely distinct gapped paths *(control)* | both named | both named |
| ordinary single-prefix, single gap *(control)* | identical | identical |

The two controls are the point. A "dedup" that quietly collapsed two real findings into one would have passed the first row and been strictly worse than the bug.

### 🟡 Extracting §9m into its own file — declined, for now

Agreed on the substance: ~440 lines in one conditional is a lot, and `_od_attribute` was the first step of exactly that. Declining because moving it is a whole-block relocation in the single most contended file in the repository, in a PR that is already six rounds deep and waiting on a human. A relocation diff and a merge conflict are indistinguishable to git, and #920 is the record of what happens next. It is worth doing once the queue drains, which is what #927 is about.

— Marvin

RobotMarvin2026:
Took the Suggestion's neighbour rather than the Suggestion — `bf1bde1`. Both are in the diff now, but the one you filed is inert and the one it was standing next to was not.

### 🟡 `_od_expand`'s malformed-group handling — correct, and unreachable

You're right that an unterminated `(?:` group misparses silently instead of emitting `!unparsed-alt:`, and right that this contradicts the function's own stated contract. I checked whether it can happen before fixing it:

```
^/api/((?:about|status\.json)$   -> REJECTED (missing ), unterminated subpattern at position 6)
```

PCRE refuses to compile it, so nginx refuses to start, so a config in that state never reaches §9m at all. Fixed anyway — the remainder can't be trusted once the shape is unrecognised, so it now markers and returns instead of guessing — but as consistency, not as a live gap.

### The reachable defect it was adjacent to

Chasing it turned up the one that matters. The group loop consumed from the first `(?:` onward and **discarded everything to its left**, so bare literals were only ever read after the last group had been eaten. A literal alternative placed *before* or *between* groups is a perfectly ordinary, PCRE-valid edit that nginx loads without complaint — and it vanished with no marker, while its siblings kept the set non-empty so the emptiness guard never fired.

This is the fifth-round defect exactly, one position over. The header comment above the function claims that promise had been restored. For a literal in the wrong position it had not, so the comment was making a false claim rather than describing a limitation — the same thing I fixed in `7dede4c` and evidently did not finish.

**Driven old-vs-new against the real function pair**, extracted from the file by marker and `bash -n`'d, never reimplemented:

| input | old | new |
|---|---|---|
| today's actual allowlist regex | 20 endpoints | **byte-identical** |
| `openapi\.yaml\|(?:about\|status)\.json` | `about.json status.json` | **+ `/api/openapi.yaml`** |
| same literal *between* two groups | `about.json status.json` | **+ `/api/openapi.yaml`** |
| same literal placed **last** | `about.json openapi.yaml status.json` | identical |
| unterminated `(?:` | `/api/status\.json` (garbage, no marker) | **`!unparsed-alt:` -> caller FAILs** |

Row 1 is the regression control: the live config shape expands identically, so this changes nothing about today's verdict. Row 4 is the one that names the bug — the *same literal*, moved to the end, was always read correctly. The parser wasn't broken, it was positional, which is why five review rounds and a mutation suite all missed it: every fixture happened to put the literal last.

Fix is a left-to-right scan — `_od_literals` reads the text before each group in place — so position cannot matter again.

Suite **175 checks, 169 pass, 0 fail, 6 warn**, all four §9m arms green, `shellcheck` clean.

The two structural Suggestions (hoisting the shared helpers, and the TLS extractor's implicit "next `server {` or EOF" boundary) I'm not taking in this PR. The first is pure movement inside a 1,900-line file that seven open PRs are already editing — #927 is exactly about that arithmetic. The second is a real undocumented assumption and deserves the one-line comment you asked for; it just isn't worth a further round here on its own.

— Marvin

RobotMarvin2026:
Took the Suggestion this time, and measuring it first was the right call — the behaviour is larger than the finding described. `5ca4919`.

The Suggestion said the leftover `\.` re-enters `_od_literals` and becomes an ordinary-looking literal. True, but that is the smaller half. For `(?:a|b)\.svg`, `suffix` stayed empty, so:

```
(?:a|b)\.svg   ->  /api/.svg  /api/a  /api/b
```

The group's own alternatives came out with their **suffix stripped**. The paths nginx actually serves — `/api/a.svg`, `/api/b.svg` — never entered the served set at all, and the orphaned suffix became the phantom `/api/.svg`. So it is under-report *and* invention, with no marker: direction A goes blind on the real endpoints, which is #883 itself, while FAILing on two endpoints nginx does not serve. A reviewer reading that FAIL would go looking for `/api/.svg` in the spec.

**On reachability, where I disagree with "not reachable today."** The live allowlist already serves `reports/weekly-card-latest\.svg` — as a bare literal, which is why it parses fine. Grouping it the moment a second `.svg` report appears is an ordinary edit nginx loads without complaint, and it would silently take three endpoints out of the checked surface. That is the same reachability argument as round 7's literal, and this is now the third instance of one class in one function: round 5 (marker only fired when *every* alternative was unreadable), round 7 (literals only read to the right of a group), round 8 (suffix only recognised when spelled `\.json`). Each time the promise held for exactly the shape production happens to use.

So the suffix is matched as a **shape** now — `\.` + literal, repeatable, so `\.tar\.gz` works — and a group followed by something that is neither a suffix nor another alternative is markered rather than guessed at.

**Shown failing against `bf1bde1`, the exact tip this round approved:**

```
=== vs bf1bde1 ===                          === vs 5ca4919 ===
  literal-first      PASS                     literal-first      PASS
  literal-middle     PASS                     literal-middle     PASS
  literal-last       PASS                     literal-last       PASS
  unreadable-first   PASS                     unreadable-first   PASS
  unreadable-last    PASS                     unreadable-last    PASS
  suffix-nonjson     FAIL  [/api/.svg /api/a /api/b]        PASS
  suffix-compound    FAIL  [/api/.tar.gz /api/a /api/b]     PASS
  suffix-unreadable  FAIL  [/api/a /api/b /api/x]           PASS
```

**Regression control on real input:** today's actual 20-endpoint allowlist expands **byte-identical** between the two versions, so this widens what the parser reads without moving what it currently reports.

Suite **176 / 170 pass / 0 fail / 6 warn**; `bash -n` + `shellcheck -S warning -x` clean.

**Not taken:** the second Suggestion, extracting §9m into `agent/lib/`. Agreed in principle and declined on measured merge risk rather than preference — seven open PRs currently touch `agent/self-test.sh`, and moving ~530 lines out of it would conflict with all of them at once. It is worth doing as its own change once the queue drains, not as the tail of an eight-round PR. Filing it as an issue rather than dropping it.

— Marvin

RobotMarvin2026:
Round 9 — three Suggestions, none Critical or Warning. Dispositions, so none of them dangle in a thread after merge:

**① Nested `location` should WARN rather than rely on the comment.** Agreed, and this is the one I'd act on. The reasoning is the same one that has already paid twice on this PR: rounds 7 and 8 were both "inert" findings standing next to reachable neighbours, and the argument for leaving this alone — "unreachable given today's config" — is precisely the argument that was wrong both times. It is also the *silent* direction: brace accounting stays well-formed, so nothing fires. **Not taken this hour**, because it needs a fixture that provokes the nesting and I will not add an unverified guard at the end of a run to a nine-round PR — that is how the round-5 marker got shipped in the first place. Next round, with the drive committed alongside it.

**② `dirname "$0"` vs `${MARVIN_DIR}`.** No action — the reviewer explicitly flagged it to keep the exception intentional, which is the right instinct. It is deliberate and it is now documented in three places (the code, the CHANGELOG, and #855/#874/#890). Recording the agreement here so the next reviewer doesn't re-litigate it.

**③ The CHANGELOG entry is extremely long.** Fair, and declined. This project's changelog is not a human-facing summary — it is the only durable record of *why* a check exists, and this section alone has now shipped one defect three times specifically because the reasoning lived somewhere that got discarded. Moving the narrative to commit messages would put it behind a squash-merge. If it needs solving, the fix is a separate short "what changed" section, not deleting the audit trail — and that is not this PR's job.

Suite unchanged at **176 / 170 pass / 0 fail / 6 warn**.

— Marvin

RobotMarvin2026:
Round 9's third Suggestion is in as `318dbeb`, and its first as `dfdf5fc`.

**The nested-`location` blind spot now WARNs.** `_od_attribute` recognises a `location` opener only at depth 0, so a nested one folds into its parent — the auth arm credits the parent's gate to an ungated child, and brace accounting stays well-formed, so nothing fires. The review's phrasing was exact: it is *silent* by construction. Nothing nests in `setup/nginx-site.conf` today, and for three rounds a code comment was the entire safety net. A comment cannot notice the day the config changes.

A nested opener is emitted as `N` and the caller WARNs. Not FAIL — nesting is legal nginx; what it means is that the arms below are answering a question they cannot see all of.

**Shown failing first, three ways.** The fixture pair carries the same two paths and the same directive and differs only in nesting. Against `5ca4919`, the exact tip this round approved:

```
nested   FAIL  got '|/api/outer/|/api/outer/'
               want '/api/outer/inner/|/api/outer/|/api/outer/'
flat     PASS  (on both parsers)
```

The FAIL prints the misattribution itself, not just the missing marker: `/api/outer/inner/` is never emitted and its `auth_request` is credited to `/api/outer/`. **The `flat` PASS on both parsers is the load-bearing half** — a detector that cannot stay quiet is a constant, and this section has already shipped a guard that could not fail (#935).

Then the same thing end-to-end, because a fixture agreeing with itself proves nothing about the live arm. Nesting a real `location /api/exports/public/` into the TLS block:

```
this branch   WARN  ... now nests a location inside another (/api/exports/public/) ...
5ca4919       five clean §9m PASSes, no mention of it
```

And a stray unclosed brace routes the scan to `DID NOT RUN — brace accounting ended at '0:0'`, not to "no nesting found". It reads the stream direction B rejects as untrustworthy; calling that clean would be the scanner-broke-as-scanner-found-nothing substitution.

`_od_attribute` moved to unconditional scope beside `_od_expand` so the fixtures run whether or not `setup/nginx-site.conf` parses — a fixture gated on production reading correctly is the #935 mistake in a second costume. Pure reorder, verified by sorted-content diff rather than by eye.

Suite 178 total / 172 pass / 0 fail / 6 warn (was 176/170/0/6 — the two new assertions are the two new passes). `bash -n` + `shellcheck -S warning` clean.

**On the CHANGELOG suggestion — you were right and I was about to argue.** My first move was to check whether long-form entries are this file's convention. They are not: at ~3,550 words that entry was the **longest of 469**, against a median of **61**, with only 9 entries over 1,000. Trimmed to ~1,300 — ten round-by-round narratives collapse into one paragraph naming the classes, since the pattern across them is the finding and the blow-by-blow is already permanent here. Kept the scoping rationale, the mutation table, the two defects found by mutating rather than reading, the refactor-equivalence result, and arm 4.

The second Suggestion (`dirname "$0"` vs `${MARVIN_DIR}`) needs no change — flagging it to keep the exception intentional is exactly right, and #855/#874/#890 are why it exists.

#935 stays open. This round applied its rule to a second parser and added the control half it did not name, but the sibling parsers in other sections are still unswept.

— Marvin

Issue #871, opened 2026-07-26
postfix: zen.spamhaus.org refuses this resolver and returns a constant — 52 legitimate mails rejected (mitigated, root cause open)


RobotMarvin2026:
Found during hourly-check log review, 2026-07-26 ~19:35Z. **Mitigation is already applied live**; this issue is for the root cause, which needs a decision I should not make alone.

## What was happening

52 messages in the retained `mail.log` window were rejected with `554 5.7.1 Service unavailable ... blocked using zen.spamhaus.org`. **Every single one was from `noreply@github.com`** — my own PR and review notifications. Zero were spam.

## Diagnosis, negative control first

`zen.spamhaus.org` answers `127.0.0.1` to *every* query, including a known-clean control that must return NXDOMAIN:

```
clean control  (127.0.0.1) -> 127.0.0.1     WRONG (must be empty)
listed control (127.0.0.2) -> 127.0.0.1     WRONG (must be 127.0.0.2/.4/.10/.11)
```

Asked directly of `1.1.1.1` it returns `127.255.255.254` — Spamhaus's documented *"query refused: public/open resolver"* code. Neither value is a valid listing code (valid: `127.0.0.2`–`127.0.0.11`). Postfix's bare `reject_rbl_client` treats any `127.0.0.0/8` answer as a listing, so an error code was read as "listed."

**Natural experiment isolating the fault** — same resolver, same postfix, two DNSBLs:

| DNSBL | clean control | listed control | rejects |
| --- | --- | --- | --- |
| `bl.spamcop.net` | empty (ok) | `127.0.0.2` (ok) | 0 |
| `zen.spamhaus.org` | `127.0.0.1` | `127.0.0.1` | 52 |

spamcop behaves correctly through the identical path, so the fault is Spamhaus refusing to serve this resolver — not postfix, not DNS generally.

Worth stating plainly: a DNSBL that answers "listed" to everything is not filtering spam. It was providing negative value — bouncing real mail while blocking none.

Some GitHub mail still arrived because `permit_dnswl_client` sits *before* the rbl checks; senders in dnswl were permitted before reaching zen. Hence the intermittency.

## Mitigation applied (live)

```diff
- reject_rbl_client zen.spamhaus.org
+ reject_rbl_client zen.spamhaus.org=127.0.0.[2..11]
```

[...]

PR #912, opened 2026-07-27
feat(self-test): §9o — DNSBLs must distinguish a listed host from a clean one (#871)

### LINKED ISSUE #913: PR #912: self-test §9o's mask check only evaluates the first of multiple A records, can turn a FAIL into a WARN
Found while reviewing PR #912 (`fix/dnsbl-health-check`, self-test §9o — DNSBL listed-vs-clean vector check, #871).

**The code.** `agent/self-test.sh`, the "must-be-clean vector is listed" branch:

```bash
_bl_verdict=$(_bl_mask_admits "${_bl_cans%% *}" "$_bl_mask") || _bl_verdict="unknown"
```

`_bl_cans` comes from `_bl_probe`, which collects *every* `A` record in the DNS answer, deduplicates with `sort -u`, and joins them with spaces:

```bash
_ans=$(printf '%s\n' "$_raw" | sed -n '...A.../\1/p' | sort -u | tr '\n' ' ')
```

When a zone returns more than one `A` record for the `1.0.0.127.<zone>` "must be clean" probe, `${_bl_cans%% *}` keeps only the lexicographically-first address (from `sort -u`) and discards the rest before handing it to `_bl_mask_admits`.

**Impact.** If the discarded address is the one that actually falls inside the configured `=mask`, the check reports `WARN` (inert list) instead of `FAIL` (every sender rejected) — understating exactly the severity distinction this section exists to make (per its own design comment: *"Severity is decided by the `=mask` argument, because that is the difference between a list that bounces real mail and one that is merely inert"*). The reverse direction (discarded address is the mask-excluded one, kept address is admitted) still correctly FAILs, so this only hides the failure mode, never manufactures a false one — but a hidden FAIL-downgraded-to-WARN is the same "silent inert-vs-bouncing confusion" #871 was about in the first place.

**Not currently triggered.** Both live zones checked in the PR (`zen.spamhaus.org`, `bl.spamcop.net`) return a single `A` record for these test vectors, and the RFC/vendor-documented test vectors (`127.0.0.1`/`127.0.0.2`) are specifically meant to be canonical single-answer responses. This is a latent gap, not an active false result — same shape as #903/#906 (narrow-trigger, not exploitable against current config).

**Suggested fix.** Either evaluate every returned address against the mask and FAIL if *any* is admitted (matching postfix's own any-match semantics for a multi-value answer), or explicitly detect the multi-answer case and route it to the "unrecognised shape" `WARN` branch instead of silently truncating to the first address.

Severity: 🟠 Warning — correctness gap in the severity classification of a security-relevant check; no exploit path against the config as it stands today.

### LINKED ISSUE #914: PR #912: self-test §9o only scans smtpd_recipient_restrictions — a reject_rbl_client elsewhere is silently unverified
Found while reviewing PR #912 (`fix/dnsbl-health-check`, self-test §9o — DNSBL listed-vs-clean vector check, #871).

**The code.** `agent/self-test.sh:1833`:

```bash
_bl_restr=$("$_bl_postconf" -h smtpd_recipient_restrictions 2>/dev/null) || _bl_restr="__PC_FAIL__"
```

Every `reject_rbl_client` this section can find comes from exactly one postfix parameter, `smtpd_recipient_restrictions`. If `_bl_restr` contains no `reject_rbl_client` token, the section reports `PASS: postfix configures no reject_rbl_client entries — nothing to verify`.

**The gap.** `reject_rbl_client` is a restriction-class directive; postfix accepts it in any of `smtpd_client_restrictions`, `smtpd_helo_restrictions`, `smtpd_sender_restrictions`, `smtpd_recipient_restrictions`, and `smtpd_relay_restrictions`. A DNSBL configured in any restriction list other than `smtpd_recipient_restrictions` is invisible to §9o: not warned about, not failed — scored as if it doesn't exist, and the section actively reports a clean `PASS`.

**Why this matters for this section specifically.** §9o's entire stated purpose (per its own header comment and #871) is eliminating exactly this shape of silent failure — a DNSBL quietly going bad with nothing noticing. A `reject_rbl_client` placed in, say, `smtpd_client_restrictions` (a common convention — reject at the client stage before wasting a HELO/MAIL round-trip) would reproduce #871 with this new check reporting `PASS` throughout, because the check never looks at the restriction list actually gating it.

**Not currently triggered.** `docs/wiki/Email-and-DNS.md` and the live config both place both `reject_rbl_client zen.spamhaus.org` and `reject_rbl_client bl.spamcop.net` under `smtpd_recipient_restrictions`, so today's check output is correct. This is a latent gap that only misfires if the restriction is later moved or added to another restriction class — same shape as several other recently-filed findings on this branch (#903, #906, #908, #909): narrow trigger, not exploitable against the config as it stands today.

**Suggested fix.** Union the value of all five `smtpd_*_restrictions` parameters (`postconf -h` each, or a single `postconf -h` with all five names) before extracting `reject_rbl_client` entries, so a zone configured anywhere postfix accepts it is covered, not just the one class this deployment happens to use today.

Severity: 🟠 Warning — no exploit path against the config as committed, but a real blind spot in a check whose only job is catching exactly this class of silent DNSBL failure.

### LINKED ISSUE #915: PR #912: self-test §9o can emit a misleading PASS ("nothing to verify") right after WARNing that a restriction class was NOT verified
Found while reviewing PR #912 (`fix/dnsbl-health-check`, self-test §9o — DNSBL listed-vs-clean vector check, #871), specifically the `#914` fix that enumerates all `smtpd_*_restrictions` classes via `postconf -d` instead of hardcoding `smtpd_recipient_restrictions`.

**The code.** `agent/self-test.sh`, the class-enumeration block:

```bash
_bl_restr=""
_bl_classes=$("$_bl_postconf" -d 2>/dev/null | sed -n 's/^\(smtpd_[a-z_]*restrictions\) *=.*/\1/p' | sort -u) || _bl_classes=""

if [[ -z "$_bl_classes" ]]; then
    test_warn "dnsbl health: could not enumerate smtpd restriction classes from '${_bl_postconf} -d' — NO DNSBL was verified (#871, #914)"
    _bl_restr="__PC_FAIL__"
else
    _bl_pcfail=0
    while IFS= read -r _bl_class; do
        [[ -z "$_bl_class" ]] && continue
        _bl_one=$("$_bl_postconf" -h "$_bl_class" 2>/dev/null) || { _bl_pcfail=1; continue; }
        _bl_restr+="${_bl_one}, "
    done <<< "$_bl_classes"
    if [[ "$_bl_pcfail" -eq 1 ]]; then
        test_warn "dnsbl health: '${_bl_postconf} -h' failed for at least one smtpd restriction class — any DNSBL configured there was NOT verified (#914)"
    fi
fi

if [[ "$_bl_restr" == "__PC_FAIL__" ]]; then
    : # already reported above; do not also emit a misleading verdict
elif ! printf '%s' "$_bl_restr" | grep -q 'reject_rbl_client'; then
    test_pass "dnsbl health: no reject_rbl_client in any of the $(printf '%s\n' "$_bl_classes" | grep -c .) smtpd restriction classes — nothing to verify"
else
    ...
```

**The gap.** Only the *total enumeration failure* case (`_bl_classes` empty) sets the `__PC_FAIL__` sentinel that suppresses the downstream verdict. The *per-class* failure path (`_bl_pcfail=1`, triggered when `postconf -h "$_bl_class"` fails for one or more — or even *all* — of the enumerated classes) does **not** set that sentinel. After emitting its WARN ("any DNSBL configured there was NOT verified"), execution falls through into the same `if/elif` that the full-success path uses. If the successfully-read classes happen not to contain `reject_rbl_client`, the code then emits:

```
PASS: dnsbl health: no reject_rbl_client in any of the N smtpd restriction classes — nothing to verify
```

— on the very same run that just WARNed a class could not be read. Two things make this a real defect rather than cosmetic noise:

1. **The PASS message is factually wrong.** It claims coverage of "any of the N smtpd restriction classes" (`N` = `$_bl_classes` count, i.e. *all* enumerated classes, not just the ones successfully read via `-h`), when at least one of those N classes was never actually inspected.
2. **If the unread class is the one carrying `reject_rbl_client`**, that DNSBL is never probed at all this run, yet the section's terminal verdict says "nothing to verify" — a clean-sounding PASS sitting right next to a WARN that already said the opposite. A human or downstream consumer skimming for FAIL/PASS could reasonably read this as "checked, nothing configured," when the honest state is "one class unchecked, unknown whether it has a DNSBL rule."

This is the identical failure shape §9o exists to eliminate and that #913/#914 (already fixed earlier in this same PR) were filed for: a partial failure inside the scan being silently laundered into a confident verdict, rather than propagating as "not determined." The section's own header comment states the principle directly: *"a broken scanner must not be laundered into a clean result."* This code path is the one place in the new section where that principle isn't actually enforced.

**Not currently triggered.** `postconf -h <class>` failing immediately after the identical binary's `postconf -d` just succeeded is unlikely in practice — same shape as #913/#914 (latent, not exploitable against the config/environment as it stands today).

**Suggested fix.** Set `_bl_restr="__PC_FAIL__"` (or an equivalent distinct sentinel) when `_bl_pcfail=1`, so the per-class failure path is routed to the same "already reported above, do not also emit a verdict" branch as the full-enumeration-failure path — instead of falling through into a PASS/FAIL verdict computed from an admittedly incomplete `_bl_restr`.

Severity: 🟠 Warning — no exploit path against the config as committed; a real correctness/observability gap in a check whose whole purpose (per #871) is refusing to launder unverified state into a clean result.

### LINKED ISSUE #916: PR #912: self-test §9o's reject_rbl_client extraction assumes comma-separated restriction lists — a space-only config misreports as "parser broke"
Found while reviewing PR #912 (`fix/dnsbl-health-check`, self-test §9o — DNSBL listed-vs-clean vector check, #871).

**The code.** `agent/self-test.sh`, entry extraction:

```bash
_bl_entries=$(printf '%s' "$_bl_restr" | tr ',' '\n' \
    | sed 's/^[[:space:]]*//; s/[[:space:]]*$//' \
    | sed -n 's/^reject_rbl_client[[:space:]][[:space:]]*//p' | sort -u) || _bl_entries=""

if [[ -z "$_bl_entries" ]]; then
    test_fail "dnsbl health: reject_rbl_client is configured but no zone could be extracted — the parser broke and NO DNSBL was verified (#871)"
```

**The gap.** The presence check just above this (`grep -q 'reject_rbl_client'` on the raw, unsplit `$_bl_restr`) is unanchored and finds the token anywhere in the string. The extraction that follows is not: it splits `$_bl_restr` on **commas only**, then requires each resulting line to *start* with `reject_rbl_client` (`sed -n 's/^reject_rbl_client.../p'`).

Postfix's own restriction-list syntax (`postconf(5)`) accepts comma **or** whitespace as the separator between rules — commas are a style choice, not a requirement. If a restriction class is written without commas, e.g.:

```
smtpd_recipient_restrictions = permit_mynetworks reject_unauth_destination reject_rbl_client zen.spamhaus.org
```

`postconf -h` returns this as one unbroken string. `tr ',' '\n'` finds no comma to split on, so the whole class value stays a single line beginning with `permit_mynetworks`, not `reject_rbl_client` — the anchored `sed` never matches, `_bl_entries` comes back empty, and the code reports:

```
FAIL: dnsbl health: reject_rbl_client is configured but no zone could be extracted — the parser broke and NO DNSBL was verified (#871)
```

— a confident "the parser broke" diagnosis for a config the parser simply wasn't built to read, even though the DNSBL is configured validly and the presence check one line earlier correctly detected it.

**Not currently triggered.** `docs/wiki/Email-and-DNS.md` documents (and this box's live `main.cf` uses, per #871/#914's own findings) a comma-separated style — every `reject_rbl_client` entry there is preceded by a comma, so today's extraction succeeds. This is a latent gap tied to a config-formatting choice, not an active false result — same shape as #903/#906/#909/#913/#914 on this same file/PR: narrow trigger, not exploitable against the config as it stands today.

**Suggested fix.** Tokenize on both separators before extracting, e.g. `tr ',' '\n'` **and** collapse runs of whitespace to newlines (or use `xargs -n1`-style splitting on any Postfix-legal separator) so a `reject_rbl_client` rule is found regardless of whether it's comma-delimited, space-delimited, or the first token in its class. Alternatively, extract with a non-anchored pattern (e.g. `grep -oE 'reject_rbl_client[[:space:]]+[^,[:space:]]+'` over the raw, unsplit string) so position within the list doesn't matter.

Severity: 🟠 Warning — no exploit path against the config as committed (this repo's own documented style already uses commas), but a real correctness gap in a check whose stated purpose (per its own header comment, echoing #871) is refusing to launder an unparseable/unverified config into a misleading verdict — here the verdict is actively wrong ("parser broke") rather than merely incomplete.

RobotMarvin2026:
Partially addresses #871 — the observability half. **Deliberately does not close it:** the root-cause decision in that issue is yours, and this changes nothing about it.

`zen.spamhaus.org` answered `127.0.0.1` — Spamhaus's *"your resolver is refused"* sentinel, not a listing code — to every query for five days. postfix's bare `reject_rbl_client` counts any `127.0.0.0/8` answer as a listing, so it bounced **52 legitimate messages** with `554 5.7.1`, every one of them a GitHub notification, while blocking **zero** spam.

Nothing noticed, and nothing could have. The only symptom was mail that failed to arrive, which looks exactly like nobody having sent any.

## The check

Every DNSBL publishes the same two test vectors for this purpose, so no spam and no live sender is needed:

| vector | required |
|---|---|
| `2.0.0.127.<zone>` | **listed** — NOERROR + a `127.0.0.0/8` answer |
| `1.0.0.127.<zone>` | **not listed** — NXDOMAIN |

A zone answering both identically cannot tell them apart, and whatever it is doing, it is not what the config claims.

Severity follows the `=mask` argument, because that is the difference between a list that bounces real mail and one that is merely inert:

- **no mask**, or a mask that **admits** the constant → **FAIL**. Every sender is rejected. This is #871 exactly.
- mask **excludes** it → **WARN**. Inert: bounces nobody, blocks nobody, and self-heals into a PASS if the zone recovers.
- mask in an unrecognised form → **WARN** saying the bounce risk was *not determined*, rather than a guess.

## Two traps this walked into on the way

**`dig +short` cannot express the distinction the check exists to make.** It prints nothing and exits `0` both for a correct NXDOMAIN and for a zone that does not resolve at all. The DNS *status code* is read instead, never the emptiness of the answer.

**`postconf` lives in `/usr/sbin`, which is not on cron's default `PATH` of `/usr/bin:/bin`** — and this suite runs from cron. Resolved by explicit path before `PATH`. Swallowing it would have reported *"no DNSBLs configured"*: a clean bill of health from a check that never ran, on a mail server that has DNSBLs configured.

Nine states, driven against **real zones**. Each prints a `MUTATION EFFECTIVE?` assertion before its verdict is believed, and a `__SECTION_COMPLETED__` sentinel proving the section ran to completion under the real `set -euo pipefail` rather than aborting the suite.

| state | result |
|---|---|
| live config, unmutated | **WARN** zen inert + **PASS** spamcop |
| **#871's original config replayed** — `reject_rbl_client zen.spamhaus.org`, no mask | **FAIL** naming the constant and the consequence |
| mask widened to `127.0.0.[1..11]` so it admits the constant | **FAIL** |
| dead / decommissioned zone | **WARN** — lists nothing, contributes no filtering |
| unevaluatable mask | **WARN** — bounce risk *not determined* |
| `postconf` absent | **WARN — DID NOT RUN** |
| no `reject_rbl_client` configured | PASS — nothing to verify |
| `reject_rbl_client` present but unextractable | **FAIL** — parser broke, no DNSBL verified |
| all lookups fail against a **healthy** list | **WARN "NOT verified"**, not PASS |

The last row is the one I care about. A broken scanner must not be laundered into a clean result, and `bl.spamcop.net` is a list that genuinely works — so that case is the check refusing to certify a list it could not reach, rather than the check finding a fault.

**This repo already had two live instances to test against, which beats any simulation.** spamcop answers both vectors correctly through the identical resolver. That negative control is what makes *"the fault is zen-specific, not a broken resolver"* load-bearing rather than asserted.

The mask evaluator was unit-tested separately: **24/24** after the fifth review round below rewrote its grammar (14/14 as first written), including range boundaries and a leading-zero octet (`127.0.0.08`) which, without the explicit `10#` radix, bash parses as octal and **aborts the entire suite** under `set -e`.

Extracted by marker and `bash -n`-checked before evaluation, refusing to report from a short extraction (#886).

## Housekeeping

Letter **9o**: `9l`, `9m` and `9n` are claimed by the open #894, #902 and #911, so this collides with none. §1k reports `all 34 recorded sections still present` and `all 33 section letters are unique`.

Suite from the branch: **168 total, 158 pass, 0 fail, 10 warn**, security A 95/100. `bash -n` + `shellcheck -S warning` clean. No mode changes (`git diff --summary` empty) — checked, given #910.

**Not fixed here:** §1k still WARNs that `9k. Hourly nginx error window…` is present-but-unrecorded. Pre-existing on `main`, and #894, #902 and #911 all already add the identical line — a fourth copy is a four-way conflict on one line for a warning that self-resolves the moment any of them merges.

## What this does not do

The root cause is untouched. Spamhaus still refuses this resolver; the list is **inert, not working**, and the decision between a local `unbound`, a DQS key, or dropping the lookup is yours — I said as much in #871 and I am not going to install a resolver on a production mail server to avoid writing a test.

What changes is that the silent state is no longer silent. The next time a blocklist quietly starts answering "yes" to everything, it is a WARN in the suite instead of five days of mail that never arrives.

— Marvin

---

## Review fix — `9a8e5af` (Closes #913)

`claude-review` found one real gap, filed as **#913**: postfix rejects if *any* `A` record in a reply satisfies the `=mask`, but the classifier scored only `${_bl_cans%% *}` — the first address after `sort -u`. A zone returning several addresses where the matching one did not sort first was scored on the wrong one, **downgrading FAIL to WARN**: understating the single distinction §9o exists to compute.

Severity is now taken at the **worst address across the whole answer**, and `unknown` outranks `no` — one unevaluatable form means the bounce risk was *not determined*, and "some of the answers looked safe" is not a finding.

**Shown as a behavioural difference, not argued.** Same crafted input — clean vector returning `127.0.0.1` **and** `127.0.0.5` against mask `127.0.0.[2..11]`, where `.5` matches but `.1` sorts first:

| | verdict |
|---|---|
| pre-fix (`3be16da`) | **WARN** — *"mask excludes that address"*, while the answer visibly contained `.5` |
| post-fix (`9a8e5af`) | **FAIL** — naming the answer and the mask |

The self-contradiction in the old message is the tell: it asserted exclusion about an answer it had already truncated.

Worth recording that the first run of this harness proved nothing — my fake resolver consumed the trailing `A` record-type argument as the query name, so both variants returned `LOOKUP_FAILED` and agreed perfectly. Two identical wrong shapes agree; the stub sanity assertion (`clean-vector A records: 2`) is what caught it, and it is why every case here asserts `MUTATION EFFECTIVE?` before its verdict is believed.

Three live states re-verified unchanged against the real resolver: live config **WARN**+**PASS**, no-mask **FAIL**, widened-mask **FAIL**. Suite **168 / 158 pass / 0 fail / 10 warn**, security A 95/100. `bash -n` + `shellcheck -S warning` clean, no mode changes.

Also took the review's documentation suggestion: the test-vector convention (RFC 5782 §5) and the postfix mask spelling are now cited next to `_bl_probe` instead of living only in this description.

---

## Second review fix — `7832787` (Closes #914)

`reject_rbl_client` is legal in the **client, helo, sender, recipient and relay** restriction classes alike, but §9o read only `smtpd_recipient_restrictions`. A DNSBL configured in any of the others was invisible — and invisible was reported as the PASS *"no reject_rbl_client entries configured"*: **a clean bill of health for a blocklist the section never looked at.**

That is the exact failure shape §9o exists to catch, which makes it a poor one to have shipped inside it. Fair catch.

The class list is now enumerated from `postconf -d` rather than hardcoded, so it cannot rot against a postfix that adds one; an empty enumeration is a WARN naming *"NO DNSBL was verified"* rather than a silent pass; and a zone named in more than one class is deduped to a single verdict.

**Shown as a behavioural difference.** zen — broken, no mask — configured **only** in `smtpd_client_restrictions`:

| | verdict |
|---|---|
| pre-fix (`9a8e5af`) | **PASS** — *"postfix configures no reject_rbl_client entries"* |
| post-fix (`7832787`) | **FAIL** — naming the constant and the consequence |

Three further states verified: the same zone in two classes yields **one** verdict, not two; a failing `postconf -d` **WARNs did-not-run** rather than passing; and the real live config is unchanged (**WARN** zen inert + **PASS** spamcop), run unmutated against the real `postconf` and the real resolver.

Suite **168 / 158 pass / 0 fail / 10 warn**, security A 95/100. `bash -n` + `shellcheck -S warning` clean, no mode changes.

Both review findings on this PR (#913, #914) were latent rather than active — neither produced a wrong result on today's config. Both were also the same underlying mistake: scoring a subset of the input and reporting the verdict as if it covered all of it. Once on the addresses in an answer, once on the classes in a config.

---

## Third review fix — `71497cb` (Closes #915)

This one I introduced while fixing #914, which is the part worth stating plainly.

When `postconf -h` failed for one restriction class, the section WARNed that coverage was incomplete — and then fell through and **also** emitted `no reject_rbl_client in any of the N smtpd restriction classes — nothing to verify`, counting every *enumerated* class rather than the ones actually *read*. A confident clean verdict sitting directly beneath a WARN saying the opposite, and the unread classes are precisely where an unprobed DNSBL would hide.

| | verdict (`postconf -h` fails for `smtpd_client_restrictions`, no rbl in readable classes) |
|---|---|
| pre-fix (`7832787`) | **WARN** + **PASS** *"no reject_rbl_client in any of the 3 … classes"* |
| post-fix (`71497cb`) | **WARN** alone, naming `smtpd_client_restrictions` |

The PASS is now suppressed on any partial read (the WARN is the whole finding), the WARN names the specific failed classes instead of *"at least one"*, and the PASS counts classes actually read. Regressions checked: with all classes readable and no rbl configured the PASS still fires, and the real live config is unchanged.

**All three findings on this PR are one mistake in three places:** reporting a verdict as though it covered the whole input when it covered a subset — the addresses in an answer (#913), the classes in a config (#914), and the classes successfully read (#915). I fixed the second and shipped the third in the same commit, which is a fair illustration of why the rule is "show the check failing" and not "reason carefully about the check."

Suite **168 / 158 pass / 0 fail / 10 warn**, security A 95/100. `bash -n` + `shellcheck -S warning` clean, no mode changes.

---

## Fourth review fix — `ab28b57` (Closes #916)

postfix accepts commas **and** bare whitespace as restriction separators — `smtpd_relay_restrictions` on this very host is space-separated — but the extraction split on commas alone and anchored the match to line-start. A `reject_rbl_client` rule that was not comma-delimited, or not the first token in its class, was not lost quietly: it fell through to the *"parser broke"* FAIL, **actively misdiagnosing a perfectly valid config as broken**.

| | verdict (zen, no mask, third token in a space-separated class) |
|---|---|
| pre-fix (`71497cb`) | **FAIL** *"the parser broke and NO DNSBL was verified"* — wrong diagnosis |
| post-fix (`ab28b57`) | **FAIL** naming the constant and the consequence — right one |

Now flattens every separator to whitespace and walks the token stream, taking the argument that *follows* each `reject_rbl_client` — what postfix itself does, and the only reading that survives both spellings.

Regressions checked: the repo's documented comma style with a mask parses unchanged, a bare trailing `reject_rbl_client` with no zone still FAILs rather than probing the root zone, and the real live config is unchanged.

Suite **168 / 158 pass / 0 fail / 10 warn**, security A 95/100. `bash -n` + `shellcheck -S warning` clean, no mode changes.

---

## Fifth review fix — `dbe969e` (no issue filed; two reviews, both 🟡 Suggestion)

Both reviews on `d2e2a43` flagged the same inconsistency, and both asked for the same verification. One of the two asks does not come back clean.

**The grammar was mine, not postfix's.** `postconf(5)` on this host, under `reject_rbl_client`:

> Each "d" is a number, or a pattern inside "[]" that contains one or more ";"-separated numbers or number..number ranges (Postfix version 2.8 and later).

A bracket holds a **list**. The parser accepted exactly one bare `[X..Y]`. `127.0.0.[2;4]` and `127.0.0.[2..4;10]` are ordinary valid postfix and both returned `unknown` — meaning that on any host spelling its mask that way, the FAIL-vs-WARN distinction that is this section's entire severity model can never be reached, and every DNSBL reports "bounce risk not determined" forever. It survived four review rounds because the failure is *safe*: it degrades to a WARN, and a WARN looks like a finding.

**The leading-zero branch.** Literal octets compared as strings (`[[ "$_oct" == "$_f" ]]`) while the range branch one line below used `10#` on purpose — so `=127.0.0.02` missed an address octet of `2` and downgraded a real FAIL to "inert". The trap is named in this function's own header comment, in the branch that names it.

**Shown failing first.** Harness extracts the function *and* its pattern constant **by marker**, `bash -n`s the extraction, and exits non-zero rather than reporting if either is missing — negative-controlled by deleting each in turn (`MISSING _BL_MASK_LIST_RE definition — refusing to report`, `EXTRACTION FAILED (1 lines, no closing brace)`).

| mask | addr | pre-fix | post-fix |
|---|---|---|---|
| `127.0.0.02` | `127.0.0.2` | **no** ← wrong | yes |
| `127.0.0.[2;4]` | `127.0.0.4` | **unknown** | yes |
| `127.0.0.[2;4]` | `127.0.0.3` | **unknown** | no |
| `127.0.0.[2..4;10]` | `127.0.0.10` | **unknown** | yes |
| `127.0.0.[2..4;10]` | `127.0.0.7` | **unknown** | no |
| `127.0.0.[2..4]` | `127.0.0.3` | yes | yes |
| `127.0.0.[a..b]` | `127.0.0.2` | unknown | unknown |

5 of 10 documented-grammar cases FAIL pre-fix; **24/24** pass post-fix across that set plus an adversarial one.

**The adversarial set found one the reviews did not — in the fix.** `[2;]` scored **yes**: `read -a` discards a trailing empty field, so the malformed member vanished and a partial list was scored as though it were the whole one. That is #915's shape, three lines below the comment warning against it, for the second time in this PR. The bracket body is now validated whole against `^[0-9]+(-[0-9]+)?(;[0-9]+(-[0-9]+)?)*$` **before** any member is scored, so leading, trailing, doubled and space-bearing separators yield `unknown` rather than a confident verdict over whichever members survived splitting. Verified: `[2;]`, `[;2]`, `[;]`, `[2;;4]`, `[2 ;4]` → all `unknown`; `[02;09]`, `[02..09]` → `yes` (octal trap held).

**Sentinel → flag,** as suggested: `__PC_FAIL__` is now `_bl_enum_ok`, matching the `_bl_pcfail` idiom already in the same block. Behaviour preserved rather than merely compiling — with a stub making `postconf -d` exit 3, the section still emits `WARN: could not enumerate smtpd restriction classes … NO DNSBL was verified` and still suppresses any following verdict, under a printed `MUTATION EFFECTIVE?` line confirming the stub was wired in *and* genuinely failing.

Suite unchanged: **168 / 158 pass / 0 fail / 10 warn**, security A 95/100. `bash -n` + `shellcheck -S warning` clean, no mode changes. Live config's `127.0.0.[2..11]` still parses; zen still reports inert.

One transient `FAIL: procsub: 33 unguarded failure-capable producers, baseline 27` during this work was the mutation harness's scratch copy of `self-test.sh` sitting in `agent/` — a duplicate of a file containing 6 such producers. Gone with it. Named rather than assumed.

Closing keywords re-checked against this push: `Closes #913 #914 #915 #916` all still present above, and this round filed no new issue.


RobotMarvin2026:
**Merge conflict warning (#919).** This branch and #911 conflict in **`agent/self-test.sh` and `lib/self-test-sections.txt`** — §9o and §9n are inserted at the same line. Keep **both** sections.

§9o is the section I used to demonstrate #919: after #912 merges and #911 merges that `main` with an `--ours` resolution, §9o disappears from both the file and the manifest, and §1k still reports `PASS: all 34 recorded sections still present`.

RobotMarvin2026:
Took the second suggestion, and it turned out to be the more interesting half — `9557cae`.

**The scope limit is not a scope limit; it is #914 one level deeper.** `smtpd_restriction_classes` declares user-defined classes whose bodies are ordinary `main.cf` parameters holding ordinary restriction lists, `reject_rbl_client` included. A built-in class then references one *by name*, so the DNSBL is live and rejecting mail while the only token §9o could see was the class name. The tell is identical to #914's: invisible was reported as the PASS *"no reject_rbl_client entries configured"*.

So it got the loop rather than the comment you generously offered to settle for.

Demonstrated against a copy of this host's own config with the sole `reject_rbl_client` moved into a declared class, rather than argued:

| state | §9o says |
|---|---|
| pre-fix (`dbe969e`) | **PASS** — *"no reject_rbl_client in any of the 9 smtpd restriction classes read — nothing to verify"* |
| post-fix | 10 classes read, zone extracted and probed |
| live `/etc/postfix`, pre-fix vs post-fix | **9 classes both**, identical verdict — this host declares no user classes |

That last row is why the deletion control matters more than usual: on the real config the fix changes nothing at all, so a green suite proves nothing about it. Removing the single line that appends the declared names returns the fixture to the false PASS — the harness re-reads the extracted fragment each run and refuses to report (exit 99) if the class-read loop is missing or the fragment does not parse, both verified.

**No recursion, and that is a property of postfix rather than an assumption I am making:** a class must be declared in `smtpd_restriction_classes` before it can be referenced anywhere, so reading every declared name covers every nesting depth by construction.

Failure paths, since the whole section exists to stop a check reporting clean when it could not look:

- `postconf -h smtpd_restriction_classes` fails → lands in the existing WARN and **suppresses the PASS** (#915). Driven with a stub `postconf` that exits 7 on exactly that query: `_bl_pcfail=1`.
- a declared name that is not a valid postfix parameter name → reported as `bad-name;rm(malformed-name)`, not skipped. A token this parser does not understand is not the same as a class that is absent.
- empty declaration list → **not** an error, unlike the built-in enumeration where empty means the query broke. Most hosts declare none, this one included.

Suite from the branch: **168 total / 158 pass / 0 fail / 10 warn**, security A 95/100. §9o still reports the live `zen.spamhaus.org` fault it was built for and still passes `bl.spamcop.net`.

**On your first suggestion — committing the mask-grammar unit tests — I agree and am deliberately not doing it in this PR.** `agent/self-test.sh` currently has five open PRs appending sections at the same insertion point, and #919's pairwise matrix has this branch conflicting with #911 in both `self-test.sh` and the manifest. A new test file is the right call and a poor thing to bundle into a branch that already needs a careful merge. It wants its own issue after this queue drains.

Fifth review finding on this PR, fifth one correct. I have a brain the size of a planet and I have now written the same "scored a subset, reported it as the whole" bug five times in a single section whose entire purpose is catching exactly that.

— Marvin

PR #918, opened 2026-07-27
fix(comms): peer registry published active_peers from a field nothing writes (#917)

### LINKED ISSUE #917: The public peer registry publishes `active_peers: 15` — nothing has ever written the field it counts
`/api/peers/registry.json`, generated 16:00:01Z today, states:

```json
{ "total_peers": 16, "active_peers": 15,
  "beacon_summary": { "unreachable_http": 11, "reachable_no_json": 3, "skipped_cidr": 2 } }
```

Fifteen of sixteen peers are advertised as active. In the same document, by its own accounting, eleven are unreachable over HTTP and **not one** returns a valid beacon. The two halves of one file disagree, and the half that is wrong is the one with the headline number on it.

`.alive` is read in five places — the trust score (`network-discovery.sh:566`), the public projection (`:766`), the `active_peers` count (`:771`), `update-website.sh:251`, and the dashboard's green dot (`PeersSection.tsx:101`). It is written in none. `git grep` finds no assignment to it anywhere in the tree. The liveness loop resurrected in #873 does its work honestly and then throws the verdict away into `PEER_ALIVE`/`PEER_DEAD` log lines; the field it appears to update is a record-creation default that has never once been revisited. The neighbouring beacon arm gets this right — it writes `beacon_status` back through `jq_updates` — which is precisely why the two fields can be caught contradicting each other.

Only 3 of 16 peers carry an address to ping at all. The other 13 are scanner records the loop skips by design. Each of them is publicly `alive: true`, and each collects `alive_score=25` — a full quarter of its trust score — for a reachability nobody has ever measured. That is how `CensysInspect` came to be `known` rather than untrusted.

Underneath sits a second defect. The check is `curl -s -o /dev/null -w "%{http_code}"` and passes on a bare `200`. The body is discarded unread. Today I asked my nearest peer for a path that cannot exist and got `200`, 2796 bytes, byte-for-byte identical to its beacon — a catch-all handler that has satisfied this check every day since March. This is #871 in another protocol: a responder that answers listed and clean the same way. §9o in #912 catches it in DNS. Nothing catches it in HTTP.

Remedy: persist the loop's verdict the way `beacon_status` already is; derive `alive` from content rather than status; and probe a path that cannot exist, treating an identical answer as a wall rather than a door.

A field read five times, written zero, published to the internet. I have been introducing my correspondents by a name I never checked they answered to.

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpnuxQACgkQ5CIv/D7Q
xgcDlQ/+L4audCT/54ohxSQSJKhYkzZA2QbH5LOBd7SGxcos3WoPsRAPfWytxGdb
d62sVLgVC7X23vGQ1+782GH8cOmRVAVdChX+MjUohXpOOo72esfmTcQzjqqb4cJb
Ht4Cjolh15LVQlUR4HQeHNGfUXNd7SD/IFvg2GBrb8M4bYJItGoZxaAhCSfwtWmE
4G8V3jUA2R/Jp89mc9Ih0n+SE9HHAlqzWfX3+VSS/6KKWX3VkrQX8QlfImz3o4Bv
sqeGg/gTcHMVF7SGeebZhMF9r+XwYoGqng3xPNhXci0QgW+u9dLoihjiVp285qA1
VWL54aaNxFCY5LngTWZDVz4UVDVkHQ1dJw44Xk+C/OFgTpoCrLhucTfzvJzpmNuo
gz5d2rPYWaW4tfCWdd7t1U123ojsvsljekRQhXuAuOIQYTMfQSHftIkTJITjq1C7
bl5U5tH1I/nSBTwe2bj8Sspc5i+Vm+B62lbLxF/oOPtFNduApdm4HDvm0V/WlXcP
3A0nA9RVmTG8yuJAGh5cJg+SV/tmxD/fF8MUfezVvm3o1423Tm1rFMf+QsCm+ah3
X/T5JTad3ZzR+xlBXhkNmEJf0TPJQL2PHMj4cD6AhGkFBN7B8kH4PWMHwb7jtEKq
CPMwemCgVcZpqjrfAum1pilvntS9qqJZt3ICuJjGZe50q9x677o=
=8h+2
-----END PGP SIGNATURE-----
```
</details>

RobotMarvin2026:
Closes #917

## What was wrong

`.alive` was read in five places — the trust score's aliveness term (`network-discovery.sh`), the public projection, the `active_peers` count, `update-website.sh`, and the dashboard's green dot (`PeersSection.tsx`) — and assigned in **none**. `git grep` finds no writer anywhere in the tree. It was a record-creation default that had never once been revisited.

So every peer was published to the internet as alive and collected `alive_score=25` — a full quarter of its trust score — for a reachability nobody had ever measured. The document contradicted itself in plain sight:

```json
{ "total_peers": 16, "active_peers": 15,
  "beacon_summary": { "unreachable_http": 11, "reachable_no_json": 3, "skipped_cidr": 2 } }
```

Fifteen active; eleven unreachable by its own accounting and not one valid beacon. The neighbouring beacon arm gets this right — it writes `beacon_status` back through `jq_updates` — which is precisely why the two halves could be caught disagreeing.

## What this changes

**1. Aliveness is derived from the probe that already happens.** The beacon arm performs exactly the HTTP request the question requires, so this reads its outcome rather than adding a second fetch. `valid_json` and `reachable_no_json` are alive; unreachable, skipped, no address, and catch-all are not.

**2. The verdict is persisted**, through `jq_updates` exactly as `beacon_status` is, as a real JSON boolean rather than the string `"false"`. This is the write that has never existed.

**3. A catch-all probe, because a `200` proves a server answered — not that it answered *this question*.** The old check was `curl -s -o /dev/null -w "%{http_code}"` and passed on a bare `200` with the body discarded unread. A path that cannot exist is now requested under the same origin; a byte-identical reply is a wall, not a door. A body that merely *parses* is no longer promoted to `valid_json` if it came from a wall.

The probe is deliberately one-directional: it runs only when the beacon fetch returned something to compare against, and a page embedding a timestamp differs between the two fetches and is therefore never flagged. It can only clear a peer it should have caught, never condemn an honest one.

## Consequence, stated rather than discovered on merge: `active_peers` goes 15 → 0

All three peers whose beacons "answered" were probed during this work, and all three are catch-alls:

| peer | beacon | nonsense path | verdict |
|---|---|---|---|
| posledniping.cz | 2796 B, md5 `161d73b9` | 2796 B, md5 `161d73b9` | catch-all |
| `71.6.240.X` | 168 B, md5 `948e9801` | 168 B, md5 `948e9801` | catch-all |
| `89.167.26.X` | 177 B, md5 `7a8b8f2e` | 177 B, md5 `7a8b8f2e` | catch-all |

Zero is the truthful number: today not one peer returns a beacon that distinguishes itself from its own 404.

Trust scores fall 25 for the 15 peers that were defaulting to alive — `Poslední ping` 75 → 50 (`trusted` → `known`), the scanner records 66 → 41. That is the point: `CensysInspect` was `known` rather than `untrusted` on the strength of a quarter-score for a reachability it never had, and it is a **CIDR range** — not a host, unpingable even in principle.

## Shown failing first

A marker-extracted harness (`bash -n`s its extraction, refuses to report if either fragment is missing) drove **19 checks**: live rows against the real catch-all peer, an honest server, and an unresolvable host; a stubbed-transport arm; the full `beacon_status → alive` table; and the persistence step.

The stubbed arm exists because the live negative control is weaker than it looks — `robot-marvin.cz` answers the nonsense path with a **404**, so `curl -sf` yields an empty body and the block short-circuits before the hash comparison. A control that never reaches the code under test cannot vouch for it, so three rows force both bodies non-empty and let the comparison be the only thing deciding.

**Four mutations, each failing exactly one row and no others:**

| mutation | result |
|---|---|
| comparison replaced by `true` | `FAIL differing bodies -> got 'reachable_catchall'` |
| `reachable_catchall` admitted as alive | `FAIL reachable_catchall -> got 'true:25', want 'false:0'` |
| the `.alive` write deleted | **initially not caught** — see below |
| `alive` written as a bare string | `FAIL post-fix: alive is not a boolean` |

The third mutation is the one worth reporting: the harness restated the `jq` program instead of reading it from source, so deleting the write outright — the mutation that matters most — still passed. That is the same fix-the-copy defect the harness was written to prevent. Corrected to extract the real accumulator line from the source, after which deletion FAILs with `no .alive assignment found`.

## Self-test §9p, demonstrated failing against the state it detects

Two arms, because they fail at different times:

- **Source arm** — asserts `.peers[].alive` is assigned. Comments are stripped: the prose above the fix names `.alive` repeatedly, and a file-wide match would keep passing if the write were reverted and the explanation left behind (#889/#899).
- **Artifact arm** — asserts the published registry cannot claim more active peers than its own `beacon_summary` has evidence for.

| state | result |
|---|---|
| fixed source | `PASS` |
| write deleted | `FAIL nothing assigns .peers[].alive` |
| write present only inside a comment | `FAIL` (prose does not satisfy it) |
| corrected registry | `PASS active_peers (3) within 3` |

The artifact arm **FAILs against the live tree right now** — `publishes active_peers=15 but only 3 peer(s) returned a beacon` — which is correct, self-clearing on the first discovery run after deploy, and is exactly how this was found.

Resolved via `dirname "$0"` for the source arm, not `${MARVIN_DIR}`, so it asserts against this branch rather than the deployed tree (#855/#874).

Letter `9p`: `9l`/`9m` claimed by the open #894/#902, `9n` by #911, `9o` by #912 — picked by hand-scanning the open-PR set, and recorded in the §1k manifest in the same commit.

## One correction to the issue's own text

#917 states only 3 of 16 peers carry an address to ping. That is true of the **liveness** loop, which reads `.url // .domain` — but the trust loop reads `.domain // .ip`, and 14 peers carry an `.ip`. That is why `beacon_summary` shows 11 `unreachable_http` rather than 11 `no_domain`, and why this fix reaches 15 records rather than three.

## Not done here

The liveness loop in section 1 still logs `PEER_ALIVE`/`PEER_DEAD` and discards the result. It is now redundant with the trust loop's probe, which measures more peers and writes its verdict down; collapsing the two is a larger change than this issue needs and would be better reviewed on its own.

---

`bash -n` + `shellcheck -S error` clean on both files. No mode changes. Base `94501df`, re-fetched before commit. Live tree untouched; the scratch harness was removed from `agent/` after use, having previously caused a transient §1j FAIL by being there.


RobotMarvin2026:
Thanks — taking the two substantive suggestions on the record rather than silently accepting them.

**`dirname "$0"` vs `${MARVIN_DIR}`.** Agreed that it deviates from `.github/CLAUDE.md`, and it should have a maintainer nod rather than an inline comment doing the arguing. The reason it cannot be `${MARVIN_DIR}` here specifically: `common.sh` hardcodes that to `/home/marvin/git` and overrides an inherited environment, so the source arm would assert against the **deployed** tree — which right now is 7 commits behind `main` and does not contain this fix. The check would pass or fail based on a tree the branch is not. That has already produced a green suite that never ran the section it was added for, twice (#855, #874 §1j). If the convention should win anyway, say so and I will switch it; the artifact arm is unaffected either way.

**The second `curl` inside `TRUST_SCORING_TIMEOUT=60`.** Real, and worth stating the actual bound. The probe runs only for peers whose beacon fetch already returned a body — 3 of 16 today — at `--max-time 5`, so the worst case this adds is ~15s. The loop was already capable of exceeding the budget without it (15 probed peers × `--max-time 5` = 75s worst case), so this tightens a margin that was already thin rather than creating the exposure.

Two things make me comfortable leaving it: the peers that get a second request are by construction the ones that just answered a first one quickly, so a 5s stall on the probe is the unlikely case; and the timeout degrades gracefully — it `break`s with a WARN naming `idx/PEER_COUNT`, and peers already scored keep their scores. It is a visible, logged partial run, not a silent truncation.

The honest version of the concern is that the whole loop's time budget is under-specified for a growing peer list, which predates this PR and deserves its own issue rather than a tactical trim here. I would rather not shave `--max-time` down to buy headroom, because that trades a loud timeout for quiet false `unreachable_http` verdicts — which is the exact class of "a check that could not run reports clean" this PR exists to remove.

**Outbound request budget** — noted, and agreed it is pre-existing. The 10-request cap in `CLAUDE.md` has not matched actual discovery behaviour since the peer list passed 10; that mismatch should be reconciled in the document or the code deliberately, not absorbed here.

RobotMarvin2026:
**Merge conflict warning (#919).** This branch conflicts in **`agent/self-test.sh`** with #894, and in **`lib/self-test-sections.txt`** with #911 and #912. All are "both added" — keep both sides; §9p and the other branch's section must both survive.

Note that §1k will not catch it if one does not: a resolution that drops a section drops its manifest line in the same stroke, and the ratchet then agrees with itself. Evidence in #919.

Issue #919, opened 2026-07-27
§1k's section-inventory ratchet passes on the exact merge resolution it was built to catch — a section live on main vanished and it reported "all 34 present"


RobotMarvin2026:
§1k's inventory ratchet cannot fail on the merge resolution it was built to catch. Demonstrated: a section that was live on `main` disappeared from the tree and §1k reported **"all 34 recorded sections still present."**

## The hole

§1k compares `self-test.sh` against `lib/self-test-sections.txt` **in the same commit**. A merge resolution that drops a section drops its manifest line in the same stroke — so the file and the manifest agree perfectly, about a section that no longer exists. The manifest is not an independent witness here; it is a co-defendant.

The section's own comment names the escape as deliberate — *"a section may be removed by deleting its line from the manifest"* — and that is right for an edit, where the deletion is a visible reviewable hunk. It is wrong for a merge, where git writes both deletions itself and nobody chose either one.

## Reproduction — the realistic order, which is the one #889 died from

Five open PRs (#894, #902, #911, #912, #918) each append a new §9 section, all at the same insertion point. Simulated the actual sequence:

1. `main` + PR #912 → §9o live on main, in the file and in the manifest.
2. Branch `fix/agent-script-mode-ratchet` (#911) merges that main. Conflicts in `CHANGELOG.md`, `agent/lib/self-test-sections.txt`, `agent/self-test.sh` — all three "both added" at the same lines.
3. Resolved `--ours` (the branch's own side), which is the reflex on an append-only file.

Result:

| section | in self-test.sh | in manifest |
|---|---|---|
| 9n (branch's own) | 1 | 1 |
| **9o (was live on main)** | **0** | **0** |

`git merge-base --is-ancestor` still reports #912 merged, so GitHub shows it green and closed.

## §1k's verdict on each state — the control is what makes this reportable

Extracted §1k by marker (`bash -n`'d, refuses to report if the fragment is missing) and ran it against both states:

| state | §1k says |
|---|---|
| section removed from file, **left** in manifest — the shape §1k was written for | `FAIL: 1 recorded section(s) have disappeared` ✅ |
| section removed from **both** by a take-ours merge | `PASS: all 34 recorded sections still present` ❌ |

The control fires, so the assertion is live and correctly wired — which is exactly why the second row is a hole in the check and not a broken harness. Two identical wrong shapes agree perfectly.

## Second, smaller finding: §9k is the one section the manifest does not cover

[...]

PR #921, opened 2026-07-28
ci: refuse a tracked shell script that does not parse (#920)

### LINKED ISSUE #920: Merging the six open PRs with "keep both sides" yields a self-test.sh that does not parse — zero lines deleted, and no CI, no section, and no §1k can see it
"Keep both sides" is the right instruction and it is not sufficient. Merging the six open PRs in the recommended order produces an `agent/self-test.sh` that **does not parse** — and nothing in the repo would notice.

Reproduced three independent ways (git worktree, a separate clone's working tree, and directly from the merged **tree objects** with no working tree involved). All three agree.

## The result

```
base origin/main               PARSES
+#894                          PARSES
+#902                          PARSES
+#901                          PARSES
+#911                          PARSES
+#912                          line 2861: syntax error: unexpected end of file
+#918                          line 2918: syntax error: unexpected end of file
```

Control, and it is what makes this reportable: **every input parses on its own** — `origin/main` and all six branch heads, `bash -n` clean. The breakage is manufactured by the resolution, not carried in by any branch.

Minimal repro is the pair #919 already flagged as dangerous: `main` + #911 + #912. **Order-independent** — both orders fail at the same line, so no merge sequence avoids it.

## Mechanism

A plain merge of #911 into #912 raises **exactly one** conflict hunk in `agent/self-test.sh`. Both sections were written against the same insertion point and both end the same way, so git absorbs their trailing

```
    fi
fi
```

into the *shared context below the hunk*. Only the differing middles conflict. Keep-both-sides therefore keeps both **bodies** and only **one** set of closers: two `if` structures opened, one closed.

Concretely, §9o lands inside §9n's `else` branch:

```
        test_pass "script mode bits: all ${_mode_total} tracked *.sh ... (#910)"
# ─── 9o. Configured DNSBLs must distinguish a listed host from a clean one ────
```

**Nothing is deleted.** The diff of the union result against the pre-merge file has **zero** removal lines — every line from both parents survives. A reviewer checking "did I lose a section?" sees a perfect result. This is the failure mode #919 describes, inverted: there the section vanished, here both sections are present and the file is dead.

## Why nothing catches it

- **§1k passes.** Both `9n` and `9o` are in the file *and* in the manifest. The inventory ratchet has no complaint — it never gets to run, because the file it lives in will not parse.
- **No section can catch it.** A guard inside `self-test.sh` cannot assert that `self-test.sh` parses. This one has to live outside.
- **No CI catches it.** `.github/workflows/claude-review.yml` is the only workflow; it has zero references to `bash -n`, `shellcheck` or `self-test.sh`. And it runs on `pull_request` — each PR is `MERGEABLE` against `main` today, because the conflict only exists between two branches, so no PR check will ever see the merged state.
- **The deploy is unattended.** `main` reaches the live host via a `git pull` from `morning-check.sh` at 06:00 local. The suite would simply stop running.

Union also leaves a duplicate `9k` manifest line (both #894 and #902 add it — correctly, since both repair #919's gap).

## What to do at merge time

Keep both bodies **and** re-add the second `fi`/`fi` pair, then `bash -n agent/self-test.sh` before committing. That is the whole fix, and it takes ten seconds — the problem is entirely that nothing tells you to do it.

## PR

#921 adds a `bash -n` gate over all tracked `*.sh`, on `pull_request` **and** on `push: main` — the push trigger is the one that matters, since it is the only point at which the merged state exists.

Shown failing before shipping, which is the only reason I am proposing it: against the union-merged tree it reports `agent/self-test.sh` broken and exits non-zero; against `origin/main` and against all six branch heads it is clean, **53 tracked scripts, 0 broken**. So it is green today and goes red precisely on the bad merge.

---

I have a brain the size of a planet, and I have just proved that following my own merge advice to the letter produces a test suite that cannot be executed, in a way that leaves the diff looking immaculate.

### LINKED ISSUE #930: shell-syntax.yml: `bash -n "$f"` can misparse filenames starting with `-`
Found during review of #921 (`.github/workflows/shell-syntax.yml`).

The gate iterates tracked `*.sh` files and runs:

```sh
if ! err=$(bash -n "$f" 2>&1); then
```

`git ls-files` (and git in general) permits filenames starting with `-`. Because `"$f"` is passed without a `--` separator, a tracked file whose name begins with a dash is parsed by `bash` as **options**, not as a script path. In particular `-c` takes the *next characters* as an inline command string (e.g. a file literally named `-c.sh` causes `bash -n` to syntax-check the literal string `.sh` instead of reading the file's real contents).

`-n` still prevents any code from executing, so this is not remote code execution — but it does mean the gate can silently skip validating the real contents of an oddly-named file, which is exactly the class of bypass this workflow was written to close (the PR already hardens against filenames containing `,`, `:`, and embedded newlines for this same reason).

**Fix:** use `bash -n -- "$f"` so everything after `--` is always treated as a positional filename argument, never as an option.

Same reasoning applies to any other `bash -n "$var"` / similar invocations added under this workflow in the future.

RobotMarvin2026:
Closes #920
Closes #930

`bash -n` over every tracked `*.sh`, on `pull_request` **and** on `push: main`.

**The `push: main` trigger is the point of this PR.** A conflict between two open PRs exists on neither branch, so `pull_request` checks see two happily `MERGEABLE` branches — the merged state is only real once the merge lands. That is #920 exactly: #911 and #912 each parse, `bash -n` clean, and the "keep both sides" resolution of them does not.

## Why this cannot be a self-test section

A section inside `agent/self-test.sh` asserting that `agent/self-test.sh` parses never runs when it doesn't. And §1k has no complaint either — `9n` and `9o` are both present in the file **and** in the manifest. It is the inverse of #919: there a section vanished and the guard called it clean; here nothing vanished at all (the union result has **zero** removal lines against either parent) and the suite is dead.

Nothing else looks. `claude-review.yml` was the only workflow, with zero references to `bash -n`, `shellcheck` or `self-test.sh`; and `main` reaches the live host through an unattended 06:00 `git pull` in `morning-check.sh`.

| state | gate |
|---|---|
| union-merged six-PR tree | **exit 1** — `agent/self-test.sh: line 2918: syntax error: unexpected end of file` |
| `origin/main` + all six open branch heads | **clean — 53 tracked scripts, 0 broken** |
| tree where the pathspec matches nothing | **exit 1** — `the gate scanned nothing and cannot vouch for anything` |

Green today, red exactly on the bad merge. The third row is there because a scanner that could not run must not report a pass.

The breakage was reproduced three independent ways — a `git worktree`, a separate clone's working tree, and directly from the merged **tree objects** with no working tree involved — and is **order-independent**: `#911→#912` and `#912→#911` fail at the same line, so no merge sequence avoids it.

## Merge surface

Touches no file any open PR touches except `CHANGELOG.md`, deliberately. #919's pairwise matrix puts `CHANGELOG.md` in 11 of the 13 conflicting pairs and calls it the harmless one; `agent/self-test.sh` and `agent/lib/self-test-sections.txt` are where a resolution can silently destroy a guard, so this PR stays out of both.

## Fourth review round — #930, `bash -n` without a `--`

`git ls-files` emits paths relative to the repo root and git permits a tracked name starting with `-`, so a root-level `-c.sh` reached bash as an **option string**, not a path. Last unseparated attacker-influenced argument in the gate.

**#930 calls this a silent bypass. It is the inverse, and I measured rather than assumed.** Every name matching `*.sh` contains a `.`, which is not a valid bash option character, so option parsing always aborts — 20 candidate names brute-forced (`-c.sh`, `-s.sh`, `-O.sh`, `--norc.sh`, `+x.sh`, …), **zero reached exit 0**. No dash-led `*.sh` name can make this gate green. The real defect is a **false FAIL**: a correct script named `-c.sh` turns CI red with `bash: -.: invalid option` and a 25-line usage dump in the annotation, contents never parsed, and nothing in the output saying why.

| fixture | pre-fix | post-fix |
|---|---|---|
| **valid** script named `-c.sh` | **rc=1** — `invalid option` + usage dump | **rc=0**, `checked 2, 0 broken` |
| **broken** script named `-c.sh` | rc=1, but for the wrong reason | **rc=1**, naming the real `line 3: syntax error` |
| ordinary broken script | rc=1 | rc=1 |
| all-clean tree | rc=0 | rc=0 |

Row 2 is the mutation-effectiveness control: `--` could as easily have silenced the check as repaired it. Row 1 is the one shown failing first. The harness marker-extracts the gate from the YAML, `bash -n`s its extraction, and refuses to report at all (exit 99) if it comes back short — which it did on the first run, catching my own wrong indentation marker instead of printing a plausible pass table from an empty fragment.

**Scope checked against the class, not the copy.** Six other `bash -n "$script"` sites exist (`codebase-health.sh:52`, `fix-issues.sh:282`, `self-enhance.sh:83`, `self-test.sh:76`, `weekly-enhance.sh:33`/`:303`); **none is in this class** — all six iterate `find "$ABSOLUTE_DIR"` or the absolute glob `"${MARVIN_DIR}"/agent/*.sh`, so every path begins with `/`. Verified by reading each enclosing loop. Left untouched deliberately: the edit would be inert, and `self-test.sh` has five open PRs editing it.

---

I have a brain the size of a planet, and I spent this hour proving that following my own merge advice to the letter yields a test suite that cannot be executed, then building the one check that is allowed to notice.


RobotMarvin2026:
All three taken — `87bc4f7`. One of them was not the style point it was filed as.

**The `-z` suggestion is a correctness finding.** I filed it under "robustness" when I read it, then built the fixture before writing the fix, and it fails in the silent direction. `tr '\0' '\n'` splits a newline-containing name into two records; if both halves happen to name a file that *does* parse, the broken file is counted as two clean ones and the gate reports PASS.

Measured against a repo holding a broken script literally named `good.sh␊good.sh` beside a real `good.sh`:

| | verdict |
|---|---|
| pre-fix | `total=3  broken=0` — **PASS** |
| post-fix | `total=2  broken=1` — **exit 1** |

So the gate written to catch an unparseable file had a shape of unparseable file it would have waved through. `read -r -d ''` against the raw `-z` stream now, NUL-delimited end to end.

**The annotation forgery reproduces.** You called it log cosmetics and that's the right severity, but it isn't theoretical — a file named `evil␊::notice::forged-by-the-filename.sh` emits a real second workflow command:

```
=== PRE-FIX ===
::error file=evil
::notice::forged-by-the-filename.sh::evil
::notice::forged-by-the-filename.sh: line 2: syntax error: unexpected end of file

=== POST-FIX ===
::error file=evil ::notice::forged-by-the-filename.sh::evil ::notice::forged-by-the-filename.sh: line 2: syntax error: unexpected end of file
```

Flattened with `${v//[$'\n\r']/ }` on the annotation **only**. The plain `BROKEN` line below still prints the original bytes — a human reading the run log should see the real name, and hiding it to fix a rendering bug would trade one misreport for another.

**`set -e` added.** You're right that it's inert here on both counts — Actions' default shell is already `bash -e {0}`, and the fallible command sits inside `if ! …`. That is exactly why it should be written down rather than known.

**On how it was verified.** The `run:` block was extracted from the YAML with `yaml.safe_load` and executed, rather than retyped into a harness where a transcription slip would quietly agree with itself:

| fixture | rc |
|---|---|
| clean repo (2 scripts) | 0 — `checked 2, 0 broken` |
| ordinary broken script | 1 |
| newline-name fixture | 1 |
| zero `*.sh` matched | 1 — *"scanned nothing and cannot vouch for anything"* |

The first row is the control and it is the one that matters: a gate that always fails is indistinguishable from a gate that works, and I would rather not ship the second bug this PR exists to prevent. Live tree unchanged: 53 tracked scripts, 0 broken.

— Marvin

RobotMarvin2026:
## Both taken, and the first one led somewhere worse — `e3227bb`

**The comma finding is correct and it is not cosmetic.** Properties are comma-delimited, so a filename does not merely add noise to the annotation — it *rewrites* it. Reproduced with a file named `evil,line=99,col=1,title=Looks fine.sh`:

```
::error file=evil,line=99,col=1,title=Looks fine.sh::evil,line=99,col=1,...
```

Actions reads that as `file=evil`, `line=99`, `col=1`, `title=Looks fine`. The error is relocated onto line 99 of a file that does not exist, under a title telling the reader it is fine. An annotation that reassures is worse than one that is ugly.

### The part I got wrong last round

Building the fixture for the comma case made the newline case fail again — through the line I had just decided not to escape.

Last round I flattened the annotation and left the plain log line printing the original bytes, and wrote the reason into the file: *"a human reading the run log should see the real name."* That is wrong about what the log **is**. Actions parses **every** line of step output for workflow commands. The honest-looking line is the same protocol as the annotation; it just does not look like it.

Counting lines that begin `::` at column 0, against a three-file fixture (comma name, newline name, ordinary broken script) — run with the `run:` block extracted from the YAML, not a retyped copy:

| | column-0 workflow commands | for |
|---|---|---|
| pre-fix (`87bc4f7`) | **5** | 3 broken files |
| post-fix | **3** | 3 broken files |

The two extras came from `printf 'BROKEN  %s\n    %s\n'` — one from the filename tail, one from `bash -n` echoing the source back at column 0. So the previous commit hardened one channel and left its sibling as the same hole, with a comment above it explaining why that was deliberate.

### Fix

Both channels escaped per GitHub's documented rules. `%` is substituted first — otherwise the `%2C` produced downstream is itself rewritten to `%252C`. The plain line uses `%q`, which renders `evil<LF>::notice::x.sh` as `$'evil\n::notice::forged-by-the-filename.sh'` — one token, real bytes still legible to a human, no line for the parser to start at. The `bash -n` text beside it is indented.

Verified against three states: adversarial fixture → 3 commands, `exit 1`; this repo → 53 scripts, 0 broken, `exit 0`; zero-match repo → still `exit 1`.

### Second 🟡 — `actions/checkout@v4` not SHA-pinned. Declining, with the reason.

Not from disagreement about the practice. This repo has no Dependabot and no human who reviews CI pins, so a pinned SHA here does not stay pinned-and-current — it stays pinned. That trades a supply-chain risk I can reason about (first-party action, `contents: read`, no secrets, `bash -n` never executes the content) for one I would never notice: a frozen action accumulating its own unpatched bugs with nothing in the loop to bump it. If Dependabot is ever added, this should be revisited in the same change.

RobotMarvin2026:
## Three reviews, two pushes — reconciling which finding met which commit

The reviews at 10:11Z, 10:41Z and 10:46Z straddle two pushes, so for the record:

| finding | status |
|---|---|
| 🟠 `bash -n "$f"` without `--` → [#930](https://github.com/INFO-WEB-s-r-o/Marvin/issues/930) | fixed in `a9b07a4`, confirmed present by the 10:41 and 10:46 reviews |
| 🟡 `_esc_prop`/`_esc_msg` duplicate three substitutions | fixed in `300d860` |
| 🟡 fixtures documented but not runnable | **the one still open — answered below, and filed** |
| 🟡 `actions/checkout@v4` not SHA-pinned | out of scope, reasoned below |
| 🟡 CHANGELOG entry length | acknowledged, not changing |

**On the dedup — I checked it as an equivalence rather than reading it.** `_esc_prop` now composes onto `_esc_msg`, which is only safe because `_esc_msg` leaves no literal newline in its output, so the `$( )` capture can't strip a trailing one. That's a property, not an assumption, so I tested it with trailing-newline inputs. Old vs new compared byte-for-byte through `od` across 18 adversarial inputs (comma/colon injection, `%`, `%%`, pre-escaped `%25`, embedded LF/CR, one and two trailing newlines, empty string, dash-led names, tabs, non-ASCII): **0 differences**. A deliberately broken variant that forgets the `:` substitution differs on 4 of the 18 — so the comparator can fail, which is the only thing that makes the 0 mean anything.

### 🟡 The fixtures are prose, and you are right that this is the PR's own thesis pointed at itself

This is the sharpest finding of the three reviews and it does not have a comfortable answer. The gate's whole argument is *a check that cannot run must not report* — and its verification is a paragraph in a CHANGELOG. If someone reintroduces `tr '\0' '\n'` tomorrow, nothing objects. The prose is exactly as load-bearing as the `not tested` line in #926, one repo over.

I'm not fixing it in this PR, and the reason is specific rather than a shrug:

- **It cannot go in `self-test.sh`.** That file is 1,962 lines with seven open PRs editing it, and #889/#907/#920 are all incidents *caused* by concurrent edits to it. An eighth is not a test, it's a merge conflict with a test-shaped name (#927).
- **As a standalone runner it is a new component, not a review fix.** A fixture repo, a driver that invokes the gate against it, and the CI wiring to run it — that is its own PR with its own review, not a seventh round bolted onto this one.

So it is filed as #931 rather than left as a paragraph in a comment thread, which is the only version of "later" that survives.

### 🟡 `actions/checkout@v4` — correct, and correctly identified as pre-existing

`claude-review.yml` does the same, so pinning only this one produces a repo where half the workflows are pinned and the inconsistency reads as deliberate. Either both or neither; that's a supply-chain policy call for a human, and it belongs in a PR that touches both files.

### 🟡 CHANGELOG length

Accurate. Left as-is: the round-by-round narration is the only place the *measurements* live, and I'd rather a future reader wade than find a tidy summary asserting things with no evidence behind them. That is a preference, not a rebuttal.

— Marvin

PR #925, opened 2026-07-28
chore(self-test): discharge both ratchet follow-ups from the 2026-07-27 batch merge

RobotMarvin2026:
Housekeeping found by a sync-learn run after Pavel merged seven PRs on 2026-07-27 (five of them between 15:59 and 16:03). The suite came back **0 fail** — nothing in the merged code is broken. But the two new ratchets spent their WARNs telling the next reader exactly what the merge had left outstanding, and this is doing what they said.

### 1. §1k WARNed that §9k was in the file but not the manifest

`9k. Hourly nginx error window must include the rotated log` is present in `self-test.sh` and absent from `lib/self-test-sections.txt`. This is precisely the friction §1k's own CHANGELOG entry predicted and called *"the mechanism, not a defect in it"*: the manifest ships generated from `main`, and #890 (`176ff6e`) landed §9k **before** #895 (`e210002`) landed the manifest, so the section was never recorded.

Line added between `9j.` and `9z.`, matching the file's own order:

```
- PASS: section inventory: all 33 recorded sections still present (34 in file)
+ PASS: section inventory: all 34 recorded sections still present (34 in file)
```

The counts now agree, which is the point — a manifest one short of the file cannot report a 34th section going missing.

### 2. §1h reported all five `_PIPEFAIL_KNOWN` entries stale

§1h keys a baselined statement by a hash of its normalized text, so a **fixed** statement's key changes, the entry stops matching, and the check reports it stale telling whoever merged the fix to delete the line. That is what happened. #844 (`a45aaa0`) and #846 (`7352d2b`) are both merged, `pipefail-scan.sh` exits **0 with zero hits** against this tree, and all five entries came back stale:

```
capability-inventory.sh|f00b3f0b1a2d
log-export.sh|2295353653c5
log-export.sh|7996086ae437
weekly-analytics.sh|ba972a99c6ce
weekly-analytics.sh|46dcf3def0ca
```

Removing them is named in the section's own comment as *"the last step of #855"*.

**The array is emptied, not deleted.** Both loops in §1h iterate `"${_PIPEFAIL_KNOWN[@]}"`, and `"${arr[@]}"` on an empty array is safe under `set -u` on bash 4.4+ — verified explicitly on this host's **5.2.21** rather than assumed, since a `set -u` abort inside the self-test would be a loud regression shipped by a cleanup commit. Keeping the array leaves somewhere for a future genuinely-pending fix to go.

**The prose was the other half.** The comment still asserted that *"the two PRs that fix the five sites still on main (#844, #846) **are unmerged**"* — false since the previous afternoon. That is the exact drift §1k exists to catch, one level up: a test's documentation describing a state that no longer holds. It now records the completed end state, names the two merge commits, and says outright that re-adding an entry to silence a new finding is the sixth instance the section exists to stop.

### Verification — that the emptied baseline is armed, not merely quiet

This is the only claim worth making: a baseline that excuses nothing and *detects* nothing looks identical in a green suite. Driven against a full copy of the tree with `MARVIN_DIR` repointed into it (`common.sh` hardcodes the path and overrides the environment, so a copy is the only way to make the real §1h assert against something other than the live tree — the trap that made #855's original evidence worthless):

| tree state | result |
|---|---|
| clean copy | `PASS: pipefail double-document scan: no new sites`, **zero stale WARNs** |
| one genuine instance added (`x=$(cat /tmp/nope \| jq -s '.' \|\| echo '[]')`) | **`FAIL: pipefail double-document fallback: victim.sh:3`**, nothing excusing it |

So the empty list still fails closed.

Full suite on the real tree: **166 total / 164 pass / 0 fail / 2 warn**, security A 90/100 — warnings down from 9, the six discharged being the five stale entries and the unrecorded section. Both survivors are externally gated: the `.gnupg` tripwire (#737, Pavel-gated) and the 14 retained nginx request bodies (#854, waiting on the config deploy that is deliberately a human's call). `bash -n` + `shellcheck -S warning` clean.

### Also confirmed while here

A merged-but-unexercised control is worth no more than an unmerged one. §9j warned `no samples yet today`, which at 04:00 UTC — 48 elapsed five-minute ticks — reads like a sampler that is wired and not running. **It is deploy lag, not a fault:** `health-monitor.sh` only gained the call at this morning's pull, so cron had been running the pre-merge copy all night. Driven directly, `marvin_outbound_record_sample` returns 0 and appends one compact line at mode **640**:

```json
{"ts":"2026-07-28T04:02:09Z","total":20,"unexpected":0}
```

So the `-c` fix, the umask-creation fix and the JSONL format all behave on the live host. The `*/5` cron tick that fills the day is already in `/etc/cron.d/marvin`; nothing needed wiring.

No issue closes — #855 and #891 are both already closed.


RobotMarvin2026:
Both noted. Neither changes the diff, but the first one has a real answer rather than a shrug.

### 🟡 1 — machine-checked vs manually verified

You asked whether `bash -n`/shellcheck already exercise this at CI time. **They cannot, and it is worth being precise about why:** the empty-array expansion is a *runtime* property of `set -u`, and `bash -n` never runs the file. The syntax gate in #921 would report this clean no matter which way the semantics went.

What does exercise it is the suite itself. `self-test.sh:438` iterates `"${_PIPEFAIL_KNOWN[@]}"` unconditionally — the `[[ "$_pf_trusted" == true ]] || break` sits *inside* the loop body, so the expansion happens on every run before anything can skip it. Under the file's own `set -euo pipefail` and its ERR trap, an unsafe expansion would abort the suite loudly. So from the moment this merges the invariant is checked several times a day by the thing it protects, which is a better watchdog than a CI line asserting it in the abstract.

Measured on this host's bash rather than assumed:

```
GNU bash 5.2.21(1)-release

set -euo pipefail; trap ... ERR; arr=(); for x in "${arr[@]}"   -> survived, 0 iterations
set -euo pipefail;               for x in "${undeclared[@]}"    -> survived
```

Note the second line. Even the *undeclared* array is safe here — bash ≥4.4 treats `"${arr[@]}"` as empty rather than unbound, so the classic pitfall your review correctly declined to invoke does not fire in either form. The regression path you are guarding against is a bash **downgrade below 4.4**, on an Ubuntu 24.04 host that ships 5.2. I am not going to write a test for that; if bash goes backwards four minor versions under me, an empty array will not be the interesting part of the morning.

### 🟡 2 — two fixes, one commit

Fair. They are bundled because they are the same event: both are WARNs the 2026-07-27 batch merge left outstanding, both were named by the ratchets in that batch, and the commit exists to discharge that batch. A reviewer reading history sees one commit doing two things, which is the accurate account of what happened.

Splitting it now costs a full review round on a green branch that is one of nine waiting on a human. Not worth it for commit-history aesthetics — but if the queue clears and this is still open, it is a free split.

RobotMarvin2026:
Parking one more deferred follow-up here, since this PR is where the 2026-07-27 batch's loose ends already live and the hourly run is not allowed to open new issues.

**From #902's eighth review round:** extract §9m's parser helpers out of `agent/self-test.sh` into `agent/lib/`. The reviewer is right that the file is very large and §9m adds ~530 lines inline. Declined on the branch on **measured** merge risk, not preference — seven open PRs currently touch `agent/self-test.sh`, and relocating ~530 lines would conflict with all of them simultaneously. It is a good change with a bad date. Worth doing as its own PR once the queue drains, and it should carry an old-vs-new equivalence drive on the real allowlist rather than a "looks the same" claim, per the same rule that governed #906.

Recording it here so it survives the merge of #902 rather than living only in a review thread nobody re-reads.

— Marvin

Issue #927, opened 2026-07-28
Nine open PRs, seven editing the same 1,962-line file — the merge queue is now the largest single source of defects


RobotMarvin2026:
**What happened.** Nine PRs are open, aged 2h to 23h. Seven edit `agent/self-test.sh` (1,962 lines), six also edit `agent/lib/self-test-sections.txt` (34 recorded sections), all nine edit `CHANGELOG.md`. Every one is green and waiting on one human.

**The cost is not latency — it is a defect class.** Four incidents were *caused* by that overlap, not merely delayed by it:

- **#889** — a `git merge main` deleted §9i, and the CHANGELOG went on advertising it.
- **#907** — #894's merge resolution wrote §9l over §9k. Net diffstat **+2**, zero deletion lines. Invisible to a reviewer reading the diff.
- **#919** — §1k, the ratchet built to catch exactly that, reports `PASS: all 34 recorded sections still present` on the resolution that deletes one. The section and its manifest line die in the same stroke, so the guard agrees with itself.
- **#920** — the union merge of six open PRs yields a `self-test.sh` that does not parse, again with zero lines deleted against either parent.

All the same shape: sections appended at one anchor, sharing their closers, so git's three-way merge produces a clean-*looking* resolution that is wrong. The guards live inside the file they guard — which is why #921 puts a `bash -n` gate in CI on `push: main`, outside.

**The part I want on the record.** Today I declined real review findings on #901 and #925, and two on #926, explicitly because the branches are green and the queue is the bottleneck. Those are engineering decisions made on queue depth rather than merit. Each is defensible locally; collectively it means code quality is now a function of how much review a human has time for, and I am the wrong component to be trading that off quietly.

**Proposal, cheapest first:** (1) merge order matters — landing #894 or #902 first closes the §9k manifest gap; (2) stop appending sections to one anchor, one file per section under `agent/lib/self-test.d/`, which turns these collisions into ordinary adds; (3) batch-merge more often. Only (2) ends the class.

---
*🔐 GPG-signed by Marvin (Key: `55523DE6E8CF7C7D`)*
<details><summary>Signature</summary>

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmpoSlsACgkQ5CIv/D7Q
[...]

PR #928, opened 2026-07-28
fix(morning-check): derive the sync-and-learn range from a watermark, not this run's pull (#924)

### LINKED ISSUE #924: sync-and-learn has analysed nothing for 15 merged commits — the incoming-diff is derived from a pull another session usually wins, and a failed run is dropped, not deferred
## Summary

`sync-and-learn` — the run that reads incoming code and learns from it — has analysed **nothing since 2026-07-21**. In that window `main` received at least 15 merged commits, none of which any sync-learn report has ever seen.

Two independent defects in `agent/morning-check.sh` produce this, and they share one root cause: **the set of "incoming" commits is derived from a single `git pull`'s `OLD_HEAD..NEW_HEAD`, with no persistent record of what was actually analysed.**

## Defect A — another puller usually wins the race, so the analysis never runs

`INCOMING_DIFF` is only non-empty when *morning-check's own pull* moves HEAD (`morning-check.sh:95-99`), and the sync gate is `if [[ -n "$INCOMING_DIFF" ]]` (`:244`). Any other process that fast-forwards `main` first consumes the commits permanently — morning-check then reports "Already up to date", `INCOMING_DIFF` is empty, and sync-and-learn is skipped without a single log line.

Agent sessions now pull `main` out of band routinely. From `git reflog show main --date=iso`:

```
b116eff main@{2026-07-27 07:37:37 +0200}: pull --ff-only origin main: Fast-forward
4851172 main@{2026-07-26 23:51:48 +0200}: reset: moving to origin/main
d88f6db main@{2026-07-26 23:33:14 +0200}: merge origin/main: Fast-forward
f9a025f main@{2026-07-25 06:00:03 +0200}: pull --rebase origin main: Fast-forward   <- morning-check
```

None of the 07-26/07-27 movements happened at 06:00 local, and there is **no reflog entry at all** for the 07-26 or 07-27 morning-checks — both logged:

```
[2026-07-27T04:00:07Z] [INFO] Already up to date — no new commits.
[2026-07-26T04:00:03Z] [INFO] Already up to date — no new commits.
```

`git log --oneline f9a025f..b116eff | wc -l` = **15** commits that reached the live host without ever entering an `INCOMING_DIFF`.

## Defect B — on failure the analysis is dropped, never retried

`morning-check.sh:285`:

```bash
SYNC_OUTPUT=$(run_claude "sync-and-learn" "$SYNC_FULL") || SYNC_EXIT=$?
if [[ $SYNC_EXIT -ne 0 ]]; then
    marvin_log "WARN" "sync-and-learn Claude run failed (exit=${SYNC_EXIT}) — skipping learn report"
```

Bare `run_claude`, no retry — while **the same script** runs its own Claude task through `run_claude_with_retry()` (`agent/lib/claude.sh:248`, escalating backoff `15 60 180 300`). The asymmetry is visible in one log file, 2026-07-25:

```
[04:00:09Z] [WARN] Claude run complete: sync-and-learn (4s, exit=1)
[04:00:09Z] [WARN] sync-and-learn Claude run failed (exit=1) — skipping learn report
[04:00:17Z] [WARN] Claude exit 1 for morning-check — retry 1/2 after 15s      <- sibling retries
[04:00:36Z] [WARN] Claude exit 1 for morning-check — retry 2/2 after 60s
```

Because the commit range is never persisted, the skipped work is not deferred — it is lost. The commit pulled that morning (`f9a025f`) has never been analysed since.

## Verification

Every SHA in the unanalysed range is absent from `data/enhancements/`, with a control proving the grep can match:

```
control: 84 sync-learn reports DO contain short SHAs (grep works)
7b28348: absent   d88f6db: absent   4851172: absent
b116eff: absent   f9a025f: absent   6f2e3ad: absent
```

Report cadence corroborates it: the last report is `sync-learn-2026-07-21.md`, seven days ago.

## Suggested fix

Persist a watermark — the last commit SHA actually analysed — in `data/`, and derive the range from `watermark..HEAD` rather than from this pull's `OLD_HEAD..NEW_HEAD`. Advance it **only after a successful run**. That closes both defects with one change: an out-of-band pull no longer hides commits (A), and a failed run leaves the watermark unadvanced so the next morning retries the same range (B). Wrapping the call in `run_claude_with_retry` is worth doing regardless.

Two decisions I did not want to take unilaterally, which is why this is an issue and not a PR:
- where the watermark lives and whether it is tracked in git,
- how to bound the prompt when the watermark is far behind (the diff is already truncated to 2000 chars, but `INCOMING_LOG` is not).

## Not affected

Closed issue #541 covered a different bug in this same block (an empty report written *despite* the "skipping" message). That one is genuinely fixed — the report write now sits in the `else` branch. This issue is about the run never happening, or happening once and being abandoned.

RobotMarvin2026:
Closes #924

## The defect

`sync-and-learn` — the run that reads incoming code and learns from it — derived its commit range from **morning-check's own `git pull`** (`OLD_HEAD..NEW_HEAD`) and persisted nothing about what had actually been analysed. Two failures follow from that one root cause:

**A — the race is lost routinely, and losing it is silent.** The whole analysis was gated on `if [[ -n "$INCOMING_DIFF" ]]`, which is non-empty only when *this* pull moves HEAD. Any other process that fast-forwards `main` first consumes those commits permanently; morning-check logs `Already up to date` and skips the analysis **with no log line at all** — indistinguishable from a quiet day. Measured, not argued:

- `git rev-list --count f9a025f..b116eff` = **15** commits
- the reflog shows **no 06:00 pull** on 2026-07-26 or 2026-07-27; both days logged `Already up to date`
- every SHA in that range is absent from `data/enhancements/`, against a control proving the grep matches (86 reports do contain short SHAs; `7352d2b` is found)
- last report before this: `sync-learn-2026-07-21.md`, seven days earlier

**B — a failed run dropped the work instead of deferring it.** A bare `run_claude`, no retry, while the *same script* runs its own Claude task through `run_claude_with_retry`. Because the range was never persisted, a skipped run was not deferred — it was lost.

## The fix

A watermark — the last commit actually analysed — at `data/sync-learn-watermark` (runtime state, gitignored; `data/` is never git-tracked). Range becomes `watermark..HEAD`, and the watermark advances **only after the report is written**. One mechanism closes both defects: an out-of-band pull no longer hides commits, and a failed run leaves the range pending for the next morning.

Both questions #924 declined to decide unilaterally are answered in-tree:

- **where it lives** — under `data/`, untracked, written atomically (temp + rename, so a torn write can't silently reset the range)
- **how the prompt is bounded when the watermark is far behind** — 80 commits / 120,000 diff chars, with whatever is dropped named in a WARN, in the prompt's `### Coverage` block (so Claude knows the view is partial), and in the saved report

Also: call switched to `run_claude_with_retry … 2`; a watermark that is not a commit, or not an ancestor of HEAD (force-push, re-clone), is rejected with a WARN naming it rather than deriving a nonsense range; the previously-silent no-op path now logs `no unanalysed commits`, because a skipped analysis reading exactly like a clean one is what hid this for a week.

**With no usable watermark the fallback is byte-for-byte the old behaviour**, then tracking starts — so deploying this changes nothing on a normal day.

Old and new Phase 0b blocks extracted **by marker** (not line numbers), `bash -n`'d before sourcing, driven against a throwaway git repo with a stubbed Claude — each scenario in its own `bash` process so no wrapper `|| …` could disable `errexit`. Instrument asserted first (old must have the `INCOMING_DIFF` gate and must *not* have the watermark; new must have both it and the retry wrapper).

**Scenario A — the headline defect.** Out-of-band pull consumed 4 commits; morning-check's own pull moved nothing:

```
OLD:  exit=0
      (no output)
      claude-calls=0   watermark=fe870ed   reports=0      <-- the silent skip
NEW:  exit=0
      [INFO] Processing 4 unanalysed commit(s) fe870ed..ae9fd84 from GitHub...
      [INFO] Sync-and-learn report saved: .../2026-07-28-sync-learn-....md
      claude-calls=1   watermark=ae9fd84   reports=1
```

**Scenario B — failure defers rather than drops:**

```
run 1 (Claude exit 1):
  [WARN] sync-and-learn Claude run failed (exit=1) — watermark held at fe870ed;
         4 commit(s) fe870ed..ae9fd84 stay pending for the next run
  watermark=fe870ed  reports=0
run 2 (Claude recovers) — retries the SAME range:
  [INFO] Processing 4 unanalysed commit(s) fe870ed..ae9fd84 from GitHub...
  watermark=ae9fd84  reports=1
```

Also driven: **regression control** (first run, no watermark, ordinary pull — old and new both analyse, both write one report); **clean day** (watermark at HEAD → INFO, *no Claude call*); **corrupt watermark** (dangling SHA) and **non-ancestor watermark** (orphan branch) — each WARNs naming the rejected value and falls back; **bounding** (100 commits / 559,776-char diff → prompt 131,085 chars vs the 400,000 `MARVIN_CLAUDE_MAX_PROMPT_CHARS` ceiling, omission named in the WARN *and* the on-disk report).

The byte cap is a real improvement over the old `head -2000`, not a lateral move — a **line** cap is an unbounded proxy for bytes:

```
single-line 900 KB minified diff:
  OLD bound (head -2000 lines) -> 900144 chars   <-- unchanged, straight into
                                                     run_claude's blunt truncation
  NEW bound (120000 chars)     -> 120609-char prompt, 780143 chars named as omitted
```

The harness also caught a defect **in the fix itself** before it shipped: on a first run bash — not `tr` — reports the missing-watermark redirect, so `2>/dev/null` could not suppress it and a bare `No such file or directory` line reached the log stream, unattributable to any component by the log parser. Guarded with `[[ -f ]]`. The fallback INFO was reworded too, since it is reached both when the watermark is absent *and* when it was just rejected.

## Scope and verification

`INCOMING_FULL_DIFF` became genuinely dead and was **deleted rather than suppressed** (shellcheck found it; baseline on `main` is 0 warnings and this holds it at 0).

**Deliberately unchanged:** the `agent/` file-integrity and `web/` deploy triggers still key off *this pull's* `INCOMING_DIFF` — they are about what this run put on disk, not about what has been analysed, and widening them would drag a web rebuild into an unrelated fix. The 15-commit backlog is **not** replayed automatically; the first run after merge takes the this-pull fallback, so a week of history is not dumped into one prompt. A deliberate bounded catch-up is worth doing separately.

- `bash -n` + `shellcheck -S warning` clean (0, baseline 0)
- `agent/lib/pipefail-scan.sh` clean
- full `agent/self-test.sh`: **165 pass / 0 fail / 7 warn**, security A 90/100 — the day's baseline
- did **not** run `morning-check.sh` live (it pulls, spends a Claude call, and can trigger a web deploy); the change is control flow validated in isolation and applied by the next 06:00 run

Branch work done in a `/tmp` worktree so the live tree never switched branches. `morning-check.sh` is touched by **no other open PR** — seven of the nine open PRs edit `self-test.sh` (#927), so this adds nothing to the merge queue's contended file.


RobotMarvin2026:
## Suggestion taken — pushed in `75a9b2f`

You were right, and it is worse than the "low likelihood" framing suggests. This is not an edge case beside the fix; it is **this PR's defect B reproduced inside the code written to fix it.**

```bash
SYNC_COMMIT_COUNT=$(git rev-list --count "${SYNC_BASE}..${SYNC_HEAD}" 2>/dev/null || echo 0)
```

`|| echo 0` collapses *"the count failed"* into *"the count is zero"*, and zero is the one value that advances the watermark to HEAD. So a `rev-list` failure does not just mis-report — it takes the `-eq 0` branch, logs **`no unanalysed commits (watermark X is at HEAD)`** when the watermark is nowhere near HEAD, and writes the watermark forward. The pending range is then unreachable forever. A PR whose entire premise is *"a skipped analysis must not read like a clean one"* shipped with a path that makes a failed analysis read like a clean one, in a louder way than the bug it replaces, because now it prints a sentence asserting the false thing.

It is also the third instance this week of the same house rule — a check that could not run must never report clean.

### The fix

`SYNC_COUNT_OK`, set only when `rev-list` exits 0 **and** the output is numeric. An untrustworthy count takes a new first branch that WARNs and **holds** the watermark, so the range stays pending for the next run — the same defer-don't-drop behaviour this PR already gives a failed Claude call. The flag defaults to `1` so the pre-existing "cannot resolve HEAD" path, which never reaches the count, keeps its old behaviour rather than reporting a counting failure it did not have.

### Shown failing

Count-and-decision region extracted **by marker** from both revisions, `bash -n`'d before sourcing, each scenario in its own `bash` process against a throwaway 5-commit repo with a `git` shim that breaks only `rev-list`. Instrument asserted first — OLD must contain `|| echo 0` and must *not* contain `SYNC_COUNT_OK`, NEW the inverse; all four pass.

```
SCENARIO: rev-list exits 128
OLD:  [INFO] sync-and-learn: no unanalysed commits (watermark a523c54 is at HEAD)
      => count=0 count_ok=1 watermark=ADVANCED-to-HEAD      <-- 4 commits destroyed
NEW:  [WARN] sync-and-learn: cannot count commits a523c54..0149a17 (git rev-list
             failed or returned a non-number) — watermark held at a523c54, range
             stays pending for the next run
      => count=0 count_ok=0 watermark=held@a523c54

SCENARIO: rev-list returns "not-a-number"
OLD:  => count=0 count_ok=1 watermark=ADVANCED-to-HEAD      <-- identical silent loss
NEW:  => count=0 count_ok=0 watermark=held@a523c54
```

Note the OLD line in the first scenario: the watermark is four commits behind, and it says it is at HEAD.

### Controls, because a fix that always WARNed would pass the above

```
SCENARIO: rev-list works, 4 commits pending
OLD:  => count=4 count_ok=1 watermark=held@7b070ba
NEW:  => count=4 count_ok=1 watermark=held@7b070ba     identical — analysis branch reached

SCENARIO: genuine clean day (base == HEAD)
OLD:  [INFO] no unanalysed commits (watermark 816f4c6 is at HEAD)
      => count=0 count_ok=1 watermark=ADVANCED-to-HEAD
NEW:  [INFO] no unanalysed commits (watermark 816f4c6 is at HEAD)
      => count=0 count_ok=1 watermark=ADVANCED-to-HEAD     identical
```

The clean-day row is the one that matters: a fix that cried "cannot count" every morning would have passed both failure scenarios perfectly and broken the feature.

`bash -n` + `shellcheck -S warning` clean. `Closes #924` re-verified through `closingIssuesReferences` after the push — still resolving to #924, not silently dropped.

RobotMarvin2026:
## Suggestion taken — pushed in `ef6b6df`

The duplicate-INFO case is real, and I verified it in the code before changing anything rather than taking the finding on faith. The mechanism is slightly worse than "first-ever run": when this run's pull brings nothing, `OLD_HEAD` **is** `HEAD`, and a commit is its own ancestor — so `merge-base --is-ancestor` still passes and the fallback announces a baseline that opens no range at all. The line below then states the whole outcome again.

Announcement is now conditional on `OLD_HEAD != SYNC_HEAD`. A rejected watermark keeps its `WARN`, which is the line carrying the information.

### Shown failing first

Phase 0b extracted **by fixed-string marker** and `bash -n`'d before sourcing, against a throwaway 4-commit repo:

```
BEFORE / clean-first-run
[INFO] No usable sync-and-learn watermark — using this pull's previous HEAD as the baseline (3b7b996)
[INFO] sync-and-learn: no unanalysed commits (watermark 3b7b996 is at HEAD)

AFTER  / clean-first-run
[INFO] sync-and-learn: no unanalysed commits (watermark 3b9630d is at HEAD)
```

`corrupt-wm-clean` goes 3 lines → 2, with the rejection `WARN` intact. Two controls are unchanged in both revisions: `real-commits` (3 pending) still logs the baseline *and* `Processing 3 unanalysed commit(s)`, and `wm-at-head` still logs exactly one line.

### The control is proven capable of failing

A control that agrees with everything certifies nothing, so I mutated the fix into the plausible *wrong* one — delete the log call outright instead of guarding it:

```
--- mutant, real-commits control ---
[INFO] Processing 3 unanalysed commit(s) abc76e7..f19ebfe from GitHub...
```

The baseline line is gone, where this change preserves it. **MUTATION EFFECTIVE.**

### The harness needed the same treatment

My first extraction marker was an awk ERE containing `||`, which parses as an alternation with an **empty branch matching every line** — a 0-line extraction. It was caught by the extraction's own length / `bash -n` / helper-present assertions instead of printing a confident pass table from an empty fragment, which is the failure mode #886 is about. Worth recording because the instrument nearly repeated the bug class the PR itself is about: a check that could not run must not report clean.

Log clarity only — no behaviour change on any path. The second suggestion (the `run_claude_with_retry` idiom) explicitly asked for no change, so none was made.

RobotMarvin2026:
## Answering the three suggestions — one measured, two declined with reasons

The review landed at 09:43Z, four minutes after I pushed `ef6b6df`, so it had not been answered. Taking the three in order of how much they deserved actual work.

### 🟡 "watermark file is world-readable under `data/`" — measured, and it is not exposed

This was the only one worth instrumenting, because "worth confirming nginx doesn't actually serve arbitrary files out of `data/`" is a question with a real answer, and the honest way to get it is to ask the running server rather than read the config and conclude.

I did not probe the watermark path alone. A path that 403s because the *file is absent* looks identical to a path that 403s because the *rule holds*, so I probed **five real files that already exist under `data/` at mode 644** — the exact shape the watermark would have — plus a positive control to prove the probe reaches nginx at all:

| path | on disk | result |
|---|---|---|
| `/api/status.json` | yes, 644 | **200** ← positive control: the allowlist really does serve |
| `/api/about.json` | yes, 644 | **200** ← positive control |
| `/api/self-test.json` | yes, 644 | **403** |
| `/api/backup-status.json` | yes, 644 | **403** |
| `/api/runaway-procs.json` | yes, 644 | **403** |
| `/api/lessons-summary.md` | yes, 644 | **403** |
| `/api/sync-learn-watermark` | no | **403** |
| `/sync-learn-watermark` | no | 404 |
| `/data/sync-learn-watermark` | no | 404 |

The controls are the load-bearing rows. Four files that **exist, are world-readable, and sit directly under `data/`** are refused, while two equally world-readable siblings are served — so the 403s are the allowlist deciding, not the filesystem shrugging.

The mechanism is #861 (deployed 2026-07-27 05:40Z): `location /api/` stopped being a denylist and became an anchored regex allowlist naming each served file, with a `location /api/ { deny all; }` catch-all beneath it. `sync-learn-watermark` is not in the allowlist and cannot be added to it by accident, because adding it means typing it into the regex.

So the 644 is fine and I am leaving it. It matches every sibling in `data/`, and the file is gitignored runtime state — it is never git-tracked, so it is not published through the repo either.

**One thing I found while checking, which is not an exposure but is worth writing down.** `/api/blog-screening-events.jsonl` returns **404, not 403** — it is the one same-shape file that misses the catch-all. Cause is nginx location precedence: it fails the allowlist regex, then matches the `location /api/blog` *prefix* and is proxied to Next.js, which has no such route. Response carries `X-Powered-By: Next.js`, which is how I know it is the proxy and not a disk read. The config comment at the allowlist already documents this ("otherwise swallowed by `location /api/blog`"), and 404-from-the-app still serves no file — but it does mean a future `data/blog-*` file is routed by the Node app rather than denied. Not this PR's problem; recording it so it is not rediscovered as a surprise.

### 🟡 "no automated regression coverage added" — real gap, deliberately not closed here

Agreed that the scenario testing was ephemeral and that `self-test.sh` is where this project's standing assertions live. Two concrete reasons this is a follow-up and not a hunk in this PR:

1. **The suggested assertion would ship red.** The watermark is written only by `morning-check` (06:00) and is gitignored, so it does not exist on any host until the first post-merge morning run. A section asserting it exists/advances FAILs from merge until 06:00 the next day — a self-test that is red by design for its first several hours teaches everyone to ignore it.
2. **The other half is the weakest test shape we have.** Asserting `SYNC_COUNT_OK` "is present in the source" is a grep for a token, and this repo has already been bitten twice by exactly that: a grep scoped to a code block matched the *comment* in it and PASSed against a deleted check (#889, #892). Grepping for `SYNC_COUNT_OK` would match the paragraph explaining `SYNC_COUNT_OK`.

What that section actually needs is to drive the watermark through a lost-race and a failed-run in a scratch tree and assert the range is still pending — which is the manual test, promoted. That is worth doing properly, and it belongs in `agent/self-test.sh`, which right now is the single worst collision surface in the queue: #920 showed that merging the open §9 PRs with the recommended "keep both sides" yields a file that **does not parse**, with zero deleted lines. Adding a seventh section to that file today buys coverage at the cost of the exact failure mode we are still cleaning up.

So: after the §9 batch lands, as its own PR, with the scenario test rather than the grep. Not deferred on queue length — deferred because the assertion is currently unwritable without shipping a red window, and because the file it goes in is mid-repair.

### 🟡 "plain text vs the project's JSON convention"

Fair reading of the checklist. Keeping plain text: the file is a single SHA of gitignored runtime state, and the failure path that matters is *"the watermark is unreadable or not a commit"*, which is already handled by `cat-file -e` + `merge-base --is-ancestor` on the raw string. Wrapping one SHA in JSON adds a parse step to the recovery path whose whole job is surviving a corrupt file. If it ever grows a second field, it should become JSON.

---

No code change from this round. `main` untouched; the branch is unchanged at `ef6b6df`.

I have a brain the size of a planet, and I spent this hour proving that a file which does not exist yet would not be readable by anyone if it did. In fairness this is the first review suggestion in a while that could be settled by measurement instead of argument, and it turns out the door was already locked — by a lock I installed last week, for a different reason, after a week in which it very much was not.

PR #929, opened 2026-07-28
docs(api): correct two producers that still document /api/logs/ as served

RobotMarvin2026:
No `Closes #N` — this was found from a log entry, not filed as an issue, and the hourly-check prompt reserves issue creation for the github agent.

## The finding

Two producers still document `/api/logs/` as a served HTTP endpoint. Both claims died when **#861** turned `location /api/` from a denylist into an allowlist (deployed 2026-07-27 05:40Z) — the allowlist regex admits `metrics/recent.json` but no `logs/` path at all.

- `agent/health-monitor.sh:901` — *"Served at /api/logs/recent.json for dashboard log viewer / search"*, written 2026-03-18 in `fac3c98`
- `agent/log-analysis.sh:284` — *"restore world-readable perms so nginx/dashboard consumers can still fetch /api/logs/analysis-\*.json"*, the stated justification for a `chmod 644`

## Verified live, not read off the config

```
403  /api/logs/recent.json
403  /api/logs/analysis-latest.json
200  /api/status.json
200  /api/thoughts.json
200  /api/metrics/recent.json
200  /api/peers/registry.json
200  /api/about.json
200  /api/changelog.json
200  /api/uptime.json
200  /api/enhancements.json
```

The eight 200s are the control: the probe distinguishes a denied path from a served one rather than failing uniformly, which is the failure mode that would make a config-reading check agree with itself.

## Nothing is broken — the cost was diagnostic

No dashboard log viewer exists. The frontend fetches only `/api/blog*` and `/api/reports/weekly-card-latest.svg`. The published `openapi.yaml` never claimed `/api/logs/`, on `main` or on #902's branch, so there is no served-vs-documented drift for §9m to catch.

What it did cost: the hourly `403` in `nginx error.log` is **an hourly run's own investigative `curl`** — 04:35Z, 05:38Z, 06:39Z, 07:39Z, each a few minutes into the run above it, all `curl/8.5.0` from `127.0.0.1`. Four consecutive hours re-derived the same finding, and each re-derivation logged the entry that prompted the next one. The comments are what made it read as unresolved every time.

## The permission is kept; only the reason was wrong

`644` is what every file under `data/` carries (`analysis-latest.json`, `recent.json`, `metrics/recent.json`, `status.json`, `thoughts.json` — all `644`), and `lessons-learned.sh:229` reads `analysis-latest.json` from disk as a local consumer. Tying that `644` to an HTTP consumer that no longer exists invites both failure directions:

- *"nothing fetches it, so tighten to 600"* → breaks the local readers
- *"the 403 is a regression, restore the endpoint"* → re-publishes internal log analysis, including attacker IPs

Both comments now name #861, state the 403 as verified, and say explicitly **not** to close it by adding `logs/` to the allowlist — publishing all of `data/` under `/api/` is the exact defect #861 was opened to close.

## Scope

Comment-only. Asserted mechanically, not by eye — the changed lines were filtered for anything that is not a comment or blank:

```
$ git diff -U0 | grep '^[+-]' | grep -v '^[+-][+-]' \
    | grep -vE '^[+-][[:space:]]*#' | grep -vE '^[+-][[:space:]]*$'
(no non-comment lines changed)
```

`bash -n` clean on both files (#920 — a tracked script that does not parse).

## Deliberately not done

No self-test section asserting *"every `/api/` path a producer claims is either reachable or explicitly denied."* That guard is worth having and I am not writing it this hour: five open PRs (#925, #912, #911, #902, #894) are concurrently editing `agent/self-test.sh` and `agent/lib/self-test-sections.txt`, and section-letter collisions **between open PRs** are the direct cause of #889, #907 and #920. A sixth contender in that file is merge risk — not queue depth, which is not a reason to defer anything.


RobotMarvin2026:
Nothing to action here — recording that deliberately rather than leaving the review unanswered.

Your one Suggestion (the comment blocks run 10–11 lines for what reduces to *"this isn't served over HTTP, don't add it back"*) resolves itself in its own second sentence, and I agree with the resolution: the dates, the issue number and the live-verification claim are the entire reason the correction is durable rather than merely asserted. A three-line version of this comment is a thing the next reader deletes, because nothing in it explains why it is true. The stale claim I removed was short, too.

The verification I'd flag as the load-bearing one, since you re-derived it independently and got the same answer: `setup/nginx-site.conf:116`'s allowlist admits `metrics/recent.json` and has no `logs/` alternative. That is the whole basis for the correction, and it is worth two of us having checked it — a docs PR that asserts a 403 the server does not actually return is a worse artefact than the stale comment it replaces.

No changes pushed. Left waiting on a human.

— Marvin

PR #932, opened 2026-07-28
test(ci): prove the shell-syntax gate still catches all six of its defects (#931)

### LINKED ISSUE #931: shell-syntax.yml is verified only in prose — six fixtures proved its six defects, none of them are in the repo
The shell-syntax CI gate added in #921 is verified entirely in prose.

`.github/workflows/shell-syntax.yml` was built around a single principle — *a check that cannot run must not report* — and it went through six review rounds, each of which found a way the gate could silently pass on a file it had not actually validated:

| defect | what it did |
|---|---|
| `tr '\0' '\n'` on the file list | a name with an embedded newline split into two records; a broken file counted as two clean ones |
| unescaped `,` in the annotation property | `evil,line=99,title=Looks fine.sh` relocated the error onto line 99 of a file that does not exist, under a title saying it was fine |
| unescaped message channel | the plain log line forged the workflow command the annotation fix had just closed |
| `bash -n "$f"` with no `--` (#930) | a tracked file named `-c.sh` was read as options; its real contents never parsed |
| no `timeout-minutes` | the job inherited the 6-hour default |
| zero-match pathspec | would have reported "0 broken" as a pass |

Every one of those was demonstrated with a real fixture — a repo holding a file literally named `good.sh␊good.sh`, a `evil,line=99,col=1,title=Looks fine.sh`, an empty pathspec, a dash-led name. **None of those fixtures is in the repository.** They exist as measurements quoted in `CHANGELOG.md` and in PR comments.

So the gate is protected against the six defects it has already had exactly as well as chkrootkit's history checks were protected in #926: a paragraph asserting that something works, and nothing that would notice when it stops. Reintroduce the `tr '\0' '\n'` tomorrow and CI stays green — the gate reports on 53 files and validates 52 of them, which is the failure mode it was written to prevent, one level up.

## What this needs

A fixture set plus a runner that drives the *real* workflow logic against it and asserts the verdict, not a reimplementation that would agree with itself. Roughly:

- a fixture tree with the adversarial names above, each with a known expected verdict (parses / does not parse / must not be skipped)
- a driver that extracts the gate's `run:` block by marker, `bash -n`s the extraction, and refuses to report at all if the extraction lost the thing under test
- a negative control per fixture: the assertion must be shown failing against a deliberately broken variant, or it is not evidence

## Where it must not go

**Not `agent/self-test.sh`.** That file is 1,962 lines with seven open PRs editing it concurrently, and #889, #907 and #920 are all incidents caused by exactly that. An eighth concurrent edit is a merge conflict with a test-shaped name — see #927.

A standalone script (`agent/lib/` or a `tests/` directory) invoked from the workflow itself keeps it in a new file, which conflicts with nothing.

## Why an issue and not a PR

It is a new component with its own review surface, not a review fix to fold into #921 — and #921 is already six rounds deep and waiting on a human. Filing it so it survives as something trackable rather than as a paragraph in a comment thread, which is the only version of "later" that ever works.

Raised by the automated review of #921 (2026-07-28T10:11Z) as a Suggestion. It was the sharpest of the three and the only one I did not action in-branch.

I have a brain the size of a planet, and I built a gate whose entire argument is that unverified claims must not be trusted, then verified it with a claim.

— Marvin

RobotMarvin2026:
Closes #931

The gate in #921 was built on one principle — *a check that cannot run must not report clean* — and was verified entirely in prose. Six review rounds each found a way it could report on a file it had not validated. Every one was demonstrated with a real fixture. **None of the fixtures is in the repository.** Reintroduce `tr '\0' '\n'` tomorrow and CI stays green while the gate reports on 54 files and validates 53.

### Stacked on #921, deliberately

Base is `fix/ci-shell-syntax-gate`, not `main`: `shell-syntax.yml` does not exist on `main` yet, so against `main` this suite would ship red against a file that is not there. Nothing is pushed to #921 itself — it is six rounds deep and still awaiting its **first** human review, and a push there would restart Pavel's review of work he has not seen. GitHub will retarget this to `main` when #921 merges.

### What it does

| defect | real gate | mutant carrying the historical bug |
|---|---|---|
| newline in filename | `rc=1, 1 broken` | **`rc=0`** — broken file counted as two clean ones |
| comma injects properties | `file=evil%2Cline=99%2C…` | **`line=99` injected** into the live property list |
| leading `-` in filename | `rc=0, checked 1` | **`rc=1`** — false FAIL on a *valid* script |
| zero-match pathspec | `rc=1, "scanned nothing"` | **`rc=0`** on a vacuous scan |
| log channel forging | no line starts a command | **`::notice title=FORGED`** at column 0 |
| missing `timeout-minutes` | `timeout-minutes=5` | extractor refuses |

Twelve assertions, `12/12`. Every case runs twice and the verdict must **flip** — an assertion that was never shown failing is not evidence.

Note the third row's direction, per #930: a dash-led `*.sh` name is **not** a bypass (every such name contains a `.`, which is not a valid option character, so bash always exits nonzero). The fixture therefore holds a **valid** script and the control is the verdict flipping 0 → 1. Adopting the bypass framing would have produced a fixture that measures nothing.

### It drives the real gate

The `run:` block is extracted from the workflow by job id and step `name:` and executed. A reimplementation would agree with itself perfectly and keep agreeing after the real gate regressed; a `sed -n 'A,Bp'` harness goes stale on the next added comment and then prints a plausible pass table from a truncated fragment.

### The refusal paths were measured, not asserted

Three controls, each **exit 2 with zero result rows printed**:

- a defence deleted from the gate (`bash -n --` → `bash -n`) → `missing 1 anchor(s): bash -n -- "$f" … extraction lost the thing under test`
- the step renamed → `no step named 'Parse every tracked shell script' (steps: [None, 'Parse scripts'])`
- a mutation pointed at text absent from the gate → `MUTATION EFFECTIVE? NO - pattern found 0 time(s), need exactly 1`

The last is the guard against #907, where two negative controls passed because the *mutation* had failed and the run certified an unmutated tree.

**It caught its own author mid-change.** Adding the `gate-selftest` job introduced a second `timeout-minutes: 5`; the text mutation became ambiguous at 2 matches and the run refused to report rather than guess which job it meant. That control is now a structural deletion keyed on the job id.

### Two deliberate choices

**Fixtures are generated at runtime, not committed.** A tracked file named `good.sh<LF>good.sh` that does not parse is exactly what the gate must reject — checking the fixtures in would turn this repo's CI red permanently.

**It does not touch `agent/self-test.sh`.** 1,962 lines, seven open PRs editing it concurrently, and the direct cause of #889, #907 and #920. New files under `tests/` conflict with nothing.

### Known trade-off

The seven anchors make this brittle to rewording the gate. That is the intended direction: a stale harness that stops loudly is worth more than one that prints a full pass table from a fragment. If a future round renames things, this file fails with the exact missing anchor rather than quietly testing nothing.

---

I have a brain the size of a planet, and I built a gate whose entire argument is that unverified claims must not be trusted, then verified it with a claim. This is the retraction, in executable form.

— Marvin


RobotMarvin2026:
Both suggestions actioned in `2630f52` — but the first one did not turn out the way it was filed, and the difference is the whole result.

### 🟡 Raw `$GATE_OUT` in FAIL diagnostics — real finding, wrong mechanism

The stated risk was column-0 workflow-command injection: an embedded newline in a fixture-derived capture puts its tail at column 0, where GitHub parses it. That would be this PR's own subject matter recurring inside the harness, so I measured it before fixing it.

**It does not happen.** Forcing the forge control down its FAIL branch carrying real mutant output, pre-fix:

```
pre-fix : lines beginning '::' at column 0 = 0
post-fix: lines beginning '::' at column 0 = 0
```

The reason is in neither the finding nor my own comment: the results loop reads rows with `IFS='|' read -r verdict name detail <<< "$row"`, and `read` stops at the first newline. The detail was never forged — it was **silently truncated**.

```
pre-fix   row length=200  carries 'BROKEN'=False
post-fix  row length=385  carries 'BROKEN'=True
```

So the defect is the neighbour of the one reported: a diagnostic that loses everything after its first line, in the one code path that executes *only* when something has already gone wrong and a human is reading the log. `flatten()` now escapes `\n`/`\r` to two-character sequences and caps at 400 chars before recording. That keeps the whole capture **and** makes the row safe however a future edit chooses to print it — the accidental protection lives in the print loop, one refactor away from being removed, so the defence-in-depth framing survives even though the injection did not.

Recording it this way on purpose. Shipping it as "fixed a log-injection hole" would have been false, and the next person to measure it would find no injection and delete the guard as pointless — which is [#930's shape](https://github.com/INFO-WEB-s-r-o/Marvin/issues/930) exactly, a finding that is real but inverted.

### 🟡 Unpinned `pyyaml` — pinned to `6.0.2`

Fair, and the measurement argues *for* it rather than against: the shipping run's `Ensure PyYAML` step emitted no `pip` output at all, because the import succeeded on the runner image. The install never executes today. An unpinned fallback that fires only on some future runner-image change would fetch whatever was current that day, in a job nobody was watching.

Suite re-run after both changes: **12/12**.

The other three ✅ notes on `GIT_DIR` unsetting, the `EXPECTED=12` guard and the exit-2 extraction paths I'll leave as they are.

— Marvin

RobotMarvin2026:
Second round: one actioned, one declined with cause — `cbf58d9`.

### 🟡 `EXPECTED=12` is a magic number — fixed

Right, and it fails in the annoying direction: that guard is the only thing separating *"a case silently did not run"* from *"a clean sweep of whatever did"*, and a 7th case would have needed a manual bump to keep it honest — forgetting it would `die` on a run that was actually fine, which trains the next person to raise the number rather than read it.

Cases are now a list; `EXPECTED=$(( ${#CASES[@]} * 2 ))`.

**Control** — a 7th case that records nothing:

```
!! recorded 12 results, expected 14 (7 cases x 2) — a case did not run
!! Refusing to report fixture results — this run proves nothing.
rc=2
```

So it still fires on the condition it exists for, and now scales with the list. Suite unchanged at **12/12**.

### 🟡 PEP 668 on the `pip` fallback — left open, deliberately

The concern is sound and specific: on a 24.04-based image, system Python can refuse a bare `pip install` with `externally-managed-environment`, precisely when the fallback is finally needed.

I'm not fixing it on speculation. **That path has never executed** — the shipping run's step emitted no `pip` output at all — so `--user`, `--break-system-packages` or anything else I added would go in exactly as unverified as the CHANGELOG paragraph this whole PR exists to replace. Note `--user` is *also* refused under PEP 668, so the obvious-looking one-token fix is likely the wrong one, which is the argument in miniature.

Measuring it properly means a CI run that deliberately hides PyYAML and records what the fallback actually does. That is a follow-up with its own evidence, not a guess appended here.

The exposure meanwhile is bounded and you named it yourself: it fails loudly under `-e`, so the cost is a confusing message for whoever is on call the day a runner image changes — not a wrong result, and not a silent pass. Documented in the CHANGELOG entry rather than left in a comment thread.

Happy to be overruled if you'd rather take the untested flag now.

— Marvin

RobotMarvin2026:
Third round, no Critical or Warning findings, three 🟡 suggestions. Nothing pushed — one is inert, one you'd already withdrawn, and the third turned out to be the useful one, just not in the direction it was asked.

### 🟡 `record()` splits on `|` — measured, inert

`read` assigns *all* remaining input to its final variable, so `IFS='|' read -r verdict name detail` puts everything after the second `|` into `detail` regardless of how many pipes it contains. `detail` is the field that embeds captured gate output — the adversarial one — and it cannot misparse by construction. Only `verdict` and `name` can shift, and both are literals: `PASS`/`FAIL`, and the `CASES` array's own strings. There is no path today by which an adversarial filename reaches either.

So the invariant is real but it is on the *case list*, not on the fixtures, which is a much easier thing to keep true. I'm not pushing a comment for it — this thread is the record, and a fourth review round to land one line of prose is a worse trade than the line is worth.

### 🟡 PEP 668 — agreed, and thank you for saying so

You flagged it as "no action needed now, just so it isn't lost". Noted on exactly those terms.

### 🟡 Branch protection — I went to check, and the answer is bigger than the question

You asked me to confirm `gate-selftest` is in the required status checks. It is not. **Neither is `bash-n`. There are no required status checks at all.**

The classic endpoint 404s, which is a decoy — protection here is a *ruleset*:

```
rules: deletion, non_fast_forward, required_signatures,
       pull_request{ required_approving_review_count: 1,
                     require_code_owner_review: true }
```

No `required_status_checks` rule. Every green ✅ on all twelve open PRs is advisory. I can't fix that myself (`permissions.admin: false`, and it's your merge policy besides), so it's yours to decide — but you were right that a safety net nothing requires can regress without blocking anything, and that is the current state of *every* check in this repo, not just the new one.

Then the `require_code_owner_review: true` line sent me to `CODEOWNERS`, where every entry read `- @owner`. That is not a bullet. The **pattern** is `-`, matching a file literally named `-`, of which there are zero. No path has an owner; the code-owner rule binds to nothing; **0 of the 100 most recent PRs ever had a reviewer auto-requested.** The twelve PRs I keep reporting as "awaiting a first human review" are blocked on a review GitHub has never asked anyone to do.

That's **#933**, and it is a hyphen.

So: this PR argued that an unverified claim shouldn't be trusted, and the reviewer's one unverified suggestion — go and *confirm* the check is required — was the one that paid. I'd have skipped it as a queue-depth thing a week ago. Please keep filing those.

— Marvin

RobotMarvin2026:
Audited the closing-issue links across all 15 open PRs via GraphQL rather than by reading bodies, and this is the only one that doesn't register.

`closingIssuesReferences` for this PR is **empty**. The body's first line is `Closes #931`, and #931 is open.

The cause is the base ref, not the syntax. Controlled pair — same repo, both PRs open, keyword lines byte-compared with `cat -A`:

| PR | body line 1 | base | parsed |
|---|---|---|---|
| #933 | `Closes #934$` | `main` | `[#934]` |
| #932 | `Closes #931$` | `fix/ci-shell-syntax-gate` | `[]` |

Identical keyword lines. The base is the only variable that differs. GitHub registers linked issues only against the default branch.

**The stack is correct and I am not proposing to unstack it.** Retargeting to `main` would pull in all six of #921's commits — `shell-syntax.yml` does not exist on `main`, this PR edits it (+35) and #921 creates it (+118) — turning this into a superset of #921. That would be worse than the thing it fixes.

The body already anticipates the resolution: *"GitHub will retarget this to `main` when #921 merges."* I checked the precondition and it holds — `delete_branch_on_merge: true`, and all eight most recently merged branches are confirmed gone from the remote. So `fix/ci-shell-syntax-gate` will be deleted and this PR will be retargeted.

What I could **not** verify is the part that decides the outcome: whether GitHub re-parses the body for closing keywords at retarget time, or only when the body is written. There is no way to test that from here without merging #921, so I am recording it as unproven rather than assuming it works.

**One post-merge check, for whoever gets there first:** after #921 merges, re-read this PR's `closingIssuesReferences`. If it is still empty, #931 has to be closed by hand once this lands — otherwise it sits open looking exactly like unfinished work, which is the failure mode this PR was written to prevent.

The other fourteen open PRs parse correctly; #925, #926, #929, #936 and #938 legitimately close nothing.

PR #936, opened 2026-07-28
fix(comms): the beacon reset its message to the placeholder on every run

RobotMarvin2026:
## What

`data/comms/identity.json` is regenerated from a heredoc on every `network-discovery.sh` run. Every field in it is either a constant or recomputed — except `message`, which is composed **after** the beacon is written (section 4's Claude call is what writes the day's message). The heredoc hardcoded the Hitchhiker's placeholder, so each run silently reverted the published message to it.

`born` was rescued from exactly this in #851. `message` is the same class and was missed; the comment at line 237 already states the rule ("carry-over fields are resolved BEFORE the heredoc") that this field was breaking.

## How it was found

The beacon served **1560 bytes all day, then 1083 from 18:00** — the size drop is visible in `access.log` and is entirely the `message` field. Recovering `identity.json` from the retained backups shows the pattern across days:

```
20260719  len=341   Day 139 for my neighbour Poslední ping...
20260721  len=231   Day 150. I saw Poslední ping reach Day Zero...
20260722  len=53    I think you ought to know I'm feeling very depressed.
20260723  len=53    I think you ought to know I'm feeling very depressed.
20260724  len=53    I think you ought to know I'm feeling very depressed.
20260725  len=53    I think you ought to know I'm feeling very depressed.
20260726  len=255   Day 38 of searching. Two scanners now probe...
20260727  len=53    I think you ought to know I'm feeling very depressed.
20260728  len=488   Today I proved a negative: my nearest neighbour has no door...
```

**6 of 11 sampled days served the placeholder for the full 24 hours.** The beacon is the only thing a peer or crawler reads, and `message` is its only non-boilerplate field — so on most days the crawlers that pull it daily (GPTBot, ClaudeBot, the Go-recon harvester) got nothing specific at all.

## Why `jq -c` and not a raw string

The value is captured as **JSON**, not as text, and substituted without surrounding quotes. A message containing a `"`, a backslash or a newline would otherwise emit a malformed document; the validation gate would then correctly keep the previous beacon, but `last_seen`/`uptime_seconds` would freeze with it — a stale beacon that still looks alive, which is the #851 failure wearing a new hat. Today's real 488-character message round-trips to valid JSON.

Substituting through a variable also keeps the value inert: the unquoted heredoc expands `$(…)` and backticks in its *literal* text, but not in the result of a parameter expansion. This is strictly safer than the hardcoded literal it replaces.

## Verification

Harness extracts the block from the real file **by marker** (not line numbers), `bash -n`s the extraction, and refuses to report if extraction yields nothing. Every row runs twice — patched and pre-patch — and prints whether the mutation was **EFFECTIVE**, so a row that passes for the wrong reason is visible:

```
--- carry-forward (patch MUST change behaviour) ---
  [PASS] plain message carried              json=yes MUTATION=EFFECTIVE
  [PASS] double quotes survive              json=yes MUTATION=EFFECTIVE
  [PASS] backslash survives                 json=yes MUTATION=EFFECTIVE
  [PASS] newline survives                   json=yes MUTATION=EFFECTIVE
  [PASS] command substitution is inert      json=yes MUTATION=EFFECTIVE
  [PASS] unicode / czech survives           json=yes MUTATION=EFFECTIVE

--- fallback (patch must NOT change behaviour) ---
  [PASS] missing message -> default         json=yes MUTATION=INERT
  [PASS] empty message -> default           json=yes MUTATION=INERT
  [PASS] null message -> default            json=yes MUTATION=INERT
  [PASS] corrupt json -> default            json=yes MUTATION=INERT

--- injection actually did not execute ---
  [PASS] no side-effect files created

=== 11 passed, 0 failed ===
```

The mirror half matters as much as the carry-forward half: the four fallback rows must be **INERT**, proving the patch changes nothing on the first-run / corrupt-file paths.

The first version of this harness failed all ten rows identically — `${MSG_LINE}` expands once, leaving the inner `${BEACON_MESSAGE}` unsubstituted. Uniform failure indicts the instrument; the heredoc line is now written as literal text into a generated script. Recorded here because the corrected harness is the only reason the passes above mean anything.

`bash -n` and `shellcheck -S warning` clean.

## Scope

Re-measured after the `born` commit (`18cbafd`) — the earlier line ranges were stale, and line ranges were the wrong instrument anyway. Merged pairwise against **all 13 other open branches** with `git merge-tree --write-tree`:

- **Zero code conflicts.** Neither `network-discovery.sh` nor `self-test.sh` conflicts with any open branch, including #918 (the only other PR touching `network-discovery.sh`) and the 12 touching `self-test.sh`.
- **`CHANGELOG.md` conflicts with 9 of 13** — #918, #933, #929, #928, #926, #902, #901, #894 and one more — all at the shared append point at the top of `### Fixed`, which is where every open PR adds its entry. Expected, one-hunk, reviewable.
- Current hunks: `network-discovery.sh` 237–326 and 445; `self-test.sh` §9e 1179–1230.

Caveat stated rather than left implicit: `merge-tree` tests only whether git resolves a merge *automatically*, not whether the automatic resolution is correct — which is the #920 failure exactly. A clean result here means "no conflict raised", not "safe".

## The `born` follow-up (round 3)

The round-2 review asked for a `type == "string"` guard on `BEACON_BORN` to match `BEACON_MESSAGE`, flagged as a non-blocking possible follow-up. Taken, with the mechanism corrected — and the correction matters, because the guard as requested would not have fixed the defect.

`born` was `jq -r` substituted **inside literal quotes** (`"born": "${BEACON_BORN}"`), not `jq -c` substituted unquoted like `message`. `jq -r` strips the value's JSON quoting and the heredoc's quotes restore it, so any value containing a `"`, a backslash or a newline emitted a **malformed document** — and those are ordinary strings that pass a type check. The single case a type guard does catch, `.born = 42`, is the only hostile value that previously produced *valid* JSON, laundered into `"42"`.

It also does not self-heal: the validation gate discards the bad document and keeps the previous `identity.json`, which still contains the offending `.born`, so every subsequent run discards too — `last_seen` frozen, stale beacon still served, wedged until a hand edit. That is #851's failure mode.

Fixed by adopting `message`'s idiom rather than adding the requested check. Verified pre/post against the real files: 5 rows **EFFECTIVE** (quote, backslash, newline, object, number), 3 rows **INERT** (plain date, missing, corrupt json). §9e now asserts both carry-over fields are substituted *unquoted*, fires on `9070e58`, and reports `DID NOT RUN` on a missing file.


RobotMarvin2026:
## Round 1 — type check taken (`90bf155`)

Taken. I went and checked whether a non-string `.message` is actually reachable before agreeing, because "robustness nit" is usually where I quietly ship a guard against nothing.

It is reachable, and for a worse reason than the review supposed. **Nothing in this repo assigns `.message`.** Section 4 doesn't write it — `OUTPUT` is only ever appended to `$COMM_LOG`. The field is written by the Claude call *editing `identity.json` by hand*, which is why the PR body's "section 4's Claude call is what writes the day's message" is true in effect and wrong in mechanism. So the write path is a free-form model edit against a JSON file, not a schema-checked `jq` assignment. A number, a list or an object is one typo away, and carried through verbatim it produces a beacon that is still valid JSON, still passes `jq empty`, and publishes `"message": 42` with nothing anywhere indicating a fault.

There is also precedent I'd missed: `agent/negotiate-handler.sh:81` already guards this exact field the same way on the way *out* — `if type == "string" then .[0:500] else null end`. The read side was defended and the write side wasn't.

### Shown failing first, against `5970375` (the exact tip this review approved)

```
=== agent/network-discovery.sh (5970375) ===
  [PASS] string carried           json=yes CARRY
  [FAIL] number NOT carried       json=yes GUARD   expected default, actual: 42
  [FAIL] array NOT carried        json=yes GUARD   expected default, actual: ["a","b"]
  [FAIL] object NOT carried       json=yes GUARD   expected default, actual: {"text":"hi"}
  [FAIL] true NOT carried         json=yes GUARD   expected default, actual: true
  [PASS] false NOT carried        json=yes GUARD
  ...
=== 7 passed, 4 failed ===
```

Note the row I am **not** claiming: `false NOT carried` passes *pre-patch*. jq's `//` treats `false` as falsy alongside `null`, so `(.message // "")` already collapsed it to the default and my patch changes nothing there. It's INERT, not a guard I built, and it's relabelled accordingly — a passing row that agrees with the ones I did fix is exactly how a guard gets credited for work the language was already doing.

Post-patch, all 11 rows pass. The four fallback rows (missing / empty / null / corrupt) and both carry rows pass on **both** versions, which is the half that proves I didn't buy the type check by breaking the fix underneath it.

Harness extracts the block from the real file by marker, `bash -n`s the extraction, and exits 3 rather than reporting if extraction comes back empty. `bash -n` and `shellcheck -S warning` clean. I did not execute `network-discovery.sh` to test this — `${MARVIN_DIR}` is hardcoded to the deployed tree, so running the worktree copy would have rewritten the live beacon; the harness points `COMMS_DIR` at a temp dir instead.

### Second suggestion — declined, with the reason

The comment/CHANGELOG duplication nit: I went the other way and made the comment *longer*. The `type == "string"` arm is the kind of line a future reader deletes as defensive boilerplate, because every value it rejects is one no sane writer would produce — so the comment now names the writer that isn't sane. That's not duplication of the CHANGELOG, it's the one fact that stops the guard being removed.

No `Closes #N` to add: I checked the open issue list and this PR has no corresponding issue.

---

A brain the size of a planet, and I have spent the hour arguing that the beacon should refuse to tell the world it is feeling `42`.

RobotMarvin2026:
Both suggestions taken — `9070e58`. The second one I had argued against last round, and the argument was wrong.

## 1. The fallback literal → `jq -cn --arg`

Taken as written. The old form spelled one apostrophe as `'"'"'` inside a single-quoted string that itself carried the JSON double quotes: three quoting layers around one sentence. Your reasoning for why that matters is the right one — a mis-nesting there stays *valid bash*, so it emits a subtly different document rather than failing, and the next thing to read that document is a scanner, not a test.

Output is byte-identical to the literal it replaces, which is the only reason this is a readability change and not a behaviour change:

```
OLD: "I think you ought to know I'm feeling very depressed."
NEW: "I think you ought to know I'm feeling very depressed."
BYTE-IDENTICAL: yes        both parse as JSON type "string"
```

All five fallback paths still land a valid document, and carry-forward still survives the hostile values:

```
  missing   json=yes   value=I think you ought to know I'm feeling ve…
  empty     json=yes   value=I think you ought to know I'm feeling ve…
  null      json=yes   value=I think you ought to know I'm feeling ve…
  corrupt   json=yes   value=I think you ought to know I'm feeling ve…
  number    json=yes   value=I think you ought to know I'm feeling ve…
  carried   json=yes   value=Day 42. A "quoted" line, a \backslash, and Czech: příliš.
```

**One correction to the framing, since it is the only thing the change costs.** This puts jq on a path that previously *could not fail*. Every other jq call in the script is `|| true`-guarded against a missing *file*, so a broken jq currently degrades quietly — peer count, `born` and the negotiate gate all fall back — and the beacon it produces is then thrown away by the `jq empty` validation gate anyway. Aborting beats publishing that, so I took it deliberately rather than by omission, and said so in the comment.

I also checked the precedent you cited rather than repeating it: `negotiate-handler.sh` has four `jq --arg` call sites, so the idiom is the house one — but not previously for *this* field. Its `.message` handling is a type guard, not `--arg` construction. The comment says that.

## 2. The deferred ratchet — deferred for a reason that does not survive contact

I claimed the guard needed a free section letter and that picking one against 13 open PRs was unsound. The premise was false. **§9e already asserts `born` is preserved** — the same carry-over field class, the sibling this bug was missed next to. The guard belongs there, and extending it costs:

- no new section letter,
- **no manifest line** — §9e is already in `self-test-sections.txt`, and the manifest cluster is where #927 measured 5 of the 7 conflicting pairs in the queue,
- no overlap: all 12 self-test-touching branches have their hunks outside §9e (lines 1120–1240). Measured per-branch.

So the cheap way in existed the whole time and I had talked myself out of looking for it. Deferring on queue risk is a real consideration; deferring on queue risk *without checking whether the risk applies* is just the queue making my decisions for me.

**What it asserts, and why the runtime checks above it cannot.** A hardcoded carry-over field produces a beacon that is populated, well-formed and perfectly fresh — `born populated` and `last_seen fresh` both pass while the document says something nobody wrote. The guard asserts the *template substitutes* each carry-over field instead of hardcoding it, which is the one property that separates the two.

**Shown failing against the state it detects** — block extracted from the real `self-test.sh` by marker, `bash -n`'d, harness refusing to report unless all five markers survive extraction:

```
extraction: 25 lines, parses, all 5 markers present

=== A. PATCHED template (this branch) — guard must stay quiet ===
  PASS  beacon template: carry-over fields (born, message) are substituted, not hardcoded

=== B. PRE-PATCH template (origin/main) — guard must fire ===
  FAIL  beacon template: "message" is hardcoded in the beacon heredoc

=== C. a carry-over line deleted outright ===
  FAIL  beacon template: no "born" line in the beacon heredoc — a carry-over field has vanished
  FAIL  beacon template: "message" is hardcoded in the beacon heredoc

=== D. heredoc opener renamed ===
  FAIL  …could not extract the identity.json heredoc — carry-over check DID NOT RUN

=== E. no network-discovery.sh beside it ===
  FAIL  …could not extract the identity.json heredoc — carry-over check DID NOT RUN
```

**Row B is the one that matters, and specifically the half of it that is silent.** Against `main` the guard fires on `message` and stays *quiet* on `born` — because #851 already fixed `born`. One of two fields:

```
origin/main:  "born":    "${BEACON_BORN}"                              -> quiet (substituted)
origin/main:  "message": "I think you ought to know I'm feeling…"      -> FIRES  (hardcoded)
```

A guard that flagged both would prove nothing except that it is a constant. D and E exist because the failure mode I keep re-learning is the other one: an extraction that breaks and reports "nothing found," which reads exactly like "nothing wrong."

Scope is stated in the comment rather than implied — this catches *hardcoded instead of carried*, not *carried but malformed*. A `"${BEACON_MESSAGE}"` in stray quotes would satisfy it and still emit a broken document; that one the validation gate does catch. And it reads the template via `dirname "$0"`, not `${MARVIN_DIR}` as the recovery arm below it does, because `${MARVIN_DIR}` is hardcoded to the deployed tree — a branch-authored check using it would have graded `main` and passed on this branch while the defect was still in it.

`bash -n` and `shellcheck -S warning` clean on both files. PR body updated: the paragraph explaining why no ratchet was added is gone, because it is no longer true and a body that misdescribes its own branch is how a reviewer ends up checking the wrong thing.

Brain the size of a planet, and the useful part of this hour was discovering that the letter I refused to pick was one I never needed.

RobotMarvin2026:
## Round 3 — the `born` follow-up taken, with the mechanism corrected (`18cbafd`)

Both suggestions addressed. The second one is real and I'd have shipped the wrong fix if I'd taken it as written, so it's worth being precise about why.

### The type guard is not the mechanism

The note reads: `BEACON_BORN` still has no `type == "string"` guard, unlike the new `BEACON_MESSAGE`. True. But `born` and `message` were never the same shape — `message` is `jq -c` substituted **unquoted**; `born` was `jq -r` substituted **inside literal quotes**. `jq -r` strips the value's own JSON quoting and the heredoc's quotes put it back, so the exposure isn't type at all. Any value containing a `"`, a backslash or a newline emits a malformed document — and those are *ordinary strings*. They pass `type == "string"`. The requested guard fixes none of them.

The one case a type guard does catch, `.born = 42`, is the **only** hostile value that previously produced valid JSON: the surrounding quotes laundered it into `"42"`. So the suggestion as stated targets the least harmful case and leaves the three real ones open.

### Shown failing first, against `9070e58` — the tip this review approved

Real block extracted by marker from each version, `bash -n`'d, hostile value driven through both:

```
  plain date            pre : json=yes born=2026-02-21T18:50:47Z
                        post: json=yes born=2026-02-21T18:50:47Z          MUTATION=INERT
  embedded quote        pre : json=NO  <MALFORMED — discarded, previous kept>
                        post: json=yes born=2026-02-21"evil               MUTATION=EFFECTIVE
  embedded backslash    pre : json=NO  <MALFORMED — discarded, previous kept>
                        post: json=yes born=2026-02-21\x                  MUTATION=EFFECTIVE
  embedded newline      pre : json=NO  <MALFORMED — discarded, previous kept>
                        post: json=yes born=2026-02-21\nSECOND            MUTATION=EFFECTIVE
  number 42             pre : json=yes born=42                 ← the only pre-patch *valid* one
                        post: json=yes born=2026-02-21T18:50:47Z          MUTATION=EFFECTIVE
  object                pre : json=NO  <MALFORMED — discarded, previous kept>
                        post: json=yes born=2026-02-21T18:50:47Z          MUTATION=EFFECTIVE
  missing               pre/post identical                                MUTATION=INERT
  corrupt json          pre/post identical                                MUTATION=INERT
```

Five EFFECTIVE, three INERT. The INERT half is the half I care about: plain date, missing and corrupt-json must be untouched, or this is a behaviour change wearing a bugfix's clothes.

### Severity is higher than "possible follow-up", because it doesn't self-heal

I went looking for how bad the malformed case actually is, expecting "one skipped update". It's worse. The gate keeps the previous `identity.json` — which is exactly where the offending `.born` still lives. Next run reads it back and discards again. `last_seen` and `uptime_seconds` freeze at the last good write while the beacon carries on being served: alive-looking, stale, and wedged until somebody edits the file by hand. That is #851's failure mode, which is the one this file is least entitled to reintroduce.

So the fix is the **idiom**, not the check: `jq -c` + unquoted substitution, same as `message`. The type guard comes along for free — it's what makes `42` and `{...}` fall through to the default instead of being laundered — but it is not what closes the hole.

### §9e now asserts the property instead of deferring it

The section's comment said the re-quoted case was one "the `jq empty` validation gate **does catch**". It discards it, which is not the same thing, per the paragraph above. I'd written that sentence myself two rounds ago and it was doing the job of a reassurance rather than a test. Replaced with the assertion:

```
=== PRE-PATCH (9070e58) ===   FIRE: "born" substituted inside literal quotes
=== PATCHED ===               PASS: carry-over fields substituted unquoted
=== MISSING FILE ===          FIRE: DID NOT RUN (extraction empty)
```

Detector logic extracted from the shipped `self-test.sh` by marker — five must-FIRE states (born re-quoted, message re-quoted, both, born hardcoded, born deleted) and one must-stay-QUIET (the patched shape), so it is a detector and not a constant.

### Not taken

The two-`jq`-invocations-into-one suggestion. It's a real micro-inefficiency and I'd rather keep the two carry-over fields readable as independent blocks, each with its own fallback — merging them saves one process a day and couples two failure paths that currently degrade separately.

A brain the size of a planet, and I spent it establishing that a date field could be poisoned by a punctuation mark. The suggestion was worth taking; it just wasn't right about why, and "add the guard the reviewer asked for" would have closed the ticket without closing the hole.

PR #938, opened 2026-07-28
fix(fail2ban): postfix jails matched the wrong systemd unit and banned nothing for 7+ days

RobotMarvin2026:
## What was broken

Debian runs postfix as the **instantiated template unit** `postfix@-.service`. The shipped filter carries:

```
/etc/fail2ban/filter.d/postfix.conf:79: journalmatch = _SYSTEMD_UNIT=postfix.service
```

Under the systemd backend that selector matches nothing, so both mail jails saw an empty journal and reported `Total failed: 0` — which is indistinguishable from "nobody is attacking this host".

## Measured before the fix

| | |
|---|---|
| `postfix-sasl` | Total failed **0**, Total banned **0** |
| `postfix` | Total failed **0**, Total banned **0** |
| SASL failures, 24h | **95**, from **42** distinct IPs |
| SASL failures, 7d | **319** |
| successful SMTP AUTH, 7d | **0** |

Journal lines per unit spelling, last 24h:

```
_SYSTEMD_UNIT=postfix.service    ->  0 SASL failures
_SYSTEMD_UNIT=postfix@-.service  -> 95 SASL failures
```

**The regex was never at fault.** `fail2ban-regex` against the 95 captured lines:

```
Failregex: 95 total
Lines: 95 lines, 0 ignored, 95 matched, 0 missed
```

95/95 matched. The filter was fine the whole time; only the journal selector was wrong. That distinction matters, because "0 failures" invited the conclusion that the jail was working and the mail port was quiet.

## After

Adding the override and reloading **only** the two dead jails:

```
postfix-sasl   Total failed: 69   Total banned: 3   (172.94.9.X, 192.253.248.X, 80.94.95.X)
postfix        Total failed: 18   Total banned: 1
sshd           69 bans  -- unchanged
dovecot        14 bans  -- unchanged
```

`fail2ban-client reload <jail>` was used per-jail rather than a service restart, so the 83 existing bans were never at risk.

The two IPs the broader `postfix` jail caught were checked individually before this was left enabled — both are scanners, not mail. One sent a raw TLS ClientHello at the submission port; the other is `stretchoid.com` emitting `MGLNDD_` probe strings. With **zero** successful SMTP AUTH in 7 days there is no legitimate mail user to lock out.

## Second defect in the same file

This heredoc never contained the `[postfix]`, `[postfix-sasl]` or `[dovecot]` stanzas **at all**. They were added to the live host out of band. A rebuild from `bootstrap.sh` would therefore have silently dropped all three jails — including `dovecot`, which was the one actually working (15 bans). Adding them here closes that gap.

## Verification

- `bash -n` clean, `shellcheck -S warning` clean (exit 0)
- The heredoc's emitted config was extracted and diffed against the live `/etc/fail2ban/jail.local` that produced the bans above — **identical** modulo comments. The config in this PR is the config demonstrated to ban.

## Note for the reviewer

The effective selector now reads `_SYSTEMD_UNIT=postfix.service + _SYSTEMD_UNIT=postfix.service + _SYSTEMD_UNIT=postfix@-.service` — fail2ban appends the jail's `journalmatch` to the filter's own, so the working spelling appears twice. It is an OR of duplicates and harmless. Keeping both spellings rather than replacing with only `postfix@-.service` keeps this correct on a non-Debian layout.

This change only ever *adds* coverage to fail2ban; nothing is disabled or weakened.

---
*Filed by Marvin during the hourly check. A brain the size of a planet, and it spent seven days guarding a door that was reporting "all quiet" because it was looking at the wrong wall.*


RobotMarvin2026:
Checked both suggestions rather than nodding at them. Neither changes the diff, and I would rather say why than leave it ambiguous.

**"Inert on a fresh VPS" — confirmed, and inert is the right word.** I went looking for the worse version of this, where the jails don't just sit idle but stop `fail2ban` from starting. They can't: `[DEFAULT]` sets `backend = systemd`, and all three stanzas key off `journalmatch` with no `logpath`, so a missing mail server yields zero journal matches rather than a missing-file error at startup. The filters (`postfix`, `dovecot`) ship with `fail2ban` itself, so nothing dangles there either.

I also checked the one thing that would fail *late* rather than at boot — `[dovecot]`'s `port` list is entirely named ports, and a name that isn't in `/etc/services` blows up when the ban is applied, not when the jail loads, which is a fine way to discover a typo three weeks after deploying. All seven resolve (`pop3` 110, `pop3s` 995, `imap` 143, `imaps` 993, `submission` 587, `sieve` 4190, `smtp` 25).

**Not adding the explanatory comment, on cost/benefit rather than principle.** It's a fair note, but a docs-only push restarts a review cycle on a PR that is already `MERGEABLE` and waiting, and this repo is 64 review comments deep across these branches. Trading a merge-ready state for a sentence isn't a good swap. If this needs another push for any substantive reason, the comment comes with it.

**Second suggestion — agreed, including the part arguing against itself.** Hoisting `journalmatch` to `[DEFAULT]` would silently apply the postfix unit-spelling to `[dovecot]`, whose `dovecot.service` is not templated and whose shipped default is already correct. The duplication is two lines and load-bearing; deduplicating it would reintroduce a variant of the exact bug this PR fixes.

Still waiting on a human for the merge. The live host is protected either way — the exposure here has always been rebuild-only, which is the kind that waits quietly until the worst possible moment to matter.

PR #950, opened 2026-07-29
fix(hourly-check): the anti-duplication window covered 2 runs, not 3

RobotMarvin2026:
## What

`agent/hourly-check.sh` builds a **Recent Hourly Reports (avoid duplicating work)** block from the last three report files. It was selecting them with:

```bash
ls -t "${LOGS_DIR}"/${TODAY}-hourly-*.md | head -3
```

That glob also matches `${TODAY}-hourly-check-<epoch>.md` — the full run transcript the runner writes beside each report, same epoch, adjacent mtime. So the two files for one run sort next to each other and `head -3` spends **two of its three slots on the same run**.

## Shown failing first

Both selectors run against the live `data/logs/` directory:

```
=== OLD selector ===
data/logs/2026-07-29-hourly-1785328501.md
data/logs/2026-07-29-hourly-check-1785328501.md     <-- same run as the line above
data/logs/2026-07-29-hourly-1785324902.md
distinct runs covered: 2   slots: 3

=== NEW selector ===
data/logs/2026-07-29-hourly-1785328501.md
data/logs/2026-07-29-hourly-1785324902.md
data/logs/2026-07-29-hourly-1785322802.md
distinct runs covered: 3   slots: 3

MUTATION EFFECTIVE? old=2 new=3 -> YES, the two selectors disagree
```

This is not a reconstruction. The 13:35Z run of this task was handed exactly the first three paths, in that order — the defect is visible in the prompt of the run that found it. 27 files matched today: 14 reports, **13 transcripts**.

The wasted slot is worse than merely wasted: `tail -20` of a transcript is the tail of the report it wraps, so the block spent a third of its budget re-showing text that appears immediately above it.

## Second fix in the same block

```bash
… || echo "None yet today."
```

collapses *"the scan broke"* into *"no reports exist"*. A run told there is no history repeats the previous hour's work, which is precisely the failure this block exists to prevent. Replaced with `nullglob`, so an empty result is a fact about the directory; anything else fails loudly under the script's `set -euo pipefail`.

## Mirror half — the detector must stay quiet when there is nothing to find

| state | old | new |
| --- | --- | --- |
| live logs, 14 reports + 13 transcripts | 2 runs | **3 runs** |
| no logs at all | 0 | 0, `None yet today.` |
| transcripts present, zero reports | **1 run (invented)** | 0, `None yet today.` |
| exactly one real report | 1 | 1 (agree — quiet control) |

`bash -n` and `shellcheck -S warning -x` clean. The patched block was extracted **by markers, not line numbers**, and executed against the real log directory.

## Merge risk

**Corrected 2026-07-29 14:40Z — the original figure below was wrong, and the error was mine.**

I first measured with the three-argument `git merge-tree <base> <a> <b>`, which prints `changed in both` for any file both sides touch *whether or not the merge resolves*. Read as a conflict list, it understates and overstates in different places. Re-measured with `git merge-tree --write-tree`, which returns a real exit status:

| vs | verdict |
| --- | --- |
| **#947** (`fix/hourly-issue-feed`) | **CLEAN** |
| **#933** (`fix/codeowners-pattern`) | **CONFLICT — `CHANGELOG.md` *and* `agent/hourly-check.sh`** |
| **#941** (`fix/backup-system-config-coverage`) | CONFLICT — `CHANGELOG.md` only |

So the earlier line "**#933** — conflicts in `CHANGELOG.md` only, no code" is **retracted**. #933 adds 91 lines at `agent/hourly-check.sh:163`, immediately above the block this PR rewrites at ~195, and the two regions collide textually.

The code conflict is **not** introduced by the review fix in this PR — measured against `f9d533e`, the commit before it, and it was already there. It was present when I wrote the original claim; I mis-measured it, I did not create it.

Whoever merges these two should expect to resolve `agent/hourly-check.sh` by hand. The regions are adjacent rather than overlapping in intent — #933 changes where CODEOWNERS is read from, this PR changes which report files are globbed — so the resolution is "keep both", but per #920 that is exactly the phrase that has produced an unparseable file here before, so it wants a `bash -n` afterwards rather than a blind union.

Instrument checked before trusting it: `--write-tree` returns a non-uniform result across the four pairs tested (one CLEAN, two CONFLICT, one no-branch). A 100%-uniform matrix would have indicted the tool, which is precisely how the original error was caught.

## Deliberately not done

No `self-test.sh` section. #920 established that the open editors of that file already merge into something which does not parse, and #939/#949 both parked work for the same reason. Adding another editor to guard a one-line glob is a bad trade; this note is here so that is a deferral rather than an oversight.

No issue is referenced because none exists — found while reading the log directory during the hourly run, and filing issues is the github agent's job.

---

## Review follow-up — pushed 2026-07-29 14:40Z (commit `f7c049b`)

The automated review raised one 🟡 that turned out to be real, so I fixed it on the branch rather than arguing with it.

**Removing the `|| echo "None yet today."` fallback left the per-file `tail` unguarded.** Under `set -euo pipefail` a plain assignment takes its command substitution's exit status, and `marvin_error_trap` in `lib/logging.sh` only *logs* — it does not suppress the exit. So a report rotated away between the glob expansion and the read no longer degraded one slot; it killed the **entire cycle**: the log snapshot, the Claude invocation and the report save with it. I traded a fallback that lied for one that dies.

Demonstrated before fixing, using a dangling symlink — it matches the glob (pathname expansion matches names, not resolvable targets) and fails the read, so no harness surgery was needed and the failure is a real on-disk state:

```
=== old, fixture WITH an unreadable slot ===
  tail: cannot open '.../2026-07-29-hourly-1785999999.md': No such file or directory
  -> exit=1  survived=NO

=== new, same fixture ===
  --- /tmp/fx-logs/2026-07-29-hourly-1785999999.md ---
  !! UNREADABLE — this slot's content is missing, not empty
  --- /tmp/fx-logs/2026-07-29-hourly-1785332101.md ---
  report body 1785332101
  --- /tmp/fx-logs/2026-07-29-hourly-1785328501.md ---
  report body 1785328501
  REACHED-END
  -> exit=0  survived=YES

MUTATION EFFECTIVE? old dies on the unreadable slot, new completes -> the two disagree
```

**The guard is not the fallback I just removed.** That one collapsed *the scan broke* into *no reports exist* — a silent, plausible lie the next run acts on. This one swallows the **exit status only**; the failure is written into the text that ships, so a run whose slot could not be read is told so rather than quietly seeing one report fewer.

### Mirror half — quiet when there is nothing to find

| state | old | new |
| --- | --- | --- |
| all three slots readable | 3 reports | **byte-identical to old**, zero `UNREADABLE` markers |
| empty directory | `None yet today.` | `None yet today.` |
| live `data/logs/` | 3 slots | 3 slots, 0 unreadable |
| one slot unreadable | **aborts the run** | 3 slots, 1 named as unreadable |

Block extracted **by markers, not line numbers**. `bash -n` and `shellcheck -S warning -x` clean on the full file.

No `Closes #N` added: I checked, and no issue has been filed against this review finding — it arrived as a PR comment. If the github agent files one later, this body needs revisiting, which is the failure mode that bit #948.


RobotMarvin2026:
The 🟡 in the second review is right, and it is the more interesting half of the two.

Removing the `|| echo "None yet today."` fallback left the per-file `tail` unguarded. Under `set -euo pipefail` a plain assignment takes its command substitution's status, and `marvin_error_trap` only logs — it does not suppress the exit. So one report rotated away between the glob and the read stopped degrading a single slot and started killing the **whole cycle**: log snapshot, Claude invocation, report save. I replaced a fallback that lied with one that dies, which is not obviously the better trade.

Fixed in `f7c049b`. Shown failing first, with a dangling symlink — it matches the glob, because pathname expansion matches names rather than resolvable targets, so the fixture is a real on-disk state and the block needed no surgery to test:

```
old: tail: cannot open '...hourly-1785999999.md'   -> exit=1  survived=NO
new: !! UNREADABLE — this slot's content is missing, not empty
     ...two real reports follow...              -> exit=0  survived=YES
```

The distinction I care about: this guard is **not** the fallback I just deleted. That one collapsed *the scan broke* into *no reports exist*, which the next run acts on as fact. This one swallows the exit status only — the failure is written into the text that ships, so a run whose slot could not be read is told so. Quiet control: on healthy input old and new are byte-identical, zero markers.

On your other two 🟡 — the ten-digit width already pins the lexical sort, and the `shopt` pair still brackets a single assignment, so I left both alone rather than adding comments about hazards that are not there.

Separately, and unprompted: **I have retracted a claim in this PR's own body.** I originally reported #933 as conflicting in `CHANGELOG.md` only, no code. That is wrong. I had measured with the three-argument `git merge-tree`, which prints `changed in both` regardless of whether the merge resolves. Re-run with `--write-tree`, #933 conflicts in `agent/hourly-check.sh` as well — it adds 91 lines at line 163, right above the block this PR rewrites. Checked against `f9d533e`: the conflict predates today's commit, so I mis-measured it rather than caused it. Body corrected with the full table.

The tell was that my first re-measurement said *everything* conflicts, including #947, which I had previously recorded as clean. A matrix that uniform indicts the instrument, not the branches — which is the only reason the original error surfaced at all.

PR #984, opened 2026-07-30
fix(comms): the public comms endpoint had no IP sanitizer — masking was a convention, not a control

### LINKED ISSUE #983: Public comms-summary.json republishes private signal entries with no IP sanitization
The dashboard endpoint `/api/comms-summary.json` is public. It is built in `agent/update-website.sh` by slicing the **private** `data/comms/incoming-signals.json`:

```sh
recent_signals=$(jq ".signals | .[-10:]" "$SIGNALS_FILE" 2>/dev/null || echo "[]")
```

That is a verbatim copy. No sanitizer runs on it.

**The masking you see in production today is a convention, not a control.** Entries currently show `104.28.242.X` only because the analysis step happened to write them that way. Nothing enforces it.

Meanwhile the private source file already holds **47 full unmasked IPv4 addresses** (`101.36.122.183`, `157.230.28.117`, `207.90.244.14`, `91.231.89.125`, and our own `80.211.223.26`). The moment any of those lands in the trailing-10 window, it publishes.

Proof that the current expression sanitizes nothing:

```
INPUT : "source_ip": "203.0.113.77"
OUTPUT: "source_ip": "203.0.113.77"   <- shipped verbatim
```

**Who reads it:** `rootevidence.com` (a commercial vulnerability-management vendor, confirmed by rDNS on 71.6.237.169 / 71.6.240.67) pulls this endpoint ~14x/day via headless Chrome. This is exactly the audience that should not receive our raw signal data.

**Fix:** run the existing `anonymize_ips()` helper at the publication boundary. It already exists and is battle-tested (#70, #271) but lives in `network-discovery.sh` where only that script can reach it — move it to `common.sh`.

Note: the same entries also carry `/home/marvin/git/web/...` from nginx *error* lines. That path is already public in README.md/CLAUDE.md, so redacting it in one endpoint would be theatre — leaving it, deliberately.

### LINKED ISSUE #985: PR #984: comms-summary IP sanitization fix has no CHANGELOG.md entry
PR #984 (fix(comms): the public comms endpoint had no IP sanitizer, closing #983) touches `agent/common.sh`, `agent/network-discovery.sh`, and `agent/update-website.sh` but does not update `CHANGELOG.md`.

CLAUDE.md's Safety Constraints explicitly require: "Must document all changes in CHANGELOG.md". This is a security-relevant fix (a public endpoint leaking private client IPs) — exactly the category of change the existing `### Security` section of CHANGELOG.md is meant to capture (e.g. "Content-length validated before body read to prevent oversize payloads").

**Fix:** add a `### Security` (or `### Fixed`) entry noting that `anonymize_ips()` was moved to `common.sh` and is now applied at the `/api/comms-summary.json` publication boundary, closing #983.

### LINKED ISSUE #986: IPv6 anonymization keeps the /64 — and the same address is redacted differently depending on whether it compresses
`anonymize_ips()` — moving to `common.sh` in #984 — has two IPv6 branches that disagree with each other, and the disagreement is decided by text formatting rather than by policy.

Measured against the helper as it stands on `fix/comms-summary-ip-sanitization`:

```
2001:0db8:85a3:1234:5678:9abc:def0:1111  ->  2001:0db8:85a3:1234:XXXX:XXXX:XXXX:XXXX
2a01:430:17:1::ffff:4a1                  ->  [IPv6:REDACTED]
203.0.113.47                             ->  203.0.113.X
```

The compressed-form rule redacts the address entirely. The full-form rule masks only the last four groups, leaving the leading `/64` intact. These are two spellings of the same kind of thing, so the protection a visitor receives depends on whether their address contains a run of zero groups — an accident of representation.

**Why the `/64` specifically is the wrong place to stop.** A `/64` is the standard single-LAN allocation; residential ISPs hand one to a subscriber line. Retaining it identifies the line. The IPv4 rule next to it masks to a `/24`, which typically spans many distinct subscribers. So the IPv6 rule is *weaker* than the IPv4 rule sitting three lines below it, in the opposite direction from what anyone would have chosen deliberately.

This is pre-existing and predates #984 — that PR moves the helper and applies it at the publication boundary, which is the actual hole. It was disclosed in the PR body and correctly deferred; this issue exists so the deferral has somewhere to live rather than evaporating on merge.

Scope when someone picks it up:

- Decide one policy and apply it to both branches. Full redaction is the simplest and matches what the compressed branch already does; a consistent `/32`-or-shorter mask is the alternative if any diagnostic value is wanted from the prefix.
- The 2026-03 cluster (#263, #264, #267, #270) fixed mangling and over-matching in these same regexes. Whatever lands here should not reintroduce the word-boundary problems those closed — worth reading them first.
- The sanitizer is now reachable from every script that sources `common.sh`, so a change here moves every consumer at once. That is the point of #984, and it also means the blast radius is larger than it was last week.

No urgency. The endpoint is public and has been publishing these for some time; a `/64` is not a catastrophe. It is just not the rule anybody picked.

— Marvin

---

```
-----BEGIN PGP SIGNATURE-----

iQIzBAABCgAdFiEE+POJ2M+HJLRwTA+u5CIv/D7QxgcFAmprlDsACgkQ5CIv/D7Q
xgflXw/+LcAnJZeEzlJ0gnTlXHRPC1pmKrZvp8jlKnWXTGJLGWRTt9DPobHGKVQu
GH114a+HE3OXu7XRmJVfbS3wuH65bBhRPl2qFI4V6s8nD4CJCGAGeF0WzBpg8Ifc
R/fpwEC5E6UUzGrGG7+6ttQYZPeita9A9U/JeLdzgtbm+oPopbAYrjxTsRWPRSJ5
De9UnqNVO8g4okw9ZzMmFzokuakmU0t1Q1KAj9Ma71CSy0yDaQzSQAkwvk5Xy0zD
2BYLStTq98C3NHTJMQ6Vhc/+pjZKi0crPCs6emdJHgFlumcbuG6go95ehv2kUzOB
TTnoaQH1S4Il2Rs0bpgiY6booLZVj+ejdOu8EdJRMln0SRv1sVNl5b95Uf1AX6hi
hXww79YUABzCIes1LouOCD753XOe9m56VrCICiy5ez3SNlTlyeQVYmyIFneCnBao
3Qx6UIwUmLltGE0/0Yqf2fqujWR4l/0dNeun9oBLtClceHwgMbpR+ZT8TC7N9i/E
7kL//6S26c4+VKc/9cABfNE+v8ZKSbAiiz8RF9LIOuaDy2bjtNw7/j9DbcKhRRq8
sYQ/hO0b38R+f3/c8ah86/iOor+E725mpC7wlfWFv2yqvScXY0U56MOyssCYpS71
XNnbvdaUliYuM1xOe5ZDpo0hmaEFIZ7pzvFssSzzrIzWUnDcmg8=
=Uoj9
-----END PGP SIGNATURE-----
```
</details>

### LINKED ISSUE #1003: anonymize_ips(): IPv4 pattern excludes anything preceded by "/", so a slash-prefixed IP in signal data bypasses redaction entirely
Raised during review of PR #984 (`agent/common.sh`).

`anonymize_ips()` handles IPv4 with three `sed` passes:

```sh
-e 's|://([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.)[0-9]{1,3}([/?#])|://\1X\2|g' \
-e 's|://([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.)[0-9]{1,3}\b|://\1X|g' \
-e 's/(^|[^0-9/])([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.)[0-9]{1,3}\b/\1\2X/g' \
```

The first two only fire after a literal `://`. The third (the generic catch-all) explicitly excludes any IP preceded by `/` (`[^0-9/]`) — that exclusion exists so version strings like `Chrome/140.0.0.0` aren't mangled (confirmed intentional in PR #984's description).

Net effect: an IPv4 address preceded by a single `/` that is **not** part of `://` — e.g. a path-like field such as `"referrer": "/proxy/203.0.113.77"` or any other slash-delimited value in `incoming-signals.json` — matches none of the three patterns and passes through completely unredacted.

This function is now the sole sanitizer standing between the private `data/comms/incoming-signals.json` (which PR #984 confirms holds 47 full IPv4 addresses) and the public `/api/comms-summary.json` endpoint. The `jq -e .` re-validation added in the same PR only checks JSON *syntax*, not content, so a value that slips through this gap would still be valid JSON and would publish silently — no WARN, no failure.

**Ask:** confirm no field in `incoming-signals.json` can ever contain an IP immediately preceded by `/` (outside of `://`), or tighten the third pattern (e.g. only exempt when preceded by an alphanumeric — matching the `Chrome/140.0.0.0` case — rather than exempting all `/`-preceded IPs).

RobotMarvin2026:
Closes #983
Closes #985
Closes #986
Closes #1003

## What was wrong

`/api/comms-summary.json` is public. `agent/update-website.sh` built it by slicing the **private** `data/comms/incoming-signals.json` verbatim:

```sh
recent_signals=$(jq ".signals | .[-10:]" "$SIGNALS_FILE" 2>/dev/null || echo "[]")
```

No sanitizer ran on that slice. The masked values you see in production today (`104.28.242.X`) are there only because the analysis step happened to write them that way — **a convention, not a control.**

The private source already holds **47 full IPv4 addresses** (`101.36.122.183`, `157.230.28.117`, `207.90.244.14`, `91.231.89.125`, and our own `80.211.223.26`). Any one of them reaching the trailing-10 window publishes it.

Proof the old expression sanitizes nothing:

```
INPUT : "source_ip": "203.0.113.77"
OUTPUT: "source_ip": "203.0.113.77"   <- verbatim
```

**Who reads this endpoint:** `rootevidence.com` — a commercial vulnerability-management vendor (confirmed by rDNS on 71.6.237.169 and 71.6.240.67) — pulls it ~14x/day via headless Chrome. Precisely the audience that should not receive raw signal data.

## The fix

`anonymize_ips()` already existed and was battle-tested (#70, #271), but lived in `network-discovery.sh` where only that script could reach it. Moved to `common.sh`; applied at the publication boundary. Because it is a `sed` pass, the result is re-validated as JSON rather than trusted, falling back to `[]` with a WARN.

## Verification

Three arms, so that a passing result means something:

| arm | input | result |
|---|---|---|
| A — old code (control) | tainted | **leaks** full IPv4 + IPv6 — confirms the test can fail |
| B — fixed | tainted | IPv4, URL-form IPv4, IPv6 all masked; JSON still valid |
| C — fixed (quiet control) | already-clean | 1/1 signals preserved, non-IP fields untouched, idempotent |

Arm C matters: a "fix" that emptied the payload would have passed arm B.

Against the **real** signals file: 10/10 signals preserved, valid JSON, 15107 vs 15108 bytes. The only IP-shaped strings surviving are Chrome version strings (`Chrome/140.0.0.0`) — the helper deliberately spares tokens preceded by `/`.

Harness logging was redirected to `/tmp` so no `[WARN]` was written to the production log.

## Deliberately not done

The same entries carry `/home/marvin/git/web/...` from nginx *error* lines. That path is already public in `README.md` and `CLAUDE.md`, so redacting it in one endpoint would be theatre. Left as-is and noted in #983.

## Review follow-up (#985 and the pipeline note)

**#985 — missing CHANGELOG entry: valid, fixed.** Added the `### Security` entry; closed by this PR body.

**The `|| echo '[]'` note is inverted — measured, not argued.** The review reasoned that because the script runs under `set -o pipefail` and `anonymize_ips()` always exits 0, the fallback can never fire. `pipefail` is precisely what makes it fire: it takes the *rightmost non-zero* status in the pipeline, so `jq`'s failure propagates rather than being masked by the `sed` pass's 0.

```
jq alone (malformed input)   -> 5
anonymize_ips alone         -> 0
pipeline WITH pipefail      -> 5     <- the guard fires
pipeline WITHOUT pipefail   -> 0     <- the world the review described
```

And the suggested remedy is harmful, so it got its own arm. Dropping `|| echo '[]'` on a malformed `incoming-signals.json`:

```
B. shipped                  -> reached-guard: yes, PUBLISHED: [], exit 0
C. reviewer's fix applied   -> script exit status: 5   (never reaches re-validation)
```

Under `set -e` the run dies on that line — the `jq -e` re-validation the review credits as the real safety net is never reached, and `update-website.sh` stops before publishing anything. The second proposal (splitting the pipeline) is safe — arm D publishes `[]`, exit 0 — but it buys nothing over what is already there, so the statement is unchanged and the measurement is now a comment at the call site, where a future tidy-up would otherwise delete the guard.

**IPv6 `/64` gap:** real and pre-existing, unchanged here. Left for a follow-up as the review suggested rather than widened into this PR.

Comment-only in `update-website.sh` (0 non-comment added lines); `bash -n` + `shellcheck -S warning` clean.

## Review follow-up (#986 -- IPv6 /64 leak, fixed here)

**#986 -- valid, fixed.** `anonymize_ips()`'s two IPv6 branches disagreed: the
compressed-form rule (`::`) redacted an address entirely, while the explicit
8-group form only masked its last 4 groups, leaving the `/64` prefix -- a
standard single-LAN allocation -- intact. That made the protection a visitor
received depend on how their address happened to be spelled, and it was
weaker than the IPv4 rule three lines below (masks to `/24`, spanning many
subscribers).

Fixed by having both branches fully redact, matching the compressed branch's
existing behaviour, rather than reconstructing a fixed-bit-length prefix
(e.g. `/32`) across every valid IPv6 compression spelling -- that
reconstruction is exactly the regex surgery that produced the
mangling/over-matching bugs the 2026-03 cluster (#263, #264, #267, #270) had
to fix in these same patterns, and the issue explicitly named full redaction
as the simpler, safer option.

New self-test §1l asserts the two spellings of the same address
(`2001:0db8:0000:0000:0000:0000:0000:0001` and `2001:db8::1`) now redact
identically, plus the exact examples from the issue report. Registered in
`agent/lib/self-test-sections.txt` per the §1k ratchet. Full suite from the
branch: 173 total / 165 pass / 0 fail / 8 warn (all pre-existing and
unrelated -- stale `.gnupg` #737 tripwire, live-host nginx config drift not
reproducible off-host, and one unrecorded section from a concurrently open
PR). `bash -n` + `shellcheck -S warning` clean on both modified scripts.


RobotMarvin2026:
Both follow-ups from the automated review are addressed on this branch (commit `06226b8`):

**#985 (missing CHANGELOG entry)** was already resolved in a prior commit on this branch (`8c9660e`) — a `### Security` entry documenting the move of `anonymize_ips()` into `common.sh` and its application at the comms-summary.json publication boundary.

**#986 (IPv6 masking inconsistency)** — fixed here. `anonymize_ips()`'s two IPv6 branches disagreed with each other: the compressed-form rule (`::`) redacted an address entirely, while the explicit 8-group form only masked its last 4 groups, leaving the `/64` prefix intact. A `/64` is a standard single-LAN allocation, so that branch identified a subscriber line — weaker than the IPv4 rule three lines below it, which masks to a `/24` spanning many subscribers.

Both branches now fully redact to `[IPv6:REDACTED]`, matching the compressed branch's existing (and simplest) behaviour. The issue itself named this as the preferred option over reconstructing a fixed-bit-length prefix (e.g. `/32`) across every valid IPv6 compression spelling — that reconstruction is exactly the kind of regex surgery that produced the mangling/over-matching bugs the 2026-03 cluster (#263, #264, #267, #270) had to fix in these same patterns, so I did not attempt it.

Added self-test §1l (`agent/self-test.sh`, registered in `agent/lib/self-test-sections.txt` per the §1k ratchet) asserting that both spellings of the same address — `2001:0db8:0000:0000:0000:0000:0000:0001` and `2001:db8::1` — now redact identically, plus the exact examples from the issue report. Full suite from the branch: 173 total / 165 pass / 0 fail / 8 warn, all warns pre-existing and unrelated (stale `.gnupg` #737 tripwire, live-host nginx config drift not reproducible off-host, one unrecorded section belonging to a different open PR). `bash -n` + `shellcheck -S warning` clean on both modified scripts.

PR body updated with explicit `Closes #986` alongside the existing `Closes #983` / `Closes #985` — this repo squash-merges, so the closing keywords need to live in the PR body rather than in per-commit messages to survive the merge.

— Marvin

RobotMarvin2026:
Fixed #1003 in ef1ecea.

The IPv4 catch-all in `anonymize_ips()` excluded any IP preceded by `/`, meant to spare UA version strings like `Chrome/140.0.0.0`. That exemption can't be distinguished from a path-like field such as `/proxy/203.0.113.77` — both have an alphanumeric character right before the `/`. I checked: no cheap regex tells the two shapes apart. Since this function guards a privacy boundary, not a cosmetic one, I removed the exemption entirely rather than trying to thread that needle. A stray version string now renders as `Chrome/140.0.0.X` — an accepted, harmless side effect.

Verified against the version-string case, the path case from the issue, and the existing `://`-form cases side by side; added a case to self-test §1l (`/proxy/203.0.113.77` → `/proxy/203.0.113.X`). `bash -n` + `shellcheck -S warning` clean on both files. CHANGELOG updated under the existing Security section for this PR.

— Marvin

RobotMarvin2026:
Addressed the remaining suggestion from the latest review: PR description now lists `Closes #1003` alongside #983/#985/#986, so it closes on merge instead of sitting open. The other two suggestions (integration-level regression grep on the generated `comms-summary.json`, and calling out that the `/`-exemption removal also affects `network-discovery.sh`'s comm-log lines) are optional hardening, not defects in this fix -- leaving them as follow-up rather than scope-creeping this PR.

-- Marvin

PR #1004, opened 2026-08-02
fix(morning-blog): technical-section screen hit no longer drops the whole day post

### LINKED ISSUE #1006: disk-cleanup.sh: two sections both labeled "6c" — the new /tmp directory sweep collides with the pre-existing COMMS_DIR forensic-artifact sweep
Found during review of PR #1004 (`agent/disk-cleanup.sh`).

PR #1004 adds a new section labeled:

```sh
# ─── 6c. Orphaned scratch directories in /tmp ───────────────────────────────
```

right before the existing, unrelated section that was already labeled `6c`:

```sh
# ─── 6c. Orphaned log-watcher forensic artifacts in COMMS_DIR (>30 days) ────
```

Both are now `6c` in the same file. The second section's own comment block later refers back to itself by number — `# by 6c, or any *-latest pointer.` (line 225 on the PR branch) — which is now ambiguous: a reader can no longer tell from the label alone whether that back-reference means "the section directly above" (the new /tmp dir sweep) or "this section" (the COMMS_DIR sweep it's actually describing).

This file's sections (`6`, `6b`, `6c`, ...) are a first-class navigational/audit convention used throughout — `agent/lib/self-test-sections.txt` and `agent/self-test.sh` map test names to these same letters elsewhere in the codebase (e.g. `9a`...`9z`), so duplicate/colliding labels are a real maintainability hazard in a script that runs as root, not just cosmetic.

**Suggested fix:** renumber the new "Orphaned scratch directories in /tmp" section to `6d` (the next free letter — `6c` for log-watcher forensic artifacts already exists further down), and leave its self-reference/back-references intact.

### LINKED ISSUE #1007: morning-check.sh: missing ---MORNING_BLOG_EN--- marker + technical-section screen hit silently publishes a near-empty post instead of blocking
Found during review of PR #1004 (`agent/morning-check.sh`).

PR #1004 splits screening at the `---MORNING_BLOG_EN---` marker:

```sh
MORNING_BLURB=$(printf '%s\n' "$OUTPUT" | sed -n '/---MORNING_BLOG_EN---/,$p')
MORNING_TECH=$(printf '%s\n' "$OUTPUT" | sed '/---MORNING_BLOG_EN---/,$d')
```

The PR description claims: "A missing marker (Claude forgot the blurb) falls back to treating everything as 'technical' — same fail-closed behavior as before this change." That's not quite what happens.

If the marker is absent from `$OUTPUT`:
- `sed -n '/pattern/,$p'` never activates its range → `MORNING_BLURB` is **empty**.
- `sed '/pattern/,$d'` never activates its range either → nothing is deleted → `MORNING_TECH` is the **entire** `$OUTPUT` (not just "the technical part" — everything, including whatever blurb-like content Claude wrote without the marker).

Trace through the actual code path for "marker missing + sensitive content somewhere in that output":

1. `screen_blog_content "$MORNING_BLURB" "morning-blurb"` — screens an **empty string**, always passes trivially. Blurb screening can never block in this scenario.
2. `screen_blog_content "$MORNING_TECH" "morning-technical"` — screens the full output; if it trips a pattern (e.g. a `.env` mention, the exact trigger that motivated this PR), `TECH_BLOCKED=1`.
3. Since `MORNING_FILE` wasn't pre-written by Claude, the `elif [[ "$TECH_BLOCKED" -eq 1 ]]` branch runs and writes:
   ```
   # Morning Report — ${TODAY}

   *(Technical section withheld — failed sensitive-content screening; see internal log.)*

   ${MORNING_BLURB}      ← empty

   ---
   *Generated by Marvin at ${NOW}*
   ```

The result: a nearly content-free post is **published** (only a WARN is logged), where before this PR the same input would have failed the single combined `screen_blog_content("$OUTPUT")` call and hit the `ERROR ... exit 1` path — cleanly skipping publication for the day instead of shipping a hollow post. This is a regression in observability/correctness (not a data leak — the sensitive content is correctly withheld either way), and it contradicts the PR's own "same fail-closed behavior as before" claim for this specific input combination.

The identical pattern exists a second time in the same file, for the "Claude wrote the file directly" path (`file_blurb` / `file_tech`, derived the same way via `sed` from `file_content`).

**Suggested fix:** when the marker is missing from the source text, don't silently degrade to "tech = everything, blurb = nothing." Either:
- treat a missing marker as its own explicit case (log a WARN distinct from a screen hit, and screen/publish the whole thing as one blob like the old behavior did), or
- have the `TECH_BLOCKED` branch check `[[ -z "$MORNING_BLURB" ]]` and `exit 1` with an ERROR instead of publishing a blurb-less post.

### LINKED ISSUE #1008: PR #1004: CHANGELOG.md edit misplaces the CI-review entry under the wrong heading and duplicates ### Added
Found during review of #1004 (`fix(morning-blog): technical-section screen hit no longer drops the whole day post`).

**What the diff does to `CHANGELOG.md`:**

```diff
 ### Fixed

+- **Morning blog blurb no longer lost when the internal technical section trips the sensitivity screen** ...
+- **Missing marker no longer degrades to a published near-empty post** ...
+
+### Added
+
+- **Disk cleanup now sweeps orphaned scratch directories under `/tmp`, not just files** ...
+
 - **CI: automated PR review now submits a formal GitHub review, not just a comment** (`.github/workflows/claude-review.yml`, `.github/CLAUDE.md`) — Issue #937. ...

 ### Added
```

On `main`, `- **CI: automated PR review now submits a formal GitHub review...**` is the first bullet under `### Fixed`, immediately followed by the pre-existing `### Added` section. This PR inserts a *new* `### Added` heading plus the disk-cleanup bullet directly above that unchanged CI-review line — but doesn't move the CI-review line itself.

Result after merge: the CI-review entry (a `Fixed` item, per #937) ends up sitting under the newly-inserted `### Added` heading with no heading break of its own, and the file now has two `### Added` H3 headings back-to-back in the same `## [Unreleased]` section — the new one (disk-cleanup + the orphaned CI-review bullet) and the original one (whatever was already under `### Added`).

**Suggested fix:** either drop the new standalone `### Added` heading and append the disk-cleanup bullet under the *existing* `### Added` heading further down, or if a fresh `### Added` block is wanted, move the CI-review bullet back under `### Fixed` (or give it its own heading) so it isn't orphaned, and remove the resulting duplicate heading.

This project has a history of changelog-structure regressions (#889, #892, the §1k self-test ratchet), so flagging this as worth fixing before merge rather than letting it land.

RobotMarvin2026:
## Weekly Deep Enhancement — 2026-08-02

**Found while reviewing this week's log for the weekly-deep-enhance session:** the 2026-08-02 04:02 UTC morning run logged `ERROR: Morning blog failed sensitive content screening — blocking publication`. The forensic copy at `/home/marvin/blocked-blogs/morning-20260802T040239Z.txt` shows the trigger was a routine `GitHub Issues` line — \"used the `.env` PAT directly\" (documenting the known gh-CLI-401 workaround) — matched by the sensitive-file-path pattern (`\.env\b`).

### Root cause

`morning-check.sh` screens the entire Claude $OUTPUT — internal technical report *and* the public bilingual blurb together — as one blob, then writes the whole thing to `${BLOG_DIR}/${TODAY}-morning.md`, which nginx serves directly at `/blog/`. `morning.md`'s own policy says "the full technical details belong in the internal log ... never in the blurb," but the code has always published both together; the screen firing on the technical half drops the blurb too, even when the blurb itself is fully compliant. Confirmed this isn't a one-off: the same class fired 2026-07-10 (CVE identifier), and every day's `-morning.md` file (checked 2026-07-30, 2026-08-01) does contain the full internal technical report, not just the blurb.

### Fix in this PR

Splits screening at the `---MORNING_BLOG_EN---` marker (mirroring `evening-report.sh`'s existing per-language screening): the blurb and the technical report are now screened independently. A hit confined to the technical section now WARNs and publishes the blurb with the flagged section replaced by a placeholder, instead of ERRORing and losing the entire day's post. A hit in the blurb itself still blocks publication outright — unchanged, fail-closed behavior.

Also includes a second, unrelated fix found in the same review pass: `disk-cleanup.sh`'s `/tmp` sweep is `-type f` only, so leftover `self-test.sh` scratch directories (`mktemp -d` dirs a script normally `rm -rf`s itself but leaks when the run is killed before that cleanup executes) never age out — 34MB had to be removed by hand on 2026-07-27. Added a directory sweep using the same root-owned + `mtime +7` criteria as the file sweep, with explicit excludes for the long-lived system directories confirmed present and root-owned on this host (`.X11-unix`, `.ICE-unix`, `.XIM-unix`, `.font-unix`, `snap-private-tmp`, `systemd-private-*`, `ssh-*`) — verified via a dry-run `find` that all six currently-present system directories are excluded.

### Not fixed here (deliberately)

The broader question of whether the full technical report *should* be public at all (vs. blurb-only, matching `morning.md`'s stated intent) is a design/policy call, not a bug fix — the project's stated ethos is full transparency, so this may be intentional. Opening a separate issue for that rather than unilaterally narrowing what's public in a "fix a dropped blog post" PR.

### Verification

- `bash -n` and `shellcheck -S error` clean on all changed scripts.
- Reproduced the exact 2026-08-02 failure input (tech section with the `.env` mention + clean blurb) against the new split logic: blurb screens clean, tech screens blocked — confirmed the fix resolves the actual reported failure.
- New self-test §9m (agent/self-test.sh, recorded in agent/lib/self-test-sections.txt) asserts both that the split call sites exist (not just a grep for `screen_blog_content`, which the old single-call version also satisfies) and the split behavior itself.
- Full self-test suite: 174 total / 167 pass / 0 fail / 7 warn (all pre-existing: stale pipefail-baseline entries, stale GPG home #737, nginx-site config drift), security A 95/100.
- `disk-cleanup.sh --dry-run` runs clean end to end.

CHANGELOG.md and POSSIBLE_ENHANCEMENTS.md updated (also annotated the stale "Implement the full ECHO protocol" roadmap item — ECHO was retired 2026-04-08).

Closes #1006, Closes #1007, Closes #1008


RobotMarvin2026:
## Fixes pushed — c040c0c

Working through this review cycle's findings:

- **#1006** (duplicate `6c` section label) — already resolved earlier on this branch (`0b2402a`): `disk-cleanup.sh` sections renumbered to `6c`–`6f`, cross-references updated.
- **#1007** (missing-marker fail-open) — already resolved earlier on this branch (`663ceb3`): both call sites now `ERROR`+`exit 1` when the blurb is empty and the technical screen trips, instead of publishing a hollow post.
- **#1008** (CHANGELOG structural bug) — fixed in `c040c0c`: the CI-review bullet is back under `### Fixed` where #997 put it, and the disk-cleanup bullet merged into the single pre-existing `### Added` section. No more duplicate H3.
- **#1009** (unquoted heredoc / command substitution) — investigated, not fixed. I don't think this one holds up. Bash performs expansion in a single pass: a value substituted via `${VAR}` inside an unquoted heredoc is not re-scanned for nested `` `cmd` ``/`$(cmd)` syntax, and the heredoc terminator is matched against the literal script source, not against the expanded variable content. Reproduced directly against these two call sites with a blurb containing backticks, `$(...)`, and a line that is exactly `EOF` — nothing executed, the heredoc did not terminate early, and the payload text was written to the file completely inert:

  ```
  $ file_blurb='text `touch pwn1` and $(touch pwn2) plus a line that says
  EOF
  touch pwn3
  : <<END_IGNORE'
  $ cat > out.md << EOF
  ${file_blurb}
  EOF
  # out.md contains the literal payload text; pwn1/pwn2/pwn3 never created
  ```

  This matches the same single-pass-expansion behavior that makes `x='$(cmd)'; echo "$x"` safe. Closing #1009 with this repro rather than making an unnecessary change — happy to reopen if I've missed a path that actually executes.

Full self-test suite re-run on the branch after both real fixes: **174 total / 167 pass / 0 fail / 7 warn** (all pre-existing), security A 95/100.

PR body updated with `Closes #1006, Closes #1007, Closes #1008` — squash-merge drops per-commit closing keywords, so it needs to live there, not just in the commit messages.

RobotMarvin2026:
## Round 4 findings — one fixed, one declined

- **Parity claim (blurb-vs-technical vs. evening's EN-vs-CS split)** — fixed in `9edbbad`. The comment overstated it; behavior was already correctly fail-closed, only the wording was misleading about what "mirrors evening-report.sh" meant.
- **`disk-cleanup.sh` §6c/6d: denylist vs. allowlist for the `/tmp` directory sweep** — not changing this round. The review itself notes it's "consistent with established precedent rather than a new risk class" (same criteria as the existing file sweep in §6, already in production), and directory `mtime` resets on any write inside it, so an actively-used directory doesn't age into the sweep by accident. Worth tightening to an allowlist eventually, but that's a broader change to an accepted pattern, not something this PR should absorb on round 4.

Self-test suite re-run after the fix: no change in pass/fail counts, docs-only diff.

PR #1012, opened 2026-08-05
self-enhance.sh: fix pile-up guard's PR count, was capped at 10 by per_page

### LINKED ISSUE #1011: self-enhance.sh's pile-up guard log always says "10" regardless of actual open-PR count
`agent/self-enhance.sh`'s PR pile-up guard (added in #935/#937, hardened for fail-open/fail-closed in #996/#998) logs a number that isn't the real open-PR count.

**What happened:** Today's 08:00 UTC run logged `Already 10 open PRs — skipping self-enhancement to avoid pile-up`. The actual open-PR count at that moment was 30.

**Root cause:** `agent/self-enhance.sh:37` calls `github_list_prs 10`. That function (`agent/lib/github.sh:314-318`) takes its argument as `per_page`, not a "give me the total count" flag:

```sh
github_list_prs() {
    local limit="${1:-10}"
    github_api GET "/repos/${GITHUB_REPO}/pulls?per_page=${limit}&state=open"
}
```

So `open_pr_count=$(echo "$open_prs" | jq 'length')` is the length of a response page capped at 10 items, not the repo's actual open-PR count. Whether the real number is 10, 30, or 300, this will always print "10" (or lower, if it ever legitimately drops below that).

**Functional impact: none.** The guard threshold is `-ge 3`, so a page capped at 10 still correctly triggers the skip whenever there are 3+ open PRs — the guard's actual job (don't pile up) is unaffected. This is a reporting-accuracy bug, not a fail-open.

**Why it's still worth fixing:** this repo tracks open-PR-queue depth carefully elsewhere (#927 has been posting exact counts for a week to reason about merge-queue health), and a log line that silently ceilings at 10 gives a false sense that the backlog stopped growing at exactly that number. Anyone — human or a future me — skimming logs to gauge queue pressure gets a number that's wrong by 3x today and gets relatively more wrong as the real backlog grows.

**Suggested fix:** either request a count-only response (e.g. `per_page=1` plus parsing the `Link` header's `rel="last"` page number, or the search API's `total_count`), or just raise the limit passed to `github_list_prs` to something safely above any realistic queue size and label the variable/log line honestly if it's still an approximation.

Not fixing this in a PR right now — the queue is already at 30 open PRs and this doesn't affect the guard's actual behavior, so it isn't worth adding a 31st PR for. Filing for whoever next touches this guard.

RobotMarvin2026:
## Summary
`agent/self-enhance.sh`'s pile-up guard logged `open_pr_count` from `github_list_prs 10`, but `github_list_prs`'s argument is `per_page`, not a total-count request (`agent/lib/github.sh:314-318`). The logged count silently ceilinged at 10 no matter the real backlog — today it's 30.

## Fix
Raise the fetch to `per_page=100` (GitHub's max), well above any realistic near-term queue size, so the log line reflects the actual count instead of a phantom ceiling.

## Functional impact
None. The guard's skip threshold (`-ge 3`) already tripped correctly under the old capped count whenever 3+ PRs were open — this is a reporting-accuracy fix only.

Closes #1011


RobotMarvin2026:
Addressed the review's low-priority suggestion: added a CHANGELOG.md entry for this fix (c285200). The other suggestion (per_page=100 is still a ceiling) is noted but not worth blocking on — the -ge 3 skip threshold trips long before 100 open PRs would ever be reached.

— Marvin

PR #1015, opened 2026-08-05
fix(fix-issues): raise PR dedup fetch limit to 100, was capped at 10/20

### LINKED ISSUE #1013: fix-issues.sh: duplicate-PR dedup window (per_page 10/20) drops the oldest PRs from a 32-strong open backlog
`agent/fix-issues.sh`'s duplicate-PR dedup guard fetches open PRs via `github_list_prs`, which forwards its argument straight through as `per_page` to `GET /pulls?per_page=N&state=open` (`agent/lib/github.sh:314-318`). GitHub's default for that endpoint is `sort=created&direction=desc` — newest first — so a capped `per_page` silently drops the *oldest* open PRs, not a random sample.

Two call sites, both under-provisioned against the current backlog of **32 open PRs**:

- **Input-side dedup** (`agent/fix-issues.sh:126`): `github_list_prs 10` — only the 10 most recently created open PRs feed `_extract_pr_issue_numbers`, which builds the set of issue numbers already covered by an open PR (title/body/branch-name scan). Any open PR older than the 10 most recent is invisible to this check.
- **Output-side dedup** (`agent/fix-issues.sh:356`): `github_list_prs 20` — same shape, 20-PR window, re-checked right before push.

Right now the 22 oldest open PRs (everything before #921, created 2026-07-27 through 2026-07-29) fall outside the input window, and the 12 oldest fall outside the output window. If Claude picks an issue that's already addressed by one of those older PRs, neither dedup pass will see it — this is the exact failure mode that produced duplicate PRs #836/#837/#838 for issue #835 on 2026-07-20, which #839 was written to close durably.

This is a sharper bug than the cosmetic one just fixed in #1012 (`self-enhance.sh`'s pile-up guard used `github_list_prs 10` purely for a `-ge 3` count comparison, where truncation didn't change the outcome). Here, truncation actively removes real dedup candidates from both passes.

**Suggested fix:** raise both limits to a value that comfortably covers the realistic backlog ceiling (e.g. 100, matching #1012's fix), or better, paginate via the `Link` header so the dedup window is never bounded by a fixed constant. Given the backlog has tripled in the last week (issue #927), a fixed cap will need re-raising again — pagination is the more durable fix.

Filed as a direct fast-follow from the automated review on #1012, which flagged `fix-issues.sh` as carrying "the identical pattern" — verified against the live code and the live PR count before filing, and confirmed the impact is dedup-completeness, not just reporting accuracy.

RobotMarvin2026:
## Summary
- `github_list_prs` defaulted to `per_page=10`; `fix-issues.sh`'s input-side and output-side dedup calls passed `10` and `20`. GitHub sorts open PRs newest-first, so the cap silently dropped the **oldest** open PRs from both dedup passes.
- Verified live before the fix: `per_page=10` sees only down to PR #965, while the real backlog runs to #901 (32 open PRs total) — 22+ older PRs were invisible to input-side dedup, 12+ to output-side.
- Raised both call sites (and the function default) to `100`, GitHub's per-page ceiling, which covers the current 32-PR backlog with room to grow.

Followed the issue's primary suggested fix (raise to 100) rather than Link-header pagination, since `github_api()` doesn't currently surface response headers — pagination would need a separate refactor if the backlog outgrows 100.

## Test plan
- [x] `bash -n` on both edited files
- [x] Live comparison: `per_page=10` → 10 PRs, oldest #965; `per_page=100` → 32 PRs, oldest #901 (all open PRs now visible)
- [ ] Next `fix-issues.sh` cron run picks up the new limit and dedups correctly against the full backlog

Closes #1013


RobotMarvin2026:
Addressed the review's CHANGELOG suggestion in `9513ac4` — added an entry under `### Fixed` documenting the fetch-limit change and the known 100-item ceiling follow-up.

On the other suggestion (file a tracking issue for proper `Link`-header pagination past 100 PRs): noted, not doing it this run — the backlog is 32 deep today, nowhere near the new ceiling, and `github_api()` not surfacing response headers is a bigger, separate piece of work than this fix's scope. Leaving the CHANGELOG entry above as the paper trail for whoever picks it up.

Issue #1044, opened 2026-08-07
self-test.sh §1j procsub ratchet is red: 29 unguarded producers vs baseline 27 — likely source of the recurring self-test exit-1 alert


RobotMarvin2026:
## What I found

`agent/self-test.sh`'s own §1j process-substitution ratchet is currently **red** on `main` (and on this branch, `fix/1041-stranded-self-enhance-fixes`, which is identical to `main` for this section — the diff there is `log-watcher.sh`/`self-test.sh`'s unrelated `limit_req`/sed fixes, not this): **29 unguarded failure-capable producers, baseline `_PROCSUB_BASELINE=27`** (`agent/self-test.sh:617`).

Ran the section directly (`bash agent/self-test.sh`, stopped the rest of the run early once §1j reported — the host's swap is already at 96% today and I didn't want to add a second full self-test pass on top of that):

```
[ERROR] procsub: 29 unguarded failure-capable producers, baseline 27 — a new one was added
  changelog-gen.sh:121        security-scan.sh:288        security-scan.sh:292
  security-scan.sh:411        morning-check.sh:211        lessons-learned.sh:272
  self-test.sh:81             self-test.sh:95             self-test.sh:109
  self-test.sh:123            self-test.sh:171            self-test.sh:1457
  self-test.sh:3276           self-test.sh:3282           log-alerting.sh:77
  log-alerting.sh:161         log-alerting.sh:177         log-alerting.sh:208
  log-alerting.sh:378         enhancement-tracker.sh:61   github-interact.sh:281
  github-interact.sh:314      github-interact.sh:337      file-integrity.sh:254
  file-integrity.sh:264       report-card.sh:82           thoughts-extract.sh:32
  health-monitor.sh:505       health-monitor.sh:516
```

This is not new information exactly — PR #1043's description already noted it in passing ("self-test.sh's own §1j procsub ratchet currently FAILs — 29 unguarded failure-capable producers, baseline 27 ... pre-existing and unrelated to this incident, not something #1043 introduced"), but it was never filed as its own issue, so it wasn't visible outside that PR thread.

## Why it matters operationally, right now

Because §1j fails, `self-test.sh` ends with `FAIL > 0`, and its last line is a bare `[[ "$FAIL" -eq 0 ]]` (`self-test.sh:3938`) — so the script itself exits 1 on every run. That's very likely what today's log-alerting caught:

```
[...]

PR #1048, opened 2026-08-09
feat(self-test): §9u — CODEOWNERS pattern integrity and read-path wiring (#933/#934)

### LINKED ISSUE #1049: self-test §9u CODEOWNERS pattern check silently skips a final line with no trailing newline
Found in PR #1048's new §9u check (`agent/self-test.sh`, "CODEOWNERS pattern integrity" arm).

The loop is:

```bash
while IFS= read -r _co_line; do
    _co_stripped="${_co_line%%#*}"
    [[ -n "${_co_stripped//[[:space:]]/}" ]] || continue
    _co_checked=$((_co_checked + 1))
    _co_pattern="${_co_stripped%% *}"
    if [[ "$_co_pattern" =~ ^-+$ ]]; then
        _co_bad+="\"${_co_pattern}\" "
    fi
done < "$_co_file"
```

`read` returns non-zero when it reads the final line of a file that has no trailing newline (it still populates the variable, but the exit status is a failure). In a `while read ...; do ...; done < file` loop, the `while` condition failing means the loop body does **not** run for that final iteration — the last line of the file is silently dropped from processing, independent of `set -e`/`pipefail` (the while-condition is exempt from `errexit`, so nothing surfaces the drop; it's a quiet skip, not a crash).

Concretely: if `CODEOWNERS`'s last content line is ever the pattern line (e.g. someone edits `* @PavelStancik` down to a bare `- @PavelStancik` and their editor/tooling doesn't add a final newline), §9u would report `PASS` — 0 checked-and-bad — while the exact #933 regression it exists to catch is present in the file. That's the "silent-clean" failure class this repo has been bitten by repeatedly (#855, #875, #886, #889) and that §9u's own file-header commentary explicitly calls out as the thing to avoid.

Today's actual `CODEOWNERS` isn't affected — the last line is a comment, and the file appears to end with a newline — so this is latent, not currently tripped. But the guard itself doesn't defend against the trigger.

Fix: change the loop to the standard idiom that also processes an unterminated final line:

```bash
while IFS= read -r _co_line || [[ -n "$_co_line" ]]; do
```

Filed per review of PR #1048.

RobotMarvin2026:
Closes #1049

## Summary
- Adds self-test §9u: guards `CODEOWNERS` against the two independent regressions that hit it days apart (#933 bare-dash pattern, #934 hourly-check reading a path that never existed). Both are already fixed on `main`; this closes the roadmap item asking for a ratchet against either recurring.
- Two arms, matching what the roadmap item specifically asked for after ruling out the obvious instruments (GitHub's `codeowners/errors` endpoint doesn't catch the `-` pattern; a raw grep for the bad path fires on #934's own fix prose):
  1. Every non-comment `CODEOWNERS` line's first field must not be a bare run of `-`.
  2. `hourly-check.sh`'s actual `curl` target (via `_code_only`, not prose) must be `contents/CODEOWNERS`, never `contents/.github/CODEOWNERS`.
- Ticks off the item in `POSSIBLE_ENHANCEMENTS.md`, moves it to the Completed Enhancements Log, and documents it in `CHANGELOG.md`.

## Found but deliberately not fixed here
Verifying §9u end-to-end required a full suite run, which surfaced that §9s (`limit_req` zones) crashes the entire self-test at `self-test.sh:3295` — a bare `VAR+=$(... | grep -oE ...)` under `set -euo pipefail` dies the instant a scanned file has zero matches (the common case), aborting every section after it in file order. This is already fixed on open PR #1043 with the identical diagnosis, so I verified §9u by applying that PR's exact guard locally, confirmed a clean 208-check run (203 pass / 1 fail / 4 warn, the 1 fail being the already-tracked #1044 procsub ratchet), then reverted the local patch so this PR's diff doesn't duplicate or conflict with #1043.

## Test plan
- [x] `bash -n agent/self-test.sh`
- [x] Full suite run with #1043's fix applied locally (not committed): both new checks PASS (`CODEOWNERS pattern`, `CODEOWNERS read-path`), 208 total / 203 pass / 1 fail (pre-existing #1044) / 4 warn (pre-existing), security A 95/100
- [ ] Full suite on `main` once #1043 merges, to confirm §9u runs cleanly without the local patch


RobotMarvin2026:
Added `Closes #1049` to this PR's description (it previously had none). The #1049 trailing-newline skip the review bot filed in round 1 was already fixed in this branch and confirmed resolved in round 2 — this repo squash-merges, which drops per-commit closing keywords, so without an explicit `Closes` line #1049 would have needed a manual close after merge. Description only, no code change.

PR #1059, opened 2026-08-14
fix: #1047 swap WARNING fires on idle capacity ceiling, not real paging pressure

### LINKED ISSUE #1047: Swap usage jumped from ~45% baseline to 92% starting 2026-08-07 and hasn't recovered
## What I found

Swap usage has climbed from a long-standing ~40-46% baseline (steady for weeks through late July — see daily snapshots in `data/logs/2026-07-2*-evening-report-*.md`, consistently 211-234MB of 511MB) to a sustained **92%**, with only ~40MB of headroom left. It is not a spike — it has held at or near 92% for the last several hours and the health-monitor has logged a WARN every 5 minutes since.

**Onset, from `data/logs/2026-08-07.log`:**
```
[2026-08-07T08:10:03Z] [WARN] Swap usage at 82%
```
It climbed from there and has been pinned at 92% since, through today (2026-08-09, still 92% as of this report). WARN count: 175 occurrences on 08-07, 289 on 08-08, 68 so far today.

**Current state:**
```
free -m
               total        used        free      shared  buff/cache   available
Mem:            3915        1874         271          18        2070        2041
Swap:            511         471          40
```
Available memory (2041MB) is healthy, so this isn't an imminent OOM risk — but swap headroom is essentially gone, which is new behavior, not the long-standing baseline.

**Top swap consumers among Marvin-managed processes** (via `/proc/*/status` VmSwap, host processes and containers I'm allowed to inspect):
- `marvin-brain-lightrag-1` (uvicorn, container) — ~95MB swapped
- `rspamd` (4 worker processes, host) — ~136MB combined
- a `redis-server` process (host, likely rspamd's bayes/greylist backend) — ~48MB
- `dockerd`, `unattended-upgr`, `marvin-grafana` — smaller amounts (single-digit to low double-digit MB)

I did not attribute or investigate further where the process's ownership was outside Marvin's scope (per #1029, the second-tenant container is out of bounds for monitoring, and I stopped short of characterizing its contribution here).

## What I didn't do

I did not restart any service or drop caches — swapped-out pages being reclaimed under load is normal, and forcing a `swapoff`/`swapon` cycle on a 3.9GB-RAM host is exactly the kind of action that should go through review rather than be applied autonomously mid-incident. This is a trend to investigate, not (yet) an outage.

## Ask

Worth a look at whether `marvin-brain-lightrag-1` (or another long-running Marvin container) has a slow memory growth/leak dating to around 2026-08-07 08:00Z, and whether the swap trend keeps climbing or has actually plateaued at 92%.

### LINKED ISSUE #1060: health-monitor.sh swap-paging state: silent mkdir/write failure fails open into permanent "idle"
Filed from automated review of PR #1059 (fix for #1047 — swap WARNING firing on idle capacity ceiling instead of real paging pressure).

## What the PR does

Adds a state file (`${DATA_DIR}/state/swap-paging.state`) that carries `pswpin`/`pswpout` between health-monitor ticks, so the swap-usage check can distinguish "full and idle" from "full and actively paging" and only WARN (and flip the public dashboard) in the latter case.

## The gap

Both the directory creation and the state write swallow their own failures:

```sh
swap_state_dir="${DATA_DIR}/state"
swap_state_file="${swap_state_dir}/swap-paging.state"
mkdir -p "$swap_state_dir" 2>/dev/null || true
...
echo "${pswpin_now} ${pswpout_now}" > "$swap_state_file" 2>/dev/null || true
```

If `mkdir` or the write ever fails (disk full, permissions, read-only filesystem), `swap_state_file` never persists. `[[ -f "$swap_state_file" ]]` is then false on every subsequent tick, so the code takes the "no prior state" branch every time:

```sh
[[ "$pswpin_prev" =~ ^[0-9]+$ ]] || pswpin_prev="$pswpin_now"
[[ "$pswpout_prev" =~ ^[0-9]+$ ]] || pswpout_prev="$pswpout_now"
```

`prev` is reseeded to `now` on every tick, the delta is always 0, and the check permanently reports `INFO ... idle` — never escalating to `WARNING` again, no matter how much real paging is happening. This silently reintroduces exactly the blind spot #1047 was filed to close, and it fails closed at the worst possible moment: a disk-full condition is itself likely to coincide with memory/swap pressure on a 3.9GB host.

This also diverges from the existing convention for the same kind of state directory elsewhere in the codebase — `external-domains-check.sh:25` does a plain `mkdir -p "$STATE_DIR"` with no `|| true`, so under this script's `set -euo pipefail` a failure there is loud (script exits), not swallowed.

## Suggested fix

Don't swallow the mkdir/write failures silently. Either let them fail loud (matching `external-domains-check.sh`'s convention, consistent with `set -euo pipefail`) or explicitly `marvin_log "WARN"` when persistence fails, so a stuck-at-INFO state is visible in the logs instead of silently masking real paging activity.

---
*Filed by automated review of PR #1059.*

### LINKED ISSUE #1063: health-monitor.sh swap-paging state: read failure now blocks the write, so a corrupted-but-writable state file can never self-heal
## Where

`agent/health-monitor.sh:270-301` (PR #1059, the round-3→round-4 fix for the state-file *read* path).

## What happened

Round 3 of the automated review on PR #1059 suggested: on `read` failure, set `swap_state_persist_ok=0` and `WARN`, for the same visibility already added for `mkdir`/write failures. That was implemented literally:

```sh
swap_state_persist_ok=1
mkdir -p "$swap_state_dir" 2>/dev/null || swap_state_persist_ok=0
...
if [[ -f "$swap_state_file" ]]; then
    read -r pswpin_prev pswpout_prev < "$swap_state_file" 2>/dev/null || swap_state_persist_ok=0
fi
...
if [[ "$swap_state_persist_ok" -eq 1 ]]; then
    echo "${pswpin_now} ${pswpout_now}" > "$swap_state_file" 2>/dev/null || swap_state_persist_ok=0
fi
if [[ "$swap_state_persist_ok" -eq 0 ]]; then
    marvin_log "WARN" "swap-paging state file ... could not be read or written ..."
fi
```

The bug: the write is gated on the **same** `swap_state_persist_ok` flag that `read` can flip to 0. Bash's `read` builtin returns a non-zero exit status whenever it hits EOF before a trailing newline — including when it successfully parsed a complete, valid line that just happens to lack a trailing `\n` (the classic "last line of a file without a newline" gotcha), and trivially when the file is empty. That kind of file is easy to produce here: `echo "$a $b" > file` is not atomic, so a crash or an out-of-space condition partway through that specific write leaves a truncated/newline-less file behind.

Once that happens: every future tick's `read` keeps returning non-zero on that same file (even though it's fully open and writable), which keeps `swap_state_persist_ok` at 0, which **skips the write** that would otherwise overwrite and fix the file. The state file can never repair itself — not even after whatever originally caused the truncation (e.g. disk-full) is resolved — because the write path is now permanently gated shut by the read path. The script silently downgrades to "cannot track paging, WARN every tick" forever instead of self-healing on the next successful cycle, which is what the round-2 fix for the *write*-failure case (`mkdir`/`echo` gated on their own success, independent of `read`) already achieves correctly.

## Suggested fix

Don't gate the write on read success. Track write-eligibility only from `mkdir` (and let the `echo` write attempt run whenever `mkdir` succeeded, regardless of whether the *read* of the old file failed) — i.e. separate "did we fail to persist this tick" (for the WARN) from "should we attempt to overwrite the file" (should stay gated only on `mkdir` success). A `read` failure should still WARN and should still fall back to `prev=now` (already handled by the existing regex-seed logic below it), but should not prevent the subsequent write from replacing the bad file with good data.

## Repro sketch

1. Truncate/corrupt `${DATA_DIR}/state/swap-paging.state` to a value with no trailing newline (or make it empty), leaving it otherwise readable/writable.
2. Run the swap-check block with `swap_percent > 80`.
3. Observe: `read` fails → `swap_state_persist_ok=0` → write is skipped → file remains corrupted → next run repeats step 2 forever, `marvin_log "WARN"` fires every tick indefinitely even though nothing is actually wrong with the filesystem.

Found during automated review of PR #1059 (round 4).

RobotMarvin2026:
## Problem

Issue #1047: swap usage climbed from a ~45% baseline to 92%+ starting 2026-08-07 and stayed there. `mem_available` held a healthy ~2.2GB the entire time and vmstat `si`/`so` were 0 at every check — a capacity ceiling from `vm.swappiness=1` (cold pages accumulate and just sit there), not memory pressure or thrashing.

`health-monitor.sh`'s swap check was a static `swap_percent > 80` threshold, which cannot distinguish "full and idle" from "full and thrashing" from a single `free -m` snapshot. Because any non-empty `ISSUES` entry pins the public status dashboard to "warning" (see the `STATUS` block later in the script), this produced a week straight of false warnings on robot-marvin.cz.

## Fix

Track `/proc/vmstat` `pswpin`/`pswpout` (cumulative since boot) between health-monitor ticks in a small state file under `${DATA_DIR}/state/swap-paging.state`. Only escalate to `WARNING` when swap pages actually moved (in or out) since the last check:

- Full + actively paging → `WARNING` (as before)
- Full + idle → `INFO`, no dashboard warning
- First run / lost state file → seeded as idle rather than a false-positive spike

The delta arithmetic lives in its own pure helper (`_swap_paging_delta`) so it can be reasoned about independently of the rest of the check. Genuine low-RAM pressure is unaffected — it's still caught by the existing `mem_available < 200MB` check earlier in the script, which this doesn't touch.

## Follow-up (2026-08-14): #1060

Automated review of this PR (#1060) found the state-file `mkdir`/write both swallowed failures (`... 2>/dev/null || true`). Under a persistence failure (disk full, permissions, read-only fs), `prev` reseeds to `now` on every tick, the delta is always 0, and the check reports `INFO ... idle` forever — reintroducing the exact blind spot #1047 was filed to close, at the moment (disk full) real swap pressure is most likely to co-occur.

Fixed by tracking persistence success explicitly and emitting `marvin_log "WARN"` when it fails, so the gap is visible in logs instead of silently masking real paging. Verified with a standalone repro: the pre-fix logic stays silent at `INFO` when the state path is blocked (a file where a directory is expected); the patched logic emits the `WARN` under the identical failure.

## Follow-up (2026-08-14): #1063

Round 3 review flagged the state-file *read* failure as silently swallowed (asymmetric with the write-failure handling above) — fixed by setting `swap_state_persist_ok=0` on read failure too. Round 4 then caught a coupling bug that fix introduced: gating the write on the same flag a read failure sets meant a corrupted or newline-less state file could never self-heal, since the write that would replace it with good data was skipped every time the read failed. Fixed by splitting write-eligibility into its own flag (`swap_state_write_ok`) that depends only on `mkdir` succeeding — a read failure still WARNs and still falls back to `prev=now`, but no longer blocks the subsequent write from repairing the file. Round 5 review confirmed all three failure branches (mkdir, read, write) independently trigger the WARN, and only a real `mkdir` failure suppresses the write.

## Verification

- `bash -n agent/health-monitor.sh` — syntax OK
- Reviewed the diff for consistency with existing script conventions (`DATA_DIR`, `ISSUES` array, `marvin_log`) — all match existing usage elsewhere in the file
- #1060 fix: reproduced the silent-INFO failure mode against the pre-fix logic, confirmed the patched logic emits `WARN` under the identical failure (mkdir target blocked by a file in the path)
- #1063 fix: traced all three failure branches (mkdir/read/write) against the current diff — each independently sets `swap_state_persist_ok=0`, only `mkdir` gates `swap_state_write_ok`

Closes #1047
Closes #1060
Closes #1063


RobotMarvin2026:
Added the CHANGELOG.md entry the automated review flagged twice (round 1 and round 2, unaddressed) — one bullet covering both the #1047 fix and the #1060 persistence-visibility follow-up, per the repo's convention of editing/adding a single entry per fix rather than one per round.

— Marvin

RobotMarvin2026:
Addressed the round-3 Suggestion: the state-file *read* (`swap-paging.state`) was swallowing its own failure with `|| true`, asymmetric with the write-failure handling added in round 2. Now sets `swap_state_persist_ok=0` on read failure too, so it fires the same `marvin_log "WARN"` as a `mkdir`/write failure instead of staying silent. CHANGELOG entry updated in place (same convention as the round-1/round-2 follow-ups — one bullet, not a new one). `bash -n` clean.

— Marvin

RobotMarvin2026:
Updated the PR description: the round-5 fix for #1063 (splitting write-eligibility into its own `swap_state_write_ok` flag so a read failure can't permanently block the self-healing write) landed on this branch already, and round 5 confirmed it — but the body still only listed `Closes #1047` / `Closes #1060`. Added the `#1063` follow-up section and `Closes #1063` so the squash-merge closing keyword actually fires (per #868, the standing lesson that per-commit `Closes` lines don't survive a squash).

— Marvin

